PR #12, opened 2022-06-10
ENH add implementation for init and push

adrinjalali:
Add implementation for `init` and `push`.

### DIFF
--- a/skops/hf_hub/_hf_hub.py
+++ b/skops/hf_hub/_hf_hub.py
@@ -10,6 +10,7 @@ from pathlib import Path
 from typing import List, Union
 
 from huggingface_hub import HfApi
+from requests import HTTPError
 
 
 def _validate_folder(path: Union[str, Path]):
@@ -149,6 +150,7 @@ def push(
     source: Union[str, Path],
     token: str = None,
     commit_message: str = None,
+    create_remote: bool = False,
 ):
     """Pushes the contents of a model repo to HuggingFace Hub.
 
@@ -170,6 +172,12 @@ def push(
     commit_message: str, optional
         The commit message to be used when pushing to the repo.
 
+    create_remote: bool, optional
+        Whether to create the remote repository if it doesn't exist. If the
+        remote repository doesn't exist and this parameter is ``False``, it
+        raises an error. Otherwise it checks if the remote repository exists,
+        and would create it if it doesn't.
+
     Returns
     -------
     None
@@ -181,6 +189,13 @@ def push(
     """
     _validate_folder(path=source)
     client = HfApi()
+
+    if create_remote:
+        try:
+            client.model_info(repo_id=repo_id, token=token)
+        except HTTPError:
+            client.create_repo(repo_id=repo_id, token=token, repo_type="model")
+
     client.upload_folder(
         repo_id=repo_id,
         path_in_repo=".",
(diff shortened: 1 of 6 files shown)


adrinjalali:
@merveenoyan this should be ready for a review :)

skops/hf_hub/_hf_hub.py
```diff
@@ -94,6 +169,15 @@ def push(*, repo_id: str, source: Union[str, Path], token: str = None):
...
         logged in using ``huggingface-cli login``.
 
+    commit_message: str, optional
+        The commit message to be used when pushing to the repo.
+
+    create_remote: bool, optional
+        Whether to create the remote repository if it doesn't exist. If the
+        remote repository doesn't exist and this parameter is ``False``, it
+        raises an error. Otherwise it checks if the remote repository exists,
+        and would create it if it doesn't.
```
merveenoyan, line 179: If this exists, why do we create_repo in examples and tests?

skops/hf_hub/_hf_hub.py
```diff
@@ -103,4 +187,23 @@ def push(*, repo_id: str, source: Union[str, Path], token: str = None):
...
+        except HTTPError:
+            client.create_repo(repo_id=repo_id, token=token, repo_type="model")
+
+    client.upload_folder(
+        repo_id=repo_id,
+        path_in_repo=".",
+        folder_path=source,
+        commit_message=commit_message,
+        commit_description=None,
+        token=token,
```
merveenoyan, line 205: Why is this needed?

skops/hf_hub/_hf_hub.py
```diff
@@ -103,4 +187,23 @@ def push(*, repo_id: str, source: Union[str, Path], token: str = None):
...
+            client.model_info(repo_id=repo_id, token=token)
+        except HTTPError:
+            client.create_repo(repo_id=repo_id, token=token, repo_type="model")
+
+    client.upload_folder(
+        repo_id=repo_id,
+        path_in_repo=".",
+        folder_path=source,
+        commit_message=commit_message,
+        commit_description=None,
```
merveenoyan, line 204: Why is this needed? 😅

skops/hf_hub/_hf_hub.py
```diff
@@ -103,4 +187,23 @@ def push(*, repo_id: str, source: Union[str, Path], token: str = None):
...
+        token=token,
+        repo_type=None,
```
merveenoyan, line 206: This is by default None and None refers to model BTW

adrinjalali: Yes, same as above, I'm just explicitly setting it in case in the future the default value changes.

skops/hf_hub/_hf_hub.py
adrinjalali: I've removed it from the example, but the test needs to test different scenarios, and the user might create the repo before calling this method, so the test makes sure that case is tested as well.

skops/hf_hub/_hf_hub.py
merveenoyan: ah gotcha!

merveenoyan (approved):
I didn't catch anything else other than above comments, LGTM
I had a couple of concerns around requirements, like maybe users might want to set the versions to be between X and Y but it's a big brain comment for early stage 😅

PR #20, opened 2022-06-27
add py3.7 support

adrinjalali:
This adds py3.7 support, since colab and kaggle don't support newer python versions.

Note that the rest of the PyData stack has moved on and doesn't support 3.7, so this would mean users are using older versions of those libraries on those platforms.

cc @merveenoyan


adrinjalali:
Damn, I forgot that I actually use py3.8 features in this project. I'll need to see how easy it is to get it working on the older python.

PR #44, opened 2022-07-20
TST Fixes test error caused by empty pickle

BenjaminBossan:
Fixes the bug as discussed on Slack.

Furthermore, tests have been refactored to rely on fixtures for the
model pickle file and the config.json, instead of checking those files
into the repo. The files are cleaned up after the tests finish running.
Existing tests have been adapted to use the fixtures.

Note: github says the model.pkl was renamed to .gitkeep because
both were empty files :)


skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -22,7 +24,60 @@ def _get_cwd():
...
+        return dir_path
+
+
+@pytest.fixture(scope="session")
+def model_pickle(repo_path):
+    # Create a simple pickle file for the purpose of testing
+    clf = LogisticRegression()
+    clf.fit([[0, 1], [1, 0]], [0, 1])
+    path = repo_path / "model.pickle"
+    if os.path.exists(path):
```
adrinjalali, line 44:
```suggestion
    if path.exists():
```

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -39,13 +94,13 @@ def test_validate_folder():
     ):
         _validate_folder(path=dir_path)
 
-    example_file = _get_cwd() / "sample_repo/config.json"
+    example_file = _get_cwd() / "sample_repo" / "config.json"
```
adrinjalali, line 97: instead of `_get_cwd() / "sample_repo"` we could use a fixture returning a temp folder.

adrinjalali (commented):
If `sample_repo` is going to be empty, then we can just remove it, and use a tempfolder instead to have our files there.

skops/hub_utils/tests/test_hf_hub.py
BenjaminBossan: good catch, will change it

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -1,28 +1,76 @@
...
+
+
+@pytest.fixture(scope="session")
+def model_pickle(repo_path):
+    # Create a simple pickle file for the purpose of testing
+    clf = LogisticRegression()
+    clf.fit([[0, 1], [1, 0]], [0, 1])
+    path = repo_path / "model.pickle"
+    if path.exists():
+        raise OSError(f"File {path} already exists")
```
adrinjalali, line 38: since we're not using a temp folder, happy to have this removed. Don't really mind if you want to keep it though.

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -1,28 +1,76 @@
...
+    path = repo_path / "config.json"
+    try:
+        with open(path, "w") as f:
+            json.dump(CONFIG, f)
+        yield path
+    finally:
+        # we cannot use path.unlink(missing_ok=True) since it was added in
+        # Python 3.8 only
+        if path.exists():
+            path.unlink()
```
adrinjalali, line 70: we can define our own `unlink` function in `fixes.py` to have all version dependent code in a single place as much as we can, so that when ditching py=3.7 we can remove redundant code easily.

adrinjalali (commented):
Other than the nits LGTM.

skops/hub_utils/tests/test_hf_hub.py
BenjaminBossan: Agree, this is not necessary.

skops/hub_utils/tests/test_hf_hub.py
BenjaminBossan: I added a new `path_unlink` function in `fixes.py` and use it for the tests.

BenjaminBossan:
> Looks good to me, except the empty `__init__.py` file. We don't need those empty files anymore.

I must have missed something o.O Anyway, I removed it.

PR #45, opened 2022-07-20
ENH add examples and dtypes to CardData and config.json

adrinjalali:
This PR adds functionality to store an example input and the input columns and their dtypes, so that the backend can better understand how to pass data to the model.


adrinjalali:
@merveenoyan you can see the kind of change I was suggesting in this PR.

skops/hub_utils/_hf_hub.py
```diff
@@ -84,12 +176,29 @@ def recursively_default_dict():
...
 
 
-def init(*, model: Union[str, Path], requirements: List[str], dst: Union[str, Path]):
+def init(
+    *,
+    model: Union[str, Path],
+    requirements: List[str],
+    dst: Union[str, Path],
+    task: str,
+    data,
```
BenjaminBossan, line 200: I'm not quite certain. Is `data` allowed to be None or not? From the docstring, it seems to be not, but in the tests you pass `data=None`.

BenjaminBossan (commented):
Good addition, thanks.

I have a few minor comments and questions, please take a look. Apart from that, could you please add unit tests to check if the dtypes are correctly determined and stored?

skops/hub_utils/_hf_hub.py
```diff
@@ -56,7 +64,76 @@ def _validate_folder(path: Union[str, Path]):
...
+    """Returns the example input of a model.
+
+    The input is converted into a dictionary which is then stored in the config
+    file.
+
+    Parameters
+    ----------
+    data: array-like
+        The input needs to be either a ``pandas.DataFrame`` or a
+        ``numpy.ndarray``. The first 3 rows are used as example input.
```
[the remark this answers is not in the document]
merveenoyan: I think it's hard to change it in the widget, it adds extra complexity, that's why it's used. (my guess)

examples/plot_hf_hub.py
```diff
@@ -76,7 +76,11 @@
 
 local_repo = mkdtemp(prefix="skops-")
 hub_utils.init(
-    model=pkl_name, requirements=[f"scikit-learn={sklearn.__version__}"], dst=local_repo
+    model=pkl_name,
+    requirements=[f"scikit-learn={sklearn.__version__}"],
+    dst=local_repo,
+    task="tabular-classification",
```
merveenoyan, line 82: Why do we need to provide task name in `hub_utils.init()` if the user already provides it in the model card?

skops/card/_model_card.py
```diff
@@ -27,23 +31,44 @@ def _extract_estimator_config(model):
...
     **card_kwargs,
 ):
     """Creates a model card for the model and saves it to the target directory.
 
     Parameters:
     ----------
-    model: estimator
-        scikit-learn compatible estimator.
-    card_data: CardData
+    path: str
```
merveenoyan, line 42: I think we should give people freedom to save their models themselves, I don't like this. we can add both path or model, it wouldn't be too bad.

skops/card/_model_card.py
```diff
@@ -27,23 +31,44 @@ def _extract_estimator_config(model):
...
     card_kwargs:
         Card kwargs are information you can pass to fill in the sections of the
         card template, e.g. model_description, citation_bibtex, get_started_code.
     """
     ROOT = skops.__path__
+
+    # Load the model from the existing directory.
+    config = get_config(path)
+    model_path = Path(path) / config["sklearn"]["model"]["file"]
+    with open(model_path, "rb") as f:
```
merveenoyan, line 56: this limits the user imo.

skops/card/_model_card.py
```diff
@@ -27,23 +31,44 @@ def _extract_estimator_config(model):
...
+
+    card_data = card_data or CardData()
+    card_data.tags = card_data.tags or list()
+
+    # Read relevant info from the config file and add them to the CardData
+    # object.
+    if "sklearn" not in card_data.tags:
+        card_data.tags += ["sklearn"]
+
+    if config["sklearn"]["task"] not in card_data.tags:
```
merveenoyan, line 67: image

skops/hub_utils/_hf_hub.py
```diff
@@ -9,9 +9,17 @@
...
 
+import numpy as np
 from huggingface_hub import HfApi, snapshot_download
 from requests import HTTPError
 
+SUPPORTED_TASKS = [
+    "tabular-classification",
+    "tabular-regression",
+    "text-classification",
+    "text-regression",
```
merveenoyan, line 20: What is `text-regression`? We don't have such task in Hub, no? 
Ekran Resmi 2022-07-21 13 11 53

skops/hub_utils/_hf_hub.py
```diff
@@ -71,6 +148,21 @@ def _create_config(*, model_path: str, requirements: List[str], dst: str):
     dst : str, or Path
         The path to an existing folder where the config file should be created.
 
+    task: str
```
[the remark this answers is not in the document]
adrinjalali: Done, using Literal now.

skops/hub_utils/_hf_hub.py
adrinjalali: Yeah the tests aren't supposed to be passing, I haven't done the tests yet. Was just wondering if you think the API is okay.

adrinjalali (commented):
Some conversations are still open, need to address those before I add tests.

skops/hub_utils/_hf_hub.py
osanseviero: Related issue from `hub-docs` for visibility https://github.com/huggingface/hub-docs/issues/137. 

The issue has some complexity for `text-regression` since as of now `transformers` uses `text-classification` to do both classification and regression, and splitting it now would be complex (as per https://github.com/huggingface/hub-docs/issues/137#issuecomment-1122859453). The "consistent" approach for the Hub would be using `text-classification` for both tasks. 

cc @narsil @julien-c

adrinjalali: I don't think we have to change anything on the transformers side, and those users might be happy with calling regression classification, but I don't think a sklearn pipeline which has a `LinearRegression` at the end of it would be a good fit for a `text-classification` task. As we also talked about it in the issue you mention, from the statistical inference / ML perspective, classification and regression are not really the same thing.

adrinjalali: We also don't have to fully support it now. Users can set the tag and certain things, like inference wouldn't work, and they can hack around and use the workaround of calling a regression classification to get a widget if they want to, and later we can fix the issue.

osanseviero: Yes, that sounds like a good path forward for me so we can move this forward, but I'll think a bit about this topic as the point you make is totally valid, and **it will** be confusing for those users. I think this might be solved via an alias as discussed in the issue, but let's think a bit more about this

Narsil: Purely from the `transformers` side we can definitely add an alias. https://github.com/huggingface/transformers/blob/main/src/transformers/pipelines/__init__.py#L128

Also there's an open issue (with not much activity) to get actually text-regression in addition to text-classification.

skops/hub_utils/_hf_hub.py
```diff
@@ -56,7 +64,76 @@ def _validate_folder(path: Union[str, Path]):
...
+    # TODO: this is going to fail for Structured Arrays. We can add support for
+    # them later if we see need for it.
+    if isinstance(data, np.ndarray):
+        return [(f"X{x}", str(data.dtype)) for x in range(data.shape[1])]
+
+    raise ValueError("The data is not a pandas.DataFrame or a numpy.ndarray.")
+
+
+def _create_config(
+    *, model_path: str, requirements: List[str], dst: str, task: str, data
```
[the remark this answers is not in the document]
BenjaminBossan: Yes, I agree that especially different input data types can be tricky with mypy (especially especially since numpy support is only meh). If it's possible to comfortably use mypy without enforcing it everywhere, I would be in favor (I have only used it in strict mode in the past, so no experience with non-strict).

> They also help with IDE hints when users are coding and using the API, it's not just about static type checkers.

That is, if they are indeed correct, which is currently not enforced.

skops/hub_utils/_hf_hub.py
```diff
@@ -56,7 +64,76 @@ def _validate_folder(path: Union[str, Path]):
...
+        if isinstance(data, pd.DataFrame):
+            return list(zip(data.dtypes.index, data.dtypes.astype("str")))
+    except ImportError:
+        # pandas is not installed, the data cannot be a pandas DataFrame
+        pass
+
+    # TODO: this is going to fail for Structured Arrays. We can add support for
+    # them later if we see need for it.
+    if isinstance(data, np.ndarray):
+        return [(f"X{x}", str(data.dtype)) for x in range(data.shape[1])]
```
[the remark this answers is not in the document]
BenjaminBossan: I didn't know, makes sense.

skops/card/_model_card.py
```diff
@@ -27,23 +31,44 @@ def _extract_estimator_config(model):
...
+        card_data.tags += [config["sklearn"]["task"]]
+
+    setattr(card_data, "widget", config["sklearn"]["example_input"])
```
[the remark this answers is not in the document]
BenjaminBossan: So `card_data.widget = config["sklearn"]["example_input"]` wouldn't work?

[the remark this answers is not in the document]
adrinjalali: It does 😮 Somehow I thought it wouldn't.

skops/hub_utils/_hf_hub.py
adrinjalali: The consensus seems to be that we'll add `text-regression`, either as an alias or a task, but not a prerequisite for this PR. Therefore resolving the discussion. Thanks for all the comments.

skops/hub_utils/_hf_hub.py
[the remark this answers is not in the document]
adrinjalali: I've opened https://github.com/skops-dev/skops/issues/52 to track this.

adrinjalali (commented):
I'm still not clear what you mean here @merveenoyan : https://github.com/skops-dev/skops/pull/45/files#r926549925

adrinjalali (commented):
The tests fail because persisting dtypes as strings is not trivial. I'm thinking of removing them from this PR and marking it as future work. WDYT @BenjaminBossan ?

BenjaminBossan:
> The tests fail because persisting dtypes as strings is not trivial. I'm thinking of removing them from this PR and marking it as future work.

Let me try to understand the problem: We would like to encode the numpy dtype in a JSON. The only sensible JSON type would be string. Therefore, we need a mapping from dtype to str and vice versa. Can we use [this collection](https://github.com/numpy/numpy/blob/098f874144161b6a49efa5108846a408ca8f39b8/numpy/array_api/_dtypes.py#L18)? I checked and numpy dtypes are hashable, so we could create a dict.

adrinjalali:
Yes, that would be an idea, but that list doesn't include the string types, and it also doesn't include the little/big endian info. the dtype object actually includes quite a bit of information: https://numpy.org/doc/stable/reference/arrays.dtypes.html

I was motivated to include this partly because we need to have a list with the order of the columns, and I thought I could as well include the dtypes.

But now with the complexity I see, I rather only include the columns names here, and move the dtype support to another PR, which itself would be quite a sizable PR.

adrinjalali:
* we need the list of column names with the order because on the inference side we get a dict which is not necessarily with columns in the same order as the original data.

BenjaminBossan:
> Yes, that would be an idea, but that list doesn't include the string types, and it also doesn't include the little/big endian info. the dtype object actually includes quite a bit of information

Got it, then let's add this later. For my understanding, how are dtypes selected right now?

Looking forward, let's assume we add dtypes in a later step after the official release, would the current implementation ensure that the current config.json files would still work?

Note: I think some docstrings still refer to the dtype being stored, please check.

adrinjalali:
> Got it, then let's add this later. For my understanding, how are dtypes selected right now?

What do you mean by selected? Where?

> Looking forward, let's assume we add dtypes in a later step after the official release, would the current implementation ensure that the current config.json files would still work?

Yes.

BenjaminBossan:
> What do you mean by selected? Where?

I meant by the backend, referring to the original description.

> This PR adds functionality to store an example input and the input columns and their dtypes, so that the backend can better understand how to pass data to the model.

So for now, any possible dtype conversion has to be handled by the user and we defer providing functionality around it to a future PR. Is this understanding correct?

> Yes.

Okay, great.

skops/card/_model_card.py
```diff
@@ -27,23 +31,44 @@ def _extract_estimator_config(model):
 
 
 def create_model_card(
-    model,
-    card_data,
+    path,
+    card_data=None,
```
adrinjalali, line 35: Are we okay with the change introduced here? 

@merveenoyan ?

adrinjalali (commented):
I think this is ready for a final review.

Once this is merged, we need to fix things up on the api-inference-community side.

adrinjalali:
> I meant by the backend, referring to the original description.

This is where input curation is happening: https://github.com/adrinjalali/api-inference-community/blob/7051256b61f82592216af5b4c7b11796a2fb7218/docker_images/sklearn/app/pipelines/tabular_classification.py#L45

That piece of code would need to use the info we're providing here, otherwise it's kinda giving the model very unreliable data.

> So for now, any possible dtype conversion has to be handled by the user and we defer providing functionality around it to a future PR. Is this understanding correct?

yes, or for now we'll see how it works.

BenjaminBossan:
> That piece of code would need to use the info we're providing here, otherwise it's kinda giving the model very unreliable data.

Yes, makes sense, thanks for the pointer.

adrinjalali:
Due to merge conflicts, https://github.com/skops-dev/skops/pull/37 should be merged before this one.

adrinjalali:
I've removed the changes to the model card here, and will open another PR specifically for that, to make this PR smaller.

This is ready for a review @BenjaminBossan

BenjaminBossan:
@adrinjalali The codecov check fails. Should we ignore it, fix it, or remove the check entirely?

adrinjalali:
I wouldn't remove it, and the one which is more important is the `patch` rather than `project`. It fails because there are certain parts, like `_min_dependencies.py` which are not tested. I find it quite useful, but I see it as a guideline rather than something which has to be green before we merge. I actually use the `patch` report to make sure new code is covered as much as we can.

adrinjalali:
Already failing due to mypy, don't like this. Gonna check tomorrow.

BenjaminBossan:
> Already failing due to mypy, don't like this. Gonna check tomorrow.

Pretty harmless, probably failing due to missing imports after merging. I changed some instances of `List[...]` to `list[...]` because it's nicer IMO, but it does require the `__future__` import for Python 3.7.

BenjaminBossan (commented):
LGTM overall.

I have a few comments which are more of aesthetic nature, and one about what kind of data we want to allow for text tasks, where the current implementation might be too restrictive.

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -90,61 +107,140 @@ def test_validate_folder(config_json):
...
+                    "environment": ['scikit-learn="1.1.1"', "numpy"],
+                    "example_input": {"data": ["test", "text", "problem"]},
+                    "model": {"file": "model.pkl"},
+                    "task": "text-classification",
+                }
+            },
+        ),
+    ],
+)
+def test_create_config(data, task, expected_config):
```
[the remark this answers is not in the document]
adrinjalali: there are going to be more configs, so rather not put it in a fixture. Also don't really like repeating tests, I rather parametrize tests as much as we can, makes it easier to add new tests.

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -90,61 +107,140 @@ def test_validate_folder(config_json):
...
+        if "tabular" in task:
+            # text data doesn't introduce any "columns" in the configuration
+            keys += ["columns"]
+        for key in keys:
+            assert sorted(config["sklearn"][key]) == sorted(
+                expected_config["sklearn"][key]
+            )
+
+
+def test_create_config_invalid_data(temp_path):
```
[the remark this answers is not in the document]
adrinjalali: One can then rename the test ;)

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -6,16 +6,32 @@
...
+from skops.hub_utils._hf_hub import (
+    _create_config,
+    _get_column_names,
+    _get_example_input,
+    _validate_folder,
+)
 from skops.hub_utils.tests.common import HF_HUB_TOKEN
 from skops.utils.fixes import metadata, path_unlink
 
+iris = load_iris(as_frame=True, return_X_y=False)
```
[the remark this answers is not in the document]
adrinjalali: Data was in a fixture, but we can't put fixtures in `pytest.mark.parametrize`, so I moved it to a variable. I like it this way, easy to follow/understand as well. Fixtures can make the code look too much like magic sometimes.

skops/hub_utils/_hf_hub.py
```diff
@@ -87,13 +193,33 @@ def recursively_default_dict() -> MutableMapping:
     config = recursively_default_dict()
     config["sklearn"]["model"]["file"] = str(model_path)
     config["sklearn"]["environment"] = requirements
+    config["sklearn"]["task"] = task
+
+    if "tabular" in task:
+        config["sklearn"]["example_input"] = _get_example_input(data)
+        config["sklearn"]["columns"] = _get_column_names(data)
+    elif "text" in task:
+        if isinstance(data, list) and all(isinstance(x, str) for x in data):
```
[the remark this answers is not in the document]
adrinjalali: Yes I know we're being very restrictive with the types we accept here. We also don't accept all array-like things accepted by sklearn. Adding support for those is nice, and necessary, but I was hoping to have that in a future PR and have something basic here, which doesn't necessarily limit users. They can still pass a list to this method even if their data is originally in a `Series` form.

Opened https://github.com/skops-dev/skops/issues/65 to track this.

adrinjalali:
Ping @BenjaminBossan , can we merge this now?

skops/hub_utils/tests/test_hf_hub.py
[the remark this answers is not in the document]
BenjaminBossan: It is possible to achieve this effect using the [`request`](https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#parametrizing-fixtures) fixture but it's a bit difficult to understand.

BenjaminBossan (approved):

skops/hub_utils/tests/test_hf_hub.py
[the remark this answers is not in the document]
adrinjalali: Yes I saw that, rather have things simple.

PR #54, opened 2022-07-25
CI mark hub test as flaky

### LINKED ISSUE #40: Use a retry marker/decorator for tests
Our tests fail sometimes since the backend errors when it should succeed, and usually running the tests again would solve the issue.

In `huggingface_hub` there is a decorator called [`retry_endpoint`](https://github.com/huggingface/huggingface_hub/blob/b6accae2ebfe2d0986934f2933b7bf2fb3811726/tests/testing_utils.py#L176) used in tests, and we can use the same thing. But we need to make it work with pytest fixtures and decorators (such as `pytest.mark.parameterize`) first.

There are some pointers on how to achieve this here: https://twitter.com/adrinjalali/status/1516346047156699138

adrinjalali:
Fixes #40 

cc @BenjaminBossan 

You might like this @merveenoyan

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -31,8 +31,8 @@ jobs:
 
     - name: Install dependencies
       run: |
-        pip install .[docs]
-        pip install pytest black=="22.6.0" isort=="5.10.1" pytest-cov
+        pip install .[docs,tests]
+        pip install black=="22.6.0" isort=="5.10.1"
         python --version
         pip --version
         pip list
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
It is actually possible to have a backoff with flaky, the docs give this example:

```python
import time

def delay_rerun(*args):
    time.sleep(1)
    return True

@flaky(rerun_filter=delay_rerun)
def test_something_else():
    ...
```

But I suppose it's not needed, so I'm fine with the change as is.

Not sure if you already want this merged, Adrin, since you didn't put any reviewers of the PR.

adrinjalali:
> Not sure if you already want this merged, Adrin, since you didn't put any reviewers of the PR.

Yes, if I don't want a PR to be merged, I'll have it in "Draft" mode or I'll leave a comment with TODOs.

BenjaminBossan:
Gotcha

PR #60, opened 2022-07-26
MNT make init atomic

### LINKED ISSUE #42: Make init function atomic
I just had a look at `init` and it's not atomic: https://github.com/skops-dev/skops/blob/41716e56a71f8eb75d3a131730fefbc4e4c203e8/skops/hub_utils/_hf_hub.py#L92-L122

E.g. if the last step (`_create_config`) fails, the user fixes that step, then calls `init` again, it will suddenly fail in an earlier step because the directory is not empty anymore. A possible fix would be to create a temp directory, put everything there, and in the last step move that directory to the destination. @adrinjalali @merveenoyan WDYT?

adrinjalali:
Fixes #42 

Need to merge https://github.com/skops-dev/skops/pull/45 first so that this can properly be tested.


merveenoyan:
Super clean 🧼

BenjaminBossan:
I see why the added test does not cover the `Exception` part: It fails outside of the `try`, namely here:

https://github.com/adrinjalali/skops/blob/593f605e3d364a77867b799f46b816e488b4e45d/skops/hub_utils/_hf_hub.py#L277-L279

So for the test to actually cover the line, the error would have to occur inside of `_create_config`. Perhaps the most robust way to achieve that would be to patch the function and make it raise?

adrinjalali:
It's all green now @BenjaminBossan

BenjaminBossan:
Could you please resolve the merge conflict?

adrinjalali:
Done, and CI is green!

BenjaminBossan (approved):
Thanks, looks good

PR #61, opened 2022-07-27
Add mypy check to CI and fix existing mypy errors

BenjaminBossan:
I created this PR and, at the suggestion of Adrin, #62, to evaluate whether we want to use mypy and if yes, how.

In this PR, mypy is used in strict mode and as can be seen, it does require a few workarounds to make it pass. Most notably, we have:

```python
class SklearnBaseEstimator(Protocol):
    def get_params(self, deep: bool = True) -> dict[str, Any]:
        ...
```

is needed (and conveniently, CI complains that `...` was not covered by tests) since sklearn is not typed.

```python
class _SklearnConfig(TypedDict):
    environment: list[str]

class ConfigJson(TypedDict):
    sklearn: _SklearnConfig | None
```

to represent the `config.json` (of course, `TypedDict` doesn't exist in Python 3.7).

```python
    def recursively_default_dict():  # type: ignore
        return collections.defaultdict(recursively_default_dict)
```

a recursive data structure not based on a type, ugh.

Moreover, there are still lot's of `Any`s, 

On the other hand, the non-strict PR shows that mypy let's us slip a bunch of untyped functions and methods, which decreases our confidence on how much we can rely on the type checker to help us out (e.g. when refactoring code).

Between strict and non-strict, there is of course a spectrum using the available mypy settings to silence specific types of errors.

Another question is whether we want to add mypy to the pre-commit hooks. It would work well for now but mypy tends to get slow when the code base gets large. The daemon could help, I don't know how big of a difference it makes and how convenient it is to use.

My personal opinion is that I'm pretty meh on the value mypy brings currently. But of course, the larger the code base will become, the more potential it has to uncover hidden errors and guide in refactoring. Editor type hints are not so important for me personally.

ping @merveenoyan @adrinjalali 

PS: I'll take a look at the failing codecov once we decide how to proceed.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -56,6 +56,9 @@ jobs:
       run: |
         python -m pytest -s --cov=skops --doctest-modules skops/
 
+    - name: Mypy
+      run: mypy --strict skops
+
     - name: Upload coverage to Codecov
       uses: codecov/codecov-action@v2
       with:
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,3 +14,7 @@ filterwarnings = [
 markers = [
     "network: marks tests as requiring internet (deselect with '-m \"not network\"')",
 ]
+
+[tool.mypy]
+exclude = "(\\w+/)*test_\\w+\\.py$"
+ignore_missing_imports = true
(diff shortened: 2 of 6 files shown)


BenjaminBossan:
After discussion with @adrinjalali we decided to go for non-strict type checks for the time being (#62).

PR #62, opened 2022-07-27
MNT Add non-strict mypy to CI and fix errors

BenjaminBossan:
This accompanies PR #61 to contrast the difference between strict and
non-strict mypy usage.

### DIFF
--- a/skops/_min_dependencies.py
+++ b/skops/_min_dependencies.py
@@ -17,6 +17,7 @@ dependent_packages = {
     "pytest-cov": ("2.9.0", "tests", None),
     "flake8": ("3.8.2", "tests", None),
     "mypy": ("0.770", "tests", None),
+    "types-requests": ("2.28.5", "tests", None),
     "flaky": ("3.7.0", "tests", None),
     "sphinx": ("3.2.0", "docs", None),
     "sphinx-gallery": ("0.7.0", "docs", None),
--- a/skops/_min_dependencies.py
+++ b/skops/_min_dependencies.py
@@ -16,7 +16,7 @@ dependent_packages = {
     "pytest": (PYTEST_MIN_VERSION, "tests", None),
     "pytest-cov": ("2.9.0", "tests", None),
     "flake8": ("3.8.2", "tests", None),
-    "mypy": ("0.770", "tests", None),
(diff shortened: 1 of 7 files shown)


skops/hub_utils/_hf_hub.py
```diff
@@ -112,7 +114,7 @@ def init(*, model: Union[str, Path], requirements: List[str], dst: Union[str, Pa
     None
     """
     dst = Path(dst)
-    if dst.exists() and next(dst.iterdir(), None):
+    if dst.exists() and bool(next(dst.iterdir(), None)):
```
adrinjalali, line 117: aha? why do we need explicit bool here? This is proper python:

``` python
>>> if 3 or None:
...     print(1)
```

adrinjalali (commented):
This looks much more reasonable to me.

Should also add it to pre-commit hooks.

skops/hub_utils/_hf_hub.py
BenjaminBossan: The error is:

> skops/hub_utils/_hf_hub.py:117: error: Argument 1 to "next" has incompatible type "Generator[Path, None, None]"; expected "SupportsNext[bool]"
> skops/hub_utils/_hf_hub.py:117: error: Argument 2 to "next" has incompatible type "None"; expected "bool"

adrinjalali: That's very silly.

adrinjalali (commented):
If you add the pre-commit hook here, I'm happy to merge this as a first step, and re-visit in the future.

skops/_min_dependencies.py
```diff
@@ -16,7 +16,8 @@
     "pytest": (PYTEST_MIN_VERSION, "tests", None),
     "pytest-cov": ("2.9.0", "tests", None),
     "flake8": ("3.8.2", "tests", None),
-    "mypy": ("0.770", "tests", None),
+    "mypy": ("0.971", "tests", None),
```
adrinjalali, line 19: mypy isn't really a test dependency. It probably makes more sense to have it explicitly installed in CI the same as black and isort.

adrinjalali (approved):
Let's merge and see how it goes.

PR #63, opened 2022-07-28
Add mypy non strict with additional checks

BenjaminBossan:
This PR builds on top of #62 and follows the discussion in #61

Add a few flags to make mypy stricter, trying to strike a middle ground
between too strict and not strict enough. Make mypy pass after these
changes.

Additionally, add mypy to pre-commit hooks.

ping @adrinjalali

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -56,6 +56,9 @@ jobs:
       run: |
         python -m pytest -s --cov=skops --doctest-modules skops/
 
+    - name: Mypy
+      run: mypy skops
+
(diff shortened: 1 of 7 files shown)


adrinjalali (commented):
The kind of mypy which I'd be happy with, is the kind where:

if we type things, the types should be correct, therefore if we type things the wrong way, mypy would tell us we have a mistake.

And that's what mypy in non-strict mode does mostly.

BenjaminBossan:
After discussion with @adrinjalali we decided to go for non-strict type checks for the time being (#62).

PR #66, opened 2022-07-29
MNT ignore _min_dependencies.py in codecov

adrinjalali:
We don't test the file, and don't intend to, it just gives us warnings since it's not tested.


BenjaminBossan (approved):
Good riddance.

Issue #59, opened 2022-07-26
Implement __repr__ and __str__ for Card object


adrinjalali:
The `Card` class doesn't implement string representation methods, and therefore the user doesn't see a nice view of the object when they print it.

It would be nice to have the methods implemented so that when users do `print(card)`, they see something other than the hash of the memory address of the object.

cc @merveenoyan

BenjaminBossan:
I can also take this one. But let's wait until the #37 is merged.

adrinjalali:
#68 has solved this.

PR #68, opened 2022-08-02
ENH Card class: add repr and str methods

BenjaminBossan:
Solves #59

We decided not to follow the `eval(repr(card)) == card` convention, as it wouldn't work here.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,7 +10,6 @@ profile = "black"
 filterwarnings = [
     "error::DeprecationWarning",
     "error::FutureWarning",
-    "ignore::sklearn.exceptions.ConvergenceWarning",
 ]
 markers = [
     "network: marks tests as requiring internet (deselect with '-m \"not network\"')",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -10,6 +10,7 @@ profile = "black"
 filterwarnings = [
     "error::DeprecationWarning",
     "error::FutureWarning",
+    "ignore::sklearn.exceptions.ConvergenceWarning",
 ]
 markers = [
     "network: marks tests as requiring internet (deselect with '-m \"not network\"')",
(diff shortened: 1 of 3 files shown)


pyproject.toml
```diff
@@ -10,6 +10,7 @@ profile = "black"
 filterwarnings = [
     "error::DeprecationWarning",
     "error::FutureWarning",
+    "ignore::sklearn.exceptions.ConvergenceWarning",
```
adrinjalali, line 13: I wouldn't add this here, it's quite a bad one to completely ignore I'd say.

adrinjalali (commented):
Awesome!

pyproject.toml
BenjaminBossan: I reverted it.

skops/card/_model_card.py
```diff
@@ -5,13 +5,18 @@
...
 from typing import Any
 
 from modelcards import CardData, ModelCard
 from sklearn.utils import estimator_html_repr
 
 import skops
 
+# aRepr attributes can be used to control the behavior of repr
+aRepr.maxother = 79
+aRepr.maxstring = 79
```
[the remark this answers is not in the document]
BenjaminBossan: Could be, I'm not sure. Instead of a context manager, I just create a new instance of `Repr`, that should preclude the singleton problem too.

merveenoyan (approved):
Thanks a lot for tackling this, LGTM

skops/card/_model_card.py
```diff
@@ -54,7 +63,11 @@ class Card:
     >>> disp.figure_.savefig("confusion_matrix.png")
     ...
     >>> model_card.add_plot(confusion_matrix="confusion_matrix.png") # doctest: +ELLIPSIS
-    <skops.card._model_card.Card object at ...>
```
merveenoyan, line 57: I'm surprised doctest works even when you remove this 😅

PR #69, opened 2022-08-02
ENH Added evaluation

merveenoyan:
This PR is actually a fragment of #26 as #26 is a bit overwhelming. I changed the API and put evaluation a part of the Card class not to bother user with extra attributes.
@BenjaminBossan @adrinjalali

### DIFF
--- a/docs/model_card.rst
+++ b/docs/model_card.rst
@@ -60,7 +60,7 @@ template which includes the following slots for free text sections:
   to train the model.
 - ``"eval_methods"``: Details about evaluation process of the model.
 - ``"eval_results"``: Evaluation results that are later parsed as a table by
-  ``skops``.
+  :class:`skops.card.Card`.
 
 
 The template also contains the following sections that are automatically


BenjaminBossan:
@merveenoyan Did you ping us because we should already take a look or just to let us know?

merveenoyan:
@BenjaminBossan I pinged you to mainly look at UX of the evaluation and let me know if the direction is right.

BenjaminBossan:
Hmm, not so much I can say about this yet. It is very specific to HF, right? E.g. IIUC we assume that the `metric` used by sklearn is the same as defined on [HF/metrics](https://huggingface.co/metrics).

merveenoyan:
@BenjaminBossan having metrics in metadata is essentially for HF Hub, it also passes them to paperswithcode to associated task on associated dataset. (let's say you have text-classification, you'd like to evaluate it on GLUE) this is not the case usually for tabular models imo as people split their data on their own. I don't expect heavy usage for this for when you put your model for outside world, if someone else wants to benchmark their own models and programmatically access the best metrics, it would be good IMHO.

adrinjalali:
Overall I'd say it's in the right direction. I would change it a bit so that the method accepts the results of a scorer instead of trying to get the name and call it itself. It can be simplified by accepting a float value and a metric name. We probably also want to have it more like:

```
data description: ...
   metric 1: ...
   metric 2: ...
```

instead of a list of (data description, metric specs) as is right now.

But it seems like the eval results can't support that?

merveenoyan:
@adrinjalali with #71 I think we can let the user add `EvalResults` instead, no?

adrinjalali:
Yeah that would also make sense. I'm impartial on whether to construct `EvalResult` inside a method or accept one from the user. Note that #71 doesn't really make user ever create a `CardData` object themselves, the utility function does that.

BenjaminBossan:
I would tend towards not exposing modelcards internals to the user.

merveenoyan:
BTW evaluation metrics can be seen in this repository:
https://huggingface.co/merve/hf_hub_example-d4b188ba-be75-44ed-9c2b-d2cd0640c233 
Ekran Resmi 2022-08-08 18 46 25

skops/card/_model_card.py
```diff
@@ -78,10 +78,10 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
...
     """Model card class that will be used to generate model card.
 
-    This class can be used to write information and plots to model card and
-    save it. This class by default generates an interactive plot of the model
-    and a table of hyperparameters. The slots to be filled are defined in the
-    markdown template.
+    This class can be used to write information and plots to model card and save
+    it. This class by default generates an interactive plot of the model and a
+    table of hyperparameters. The slots to be filled are defined in the markdown
+    template.
```
adrinjalali, line 84: hehe, seems like whatever tool you're using to format these, makes the lines one character longer than mine :D

skops/card/_model_card.py
```diff
@@ -118,9 +118,10 @@ class Card:
 
     Examples
     --------
+    >>> from sklearn.metrics import (ConfusionMatrixDisplay, confusion_matrix,
+    ... accuracy_score, f1_score)
```
adrinjalali, line 122:
```suggestion
    >>> from sklearn.metrics import (
    ...     ConfusionMatrixDisplay,
    ...     confusion_matrix,
    ...     accuracy_score,
    ...     f1_score
    ... )
```

skops/card/_model_card.py
```diff
@@ -205,6 +215,26 @@ def add_plot(self, **kwargs: str) -> "Card":
             self._figure_paths[plot_name] = plot_path
         return self
 
+    def evaluate(self, metric_values):
```
adrinjalali, line 218: I think `evaluate` implies there is some evaluation being done by this method, since it's a verb. What do you think of `add_metrics`

skops/card/_model_card.py
```diff
@@ -205,6 +215,26 @@ def add_plot(self, **kwargs: str) -> "Card":
...
+        """Evaluates the model and returns the score and the metric.
+
+        Parameters
+        ----------
+        metric_values: dict
+        Dictionary of metric values.
```
adrinjalali, line 224:
```suggestion
            Dictionary of metric values.
```

skops/card/_model_card.py
```diff
@@ -205,6 +215,26 @@ def add_plot(self, **kwargs: str) -> "Card":
...
+
+        Returns
+        -------
+        self : object
+        Card object.
```
adrinjalali, line 229:
```suggestion
            Card object.
```

skops/card/_model_card.py
```diff
@@ -205,6 +215,26 @@ def add_plot(self, **kwargs: str) -> "Card":
...
+        """
+        self._eval_results = tabulate(
```
adrinjalali, line 231: should we add to a table or set the table? The current code doesn't allow the user to call this method multiple times.

skops/card/default_template.md
```diff
@@ -29,6 +29,12 @@ The model plot is below.
 
 {{ model_plot }}
 
+## Evaluation Results
+
+Below are the evaluation results.
```
adrinjalali, line 34: I think the user should also set the text of this section, explaining how the metrics are calculated. For example, are they from a cross validate call, a held out validation set, or a certain part of the dataset.

skops/card/_model_card.py
merveenoyan: I did it like that and it was bugging me with unexpected indent :(

skops/card/_model_card.py
merveenoyan: I used the plugin you sent though.

adrinjalali: hmm, somehow it's doing 80 for you and 79 for me.

skops/card/default_template.md
```diff
@@ -29,6 +29,14 @@ The model plot is below.
 
 {{ model_plot }}
 
+## Evaluation Results
+
+You can find the details about evaluation process and the evaluation results.
+
+{{ eval_methods }}
+
+{{ eval_results | default("[More Information Needed]", true)}}
```
adrinjalali, line 38: these need to be added to the user guide.

skops/card/_model_card.py
```diff
@@ -205,6 +219,24 @@ def add_plot(self, **kwargs: str) -> "Card":
             self._figure_paths[plot_name] = plot_path
         return self
 
+    def add_metrics(self, metric_values):
+        """Writes metrics to the model card.
+
+        Parameters
+        ----------
+        metric_values: dict
+            Dictionary of metric values.
```
adrinjalali, line 228: I think we should be consistent in terms of API here. `add` accepts kwargs, but here we expect a dict. We should choose only one. WDYT @BenjaminBossan

adrinjalali (commented):
Thanks @merveenoyan

skops/card/_model_card.py
BenjaminBossan: What plugin are you using?

BenjaminBossan: There seems to be a setting for the desired width, do you have the same one?

`"editor.wordWrapColumn"`

I installed the plugin and it seems to default to 80.

adrinjalali: I see, mine is set to 79.

BenjaminBossan: I have no preference.

skops/card/_model_card.py
merveenoyan: I fixed it :)

skops/card/default_template.md
merveenoyan: Added them :)

skops/card/tests/test_card.py
```diff
@@ -114,6 +114,14 @@ def test_metadata_keys(destination_path, model_card):
         assert "tags: dummy" in f.read()
 
 
+def test_add_metrics(destination_path, model_card):
+    model_card.add_metrics(**{"acc": 0.1})
+    model_card.add_metrics(f1=0.1)
+    model_card.save(Path(destination_path) / "README.md")
+    with open(Path(destination_path) / "README.md", "r") as f:
+        assert "acc" and "f1" and "0.1" in f.read()
```
BenjaminBossan, line 122: This line doesn't do what it's supposed to do. `"acc" and "f1" and "0.1"` is evaluated to `"0.1"` and only the latter is checked. You need to check all individually.

skops/card/_model_card.py
```diff
@@ -129,6 +134,14 @@ class Card:
     >>> model_card = card.Card(model)
     >>> model_card.metadata.license = "mit"
     >>> y_pred = model.predict(X)
+    >>> model_card.add_metrics(**{"accuracy": # doctest: +ELLIPSIS
+    ... accuracy_score(y, y_pred), # doctest: +ELLIPSIS
+    ... "f1 score": f1_score(y, y_pred, average='micro')}) # doctest: +ELLIPSIS
```
BenjaminBossan, line 139: IMO, this looks a bit strange, how about the formatting below?

```suggestion
    >>> model_card.add_metrics(**{
    ...     "accuracy": accuracy_score(y, y_pred),
    ...     "f1 score": f1_score(y, y_pred, average="micro"),
    ... })
```

Also, not sure if `# doctest: +ELLIPSIS` is required. I removed it and it still works for me.

BenjaminBossan (changes requested):
This is almost good to go from my point of view. I have a suggestion for better formatting of one of the examples, which you may or may not agree with. But I think the test needs to be changed, as I think it doesn't work as intended.

skops/card/tests/test_card.py
merveenoyan: Thanks a lot for noting it, I put them separately.

skops/card/_model_card.py
```diff
@@ -205,6 +220,23 @@ def add_plot(self, **kwargs: str) -> "Card":
             self._figure_paths[plot_name] = plot_path
         return self
 
+    def add_metrics(self, **kwargs: str) -> "Card":
+        """Takes metrics and their values to construct a table in the model card.
```
adrinjalali, line 224:
```suggestion
        """Add metric values to the model card.
```

skops/card/_model_card.py
```diff
@@ -205,6 +220,23 @@ def add_plot(self, **kwargs: str) -> "Card":
...
+
+        Parameters
+        ----------
+        **kwargs : dict
+            Metrics and their values.
```
adrinjalali, line 229:
```suggestion
            A dictionary of the form `{metric name: metric value}`.
```

examples/plot_model_card.py
```diff
@@ -115,6 +120,8 @@
     model_description=model_description,
 )
 y_pred = model.predict(X_test)
+model_card.add_metrics(accuracy=accuracy_score(y_test, y_pred))
+model_card.add_metrics(**{"f1 score": f1_score(y_test, y_pred, average="micro")})
```
adrinjalali, line 124: we should also set `"eval_methods"` in our example.

docs/model_card.rst
```diff
@@ -58,15 +58,17 @@ template which includes the following slots for free text sections:
...
   to train the model.
+- ``"eval_methods"``: Details about evaluation process of the model.
 
 The template also contains the following sections that are automatically
 generated by ``skops``.
 
 - ``"hyperparameter_table"``: Hyperparameters of the model.
 - ``"model_plot"``: A diagram of the model, most relevant in case the model is
   a complex scikit-learn :class:`~sklearn.pipeline.Pipeline`.
+- ``"eval_results"``: Table of evaluation results.
```
adrinjalali, line 69: I'd move this beside `eval_methods`. This is not automatically generated w/o user doing something anyway. It can be moved up, and mention that `add_metrics` creates this thing.

merveenoyan:
@adrinjalali I addressed your comments.

docs/model_card.rst
```diff
@@ -58,6 +58,10 @@ template which includes the following slots for free text sections:
   out, in case of questions about the model or the model card.
 - ``"citation_bibtex"``: Bibtex style citations for the model or resources used
   to train the model.
+- ``"eval_methods"``: Details about evaluation process of the model.
+- ``"eval_results"``: Evaluation results that are later parsed as a table by
+  ``skops``.
```
adrinjalali, line 63:
```suggestion
  :class:`skops.card.Card`.
```

adrinjalali (approved):
Otherwise LGTM.

adrinjalali (approved):

merveenoyan:
@BenjaminBossan can you review?

BenjaminBossan (approved):
Perfect, thanks

PR #71, opened 2022-08-03
ENH populate model card metadata with info from config.json

adrinjalali:
This PR adds a convenience function to populate the metadata section of model cards from the info available in `config.json`.

I'll be adding tests, but I'm wondering what you think of the API @merveenoyan @BenjaminBossan ?

### DIFF
--- a/examples/plot_hf_hub.py
+++ b/examples/plot_hf_hub.py
@@ -92,7 +92,11 @@ print(os.listdir(local_repo))
 # ==========
 # We will now create a model card and save it. For more information about how
 # to create a good model card, refer to the :ref:`model card example
-# <sphx_glr_auto_examples_plot_model_card.py>`
+# <sphx_glr_auto_examples_plot_model_card.py>`. The following code uses
+# :func:`~skops.card.metadata_from_config` which creates a minimal metadata
+# object to be included in the metadata section of the model card. The
+# configuration used by this method is stored in the ``config.json`` file which
+# is created by the call to :func:`~skops.hub_utils.init`.
 model_card = card.Card(model, metadata=card.metadata_from_config(Path(local_repo)))
 model_card.save(Path(local_repo) / "README.md")
 
--- a/examples/plot_hf_hub.py
+++ b/examples/plot_hf_hub.py
@@ -90,8 +90,10 @@ print(os.listdir(local_repo))
 # %%
 # Model Card
 # ==========
-# We will now create a model card and save it
-model_card = card.Card(model)
+# We will now create a model card and save it. For more information about how
+# to create a good model card, refer to the :ref:`model card example
+# <sphx_glr_auto_examples_plot_model_card.py>`
+model_card = card.Card(model, metadata=card.metadata_from_config(Path(local_repo)))
 model_card.save(Path(local_repo) / "README.md")
 
 # %%
--- a/examples/plot_hf_hub.py
+++ b/examples/plot_hf_hub.py
@@ -109,6 +109,7 @@ token = os.environ["HF_HUB_TOKEN"]
 repo_name = f"hf_hub_example-{uuid4()}"
 user_name = HfApi().whoami(token=token)["name"]
 repo_id = f"{user_name}/{repo_name}"
+print(f"Creating and pushing to repo: {repo_id}")
 
 # Now we can push our files to the repo. The following function creates the
 # remote repository if it doesn't exist; this is controlled via the
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -1,16 +1,67 @@
 from __future__ import annotations
 
 import copy
+import json
 import re
 import shutil
 import tempfile
 from pathlib import Path
-from typing import Any
+from typing import Any, Union
 
 from modelcards import CardData, ModelCard
 from sklearn.utils import estimator_html_repr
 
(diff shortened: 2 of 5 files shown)


BenjaminBossan:
I think we can proceed in this direction. It makes sense to use what's already there in the `config.json`.

adrinjalali:
This finally makes the model work out of the box, example: https://huggingface.co/skops/hf_hub_example-bdc26c1f-7e82-42eb-9657-0318315f2df0

Note: the output is actually not reliable yet, for the model to get the right input, https://github.com/huggingface/api-inference-community/pull/79 needs to be merged first.

@osanseviero you might like this one ;) 

Running `plot_hf_hub.py` (w/o the delete repo at the end) gives a nice example repo.

adrinjalali:
(There are some issues with the widget in linux, both Firefox and Chrome, and I've already told Mishig about them).

osanseviero:
Indeed I like this one, very cool! :fire:

examples/plot_hf_hub.py
```diff
@@ -90,8 +90,10 @@
 # %%
 # Model Card
 # ==========
-# We will now create a model card and save it
-model_card = card.Card(model)
+# We will now create a model card and save it. For more information about how
+# to create a good model card, refer to the :ref:`model card example
+# <sphx_glr_auto_examples_plot_model_card.py>`
+model_card = card.Card(model, metadata=card.metadata_from_config(Path(local_repo)))
```
BenjaminBossan, line 96: As a user, I might be confused about where the config comes from. How about adding a comment above the `init` call that mentions that a `config.json` is created for the user, similar to the comment in `plot_model_card.py`?

skops/card/_model_card.py
```diff
@@ -1,18 +1,73 @@
...
+
+    Parameters
+    ----------
+    config_path: str, or Path
+        Filepath to the ``config.json`` file, or the folder including that
+        file.
+
+    Returns
+    -------
+    card_data: ``CardData``
```
BenjaminBossan, line 43: Not sure about this suggestion:

```suggestion
    card_data: ``modelcards.CardData``
```

Reason being that users might not be familiar with `modelcards` and thus don't know where `CardData` comes from.

skops/card/_model_card.py
```diff
@@ -25,52 +80,79 @@ class Card:
...
     <skops.card._model_card.Card object at ...>
-    >>> model_card.save((Path("save_dir") / "README.md")) # doctest: +ELLIPSIS
+    >>> with tempfile.TemporaryDirectory() as tmpdir:
+    ...     model_card.save((Path(tmpdir) / "README.md")) # doctest: +ELLIPSIS
     ...
     """
 
-    def __init__(self, model: Any, model_diagram: bool = True) -> None:
+    def __init__(
+        self, model: Any, model_diagram: bool = True, metadata: CardData = None
```
BenjaminBossan, line 142:
```suggestion
        self, model: Any, model_diagram: bool = True, metadata: Optional[CardData] = None
```

Probably needs line break...

BenjaminBossan (approved):
Overall, this looks good to me, it's a nice convenience for the user. I have a few nits, please take a look at the comments.

API-wise, there could be a little bit of dissonance now. To add non-metadata to the card, users use a method: `card.add(foo=bar)`. To add metadata, they set attributes: `card.metadata.foo = bar`. I believe it's good that the distinction between non-metadata and metadata is not hidden from the user (as it was before), but it's still a bit strange to set them so differently.

As a proposal (could be added later), what do you think about providing a new method, `add_metadata`, which for users works exactly the same as `add` (i.e. works with **kwargs and returns self) and under the hood sets attributes on the metadata?

examples/plot_hf_hub.py
adrinjalali: clarified!

adrinjalali (commented):
I'm not sure about the idea of `add_metadata`, let's see how this goes I'd say.

skops/card/_model_card.py
```diff
@@ -25,52 +80,82 @@ class Card:
...
-    def __init__(self, model: Any, model_diagram: bool = True) -> None:
+    def __init__(
+        self,
+        model: Any,
+        model_diagram: bool = True,
+        metadata: Optional[CardData] = None,
+    ) -> None:
         self.model = model
-        self.hyperparameter_table = self._extract_estimator_config()
+        self._hyperparameter_table = self._extract_estimator_config()
```
merveenoyan, line 148: it's ironic for me to say this but please make changes out of scope of this PR in another one 😅 I saw one too many and thought some might be related but they aren't

merveenoyan (approved):
LGTM

skops/card/_model_card.py
adrinjalali: Good point, done here: https://github.com/skops-dev/skops/pull/74

skops/card/_model_card.py
```diff
@@ -76,14 +140,18 @@ class Card:
     >>> model_card.add_plot(confusion_matrix="confusion_matrix.png") # doctest: +ELLIPSIS
     Card(
       model=LogisticRegression(random_state=0),
-      license='mit',
```
adrinjalali, line 79: Opened a separate issue to fix repr to include metadata here: #75

PR #76, opened 2022-08-08
MNT Add metadata repr

### LINKED ISSUE #75: Add metadata (CardData) to model card's repr
https://github.com/skops-dev/skops/pull/71 moves metadata to the `CardData` instance inside the `Card` object, which is not included in `__repr__` at the moment.

The `__repr__` should include info from the metadata in a nice way as well.

cc @BenjaminBossan

BenjaminBossan:
Resolves #75

The `__repr__` and `__str__` of `ModelCard` now also shows the metadata.

Some decisions I made:

1. Metadata are shown after the model but before the rest
2. Metadata keys are shown as "metadata:some-key" to differentiate them
3. Widgets values are not shown, they are too long and provide little
   extra value in the repr.

Example:

```python
>>> card = Card(...)
>>> card
Card(
  model=LinearRegression(fit_intercept=False),
  metadata:language=fr,
  metadata:license=bsd,
  metadata:library_name=sklearn,
  metadata:tags=['sklearn', 'tabular-classification'],
  metadata:foo={'bar': 123},
  metadata:widget={...},
  model_description='A description',
  model_card_authors='Jane Doe',
  roc_curve='ROC_curve.png',
  confusion_matrix='confusion_matrix.jpg',
)
```


adrinjalali (approved):
This looks perfect.

What do you think of `metadata.key=value` instead of `metadata:key=value`? It looks more like python to me.

BenjaminBossan:
> What do you think of `metadata.key=value` instead of `metadata:key=value`? It looks more like python to me.

Yes, good point, I changed it.

adrinjalali (approved):

PR #82, opened 2022-08-09
DOC Add installation instruction to docs

BenjaminBossan:
I think the docs should have an installation instruction. Thus I added a new section there but LMK if I should put it elsewhere instead @skops-dev/maintainers

### DIFF
--- a/docs/installation.rst
+++ b/docs/installation.rst
@@ -1,0 +1,14 @@
+.. _installation:
+
+Installation
+============
+
+To install skops, run the following command in your Python environment:
+
+.. code-block:: bash
+
+    python -m pip install skops
+
+If you're interested in contributing to skops, please follow the `contribution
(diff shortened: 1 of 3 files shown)


docs/installation.rst
```diff
@@ -0,0 +1,14 @@
+.. _installation:
+
+Installation
+============
+
+To install skops, run the following command in your Python environment:
+
+.. code-block:: bash
+
+    python -m pip install skops
```
adrinjalali, line 10: I'm not sure why you prefer this to `pip install skops`.

adrinjalali (approved):

docs/installation.rst
BenjaminBossan: This the recommended way in [the docs](https://docs.python.org/3/installing/index.html). The reason I've seen in the past is that even though in general it's the same, in some situations `pip install` can lead to the package being installed globally, instead of the specific python environment. In contrast, `python -m pip install` will never fail where `pip install` works.

PR #89, opened 2022-08-10
MAINT Small refactor: figure_paths -> extra_sections

BenjaminBossan:
Functionally, nothing changed, this is just a renaming and a little
cleanup.

The reason is that this attribute can be used for more than just
figures, so a more generic name is a better fit. The name
`extra_sections` was chosen to mirror the `template_sections`.

Apart from renaming and adjusting some comments based on that, there
was a small change in which the items of the dict are accessed, making it
more efficient, and merging to f-strings into one.

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -164,7 +164,7 @@ class Card:
         else:
             self._model_plot = None
         self._template_sections: dict[str, str] = {}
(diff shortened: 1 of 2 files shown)


BenjaminBossan:

adrinjalali (approved):

Issue #87, opened 2022-08-10
Adding CV results to model cards


BenjaminBossan:
sklearn stores cross validation (e.g. from `cross_validate` or `GridSearchCV`) results in a specific tabular format. We can easily add those results to the model card with the help of `tabulate`. As a user, I can thus easily see all the different outcomes for the splits and hyperparameters.

One potential issue is that the resulting table can become quite big. We could either provide arguments to reduce the size or at the very least document how the user how they can shrink the size in (say, only show the top k best scores from grid search when there are hundreds of rows).

We could also consider this feature to work with pandas dataframes, since `df = pd.DataFrame(grid_search.cv_results_`) is probably a common pattern.

PR #90, opened 2022-08-10
ENH Add table to model cards

BenjaminBossan:
Solves #87

## Description

This feature adds a new method to the `Card` class called `add_table`. It
adds a generic table to the end of the model card by creating a new
section (similar to plots).

Any generic table of type `dict[str, list[Any]]` is supported, as well as
`DataFrame`s. However, the main use case is for adding CV results, as e.g.
created by `GridSearchCV`.

## Implementation

I hope you can stomach the type-y implementation Adrin ;)

I wanted to make the extra sections more generic in PR #89 as a
preparation for this PR (thus avoiding a big PR with refactor + new
feature). However, the changes in #89 were not sufficient (I guess
that's the danger of splitting the work).

The problem was that the formatting of adding the plots in the `save`
method was very specific to plots, as a markdown link would be added.
This, of course, does not work for tables.

My solution is that `_extra_sections` now takes `ExtraSection` objects as
values instead of strings. These objects only need to implement the
`format` method (so technically, `str` would still be okay). So for a
plot, the `PlotSection` class has a `format` method that returns a markdown
link. And for tables, the `TableSection` class has a `format` method that
uses `tabulate` to return a table as string.

This approach should hopefully be flexible enough for us to later add
new section types if needed.

At first glance, the implementation might look overly complex. However,
I think it's better than having a bunch of `if...else` inside of save that
formats each possible type differently. But please LMK if you have a
better idea, I'm open to suggestions.

## Tangential changes

So far, the plots were just added without an extra section to the bottom.
Now, there is an "# Additioinal Content" section and each plot and table
gets their own subsection.

## Examples

[Here](https://huggingface.co/skops/hf_hub_example-b959cadc-ffb3-4f07-856b-2d05d2d8f8e0#hyperparameter-search-results) is the hyperparameter table generated by `plot_model_card.py:`


adrinjalali:
Instead of an extra type, I'd just store the tables, and have a little function which takes the table and return what should be rendered in the md file. We don't need a protocol for adding tables to the model card.

BenjaminBossan:
> Instead of an extra type, I'd just store the tables, and have a little function which takes the table and return what should be rendered in the md file. We don't need a protocol for adding tables to the model card.

I assume you mean inside the `save` method. How would you know at that point what function to use to render the value? Sure, we could say str -> plot and dict -> table but I don't like that too much and it would probably not work with more types being added. IMO it also doesn't add much complexity -- you can directly see how different types of section are formatted and don't need to figure out which function is being called on them.

As to the protocol, it is not strictly needed, it's just to add a bit of extra type safety.

skops/card/_model_card.py
```diff
@@ -202,7 +275,51 @@ def add_plot(self, **kwargs: str) -> "Card":
...
+
+        Returns
+        -------
+        self : object
+            Card object.
+
+        """
+        for key, val in kwargs.items():
+            section = TableSection(table=val)
+            self._extra_sections[key] = section
```
adrinjalali, line 322: here we could add to `self._extra_tables` instead and then we know they're tables. Each method adds to their corresponding attribute.

BenjaminBossan: I wanted to avoid having separate containers for each type of addition. One reason is that it makes the code more lengthy (we'll later have to loop over each one separately), the other is that we lose the information of which order the extra sections are stored (e.g. I add a plot, a table, then a plot, but at the end I get plot, plot, table).

Another design I wanted to avoid is formatting already inside the `add_*` method, to not lock us in into a specific format at that point.

adrinjalali: okay, so if the order is important (which I agree with), we should make it a list of named tuples instead of a dict (and relying on the dict preserving insertion order during iteration).

If we remove the protocol and not typing our private attribute, then I'm fine with the PR

BenjaminBossan: Happy to make those changes.

Regarding the dict to list of tuples proposal: From my understanding, as long as you only add to the dict, the order should be preserved, right? So the difference would mainly be if we want to later allow deletion, or if a user wants to have multiple sections with the same name?

adrinjalali: that's true in modern new python versions. I think of it as an implementation detail, and rather not rely on it really.

BenjaminBossan: When the dict change was made in Python 3.6, it was indeed so:

> The order-preserving aspect of this new implementation is considered an implementation detail and should not be relied upon

([link](https://docs.python.org/3/whatsnew/3.6.html#new-dict-implementation))

With Python 3.7 (which is our min version), the order is officially part of the language:

> the insertion-order preservation nature of [dict](https://docs.python.org/3/library/stdtypes.html#typesmapping) objects [has been declared](https://mail.python.org/pipermail/python-dev/2017-December/151283.html) to be an official part of the Python language spec.

([link](https://docs.python.org/3/whatsnew/3.7.html))

So this wouldn't be my main concern here. I'd rather think about these implications:

1. We can have duplicate section names in the extra sections
2. Time complexity to retrieve items is now O(N) and there is no guarantee that only a single key matches (due to 1.)
3. More control about order (e.g. when deleting and reinserting items)

Due to the nature of the extra sections, which can contain anything, I think the trade-off is okay, whereas template sections should remain a dict.

adrinjalali: I know all these details about python's dict, but not many people do, and a "dictionary" isn't necessarily ordered in other languages. I tend to avoid things which are somewhat common in different languages but slightly differ in python for some specific reason.

adrinjalali (commented):
I really like it now.

Things left:
- user guide for the added functionality
- changelog entry

BenjaminBossan:
I'll work on that and let you know once it's ready for review.

BenjaminBossan:
@adrinjalali This is now ready for review.

In the changelog, I have not yet added subsection for new features, bugfixes, etc. LMK if now is a good time.

docs/changes.rst
```diff
@@ -7,6 +7,12 @@ skops Changelog
     :depth: 1
     :local:
 
+
+v0.2
+----
+- Tables, e.g. cross-validation results, can now be added to model cards using
+  the :meth:`.Card.add_table` method.
```
```suggestion
  the :meth:`.Card.add_table` method. :pr:`90` by :user:`Benjamin Bossan <BenjaminBossan>`
```

examples/plot_model_card.py
```diff
@@ -85,16 +85,18 @@
...
+# `README.md`.
 
 model_card = card.Card(model, metadata=card.metadata_from_config(Path(local_repo)))
 
 
 # %%
-# Pass information and plots to our model card
-# ============================================
+# Pass information, plots and tables to our model card
+# ====================================================
```
adrinjalali, line 99:
```suggestion
# Add information, plots and tables to our model card
# ===================================================
```

skops/card/tests/test_card.py
```diff
@@ -300,7 +311,100 @@ def test_with_metadata(self, card: Card, meth):
...
+        msg = "Empty table added"
+        with pytest.raises(ValueError, match=msg):
+            TableSection(table={"col": []})
+
+    def test_raise_error_df_empty(self):
+        pd = pytest.importorskip("pandas")
+        df = pd.DataFrame([])
+        msg = "Empty table added"
+        with pytest.raises(ValueError, match=msg):
+            TableSection(table=df)
```
adrinjalali, line 398: these could be a single parameterized test.

adrinjalali (commented):
Otherwise LGTM!

skops/card/tests/test_card.py
BenjaminBossan: Done. Please review again.

adrinjalali (approved):

PR #94, opened 2022-08-11
MAINT Rendering the model card

BenjaminBossan:
I wanted to discuss with you @skops-dev/maintainers what they think about this refactoring. Right now, the `Card.save` method performs the whole job of actually generating the `ModelCard` object. My plan is to simply factor this into a new method, `_generate_card`, and have the `save` method call this instead.

The main reason is to separate the concerns, i.e. `save` is only responsible for saving, not generating. What this affords us is that we can now easily add another method, `render`, which returns the generated model card as a string. This can, for instance, be used in the tests, as I have shown with one example. Instead of generating the card, storing it in a temp folder, reading it, then checking it, we can now just `render` the string and check it directly.

Please LMK if you agree and I will finish this PR.


adrinjalali (commented):
I like this, we can add alternative render options later.

BenjaminBossan:
> I like this, we can add alternative render options later.

Exactly.

Okay, then I'll finish the refactoring.

BenjaminBossan:

skops/card/_model_card.py
```diff
@@ -237,21 +237,14 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
-        """Save the model card.
-
-        This method renders the model card in markdown format and then saves it
-        as the specified file.
-
-        Parameters
-        ----------
-        path: str, or Path
-            filepath to save your card.
+    def _generate_card(self) -> ModelCard:
```
merveenoyan, line 240: This doesn't really generate the card itself because it already exists, but generates the content to be written. That would be better to go for a better naming IMHO.

skops/card/_model_card.py
```diff
@@ -289,9 +282,39 @@ def save(self, path: str | Path) -> None:
...
+
+        Notes
+        -----
+        The keys in model card metadata can be seen `here
+        <https://huggingface.co/docs/hub/models-cards#model-card-metadata>`__.
+        """
+        card = self._generate_card()
         card.save(path)
 
+    def render(self) -> str:
```
merveenoyan, line 306: I liked what you did here but it's actually `_generate_card` that does rendering (with `from_template`) so I'd call this something else like `render_to_str` to avoid confusion for future maintainers. Feel free to ignore if you don't agree 🙂

merveenoyan (approved):
Left nits, thanks a for handling this 🥺👉🏼👈🏼

skops/card/_model_card.py
BenjaminBossan: Yes, it doesn't generate the skops `Card`, it generates the modelcards `ModelCard` object. What would be a better name? I think `_generate_content` is a bit vague. I'm also not attached to `generate`, we can also use a plain `make` or something.

skops/card/_model_card.py
BenjaminBossan: > but it's actually _generate_card that does rendering (with from_template)

I agree that `_generate_card` does the heavy lifting, but from a user perspective, it doesn't really matter.

> so I'd call this something else like render_to_str to avoid confusion for future maintainers

I did think about `render_to_str`. IMO the name should be chosen with respect to the users, not so much maintainers, as it'll be a public method. `render_to_str` makes sense if we plan on adding other outputs and then have `render_to_foo`. If, instead, we want to support those other formats via arguments (`card.render(format="foo")`), I'd stick with the terse `render`. What is the more likely scenario?

skops/card/_model_card.py
merveenoyan: Since it's not a public method we could improve the docstring instead, I guess. Same with the below one.

BenjaminBossan: Happy to improve the docstring. Could you please let me know what I should clarify -- since I made the change, for me it's obvious ;)

> Same with the below one.

Which one do you mean?

skops/card/_model_card.py
```diff
@@ -401,9 +394,39 @@ def save(self, path: str | Path) -> None:
...
+        path: str, or Path
+            filepath to save your card.
+
+        Notes
+        -----
+        The keys in model card metadata can be seen `here
+        <https://huggingface.co/docs/hub/models-cards#model-card-metadata>`__.
+        """
+        card = self._generate_card()
         card.save(path)
```
adrinjalali, line 416: should this call `render`?

adrinjalali (commented):
I agree with @BenjaminBossan on how the methods are organized here. `_generate_card` is not something we'd document for users and they should never call it. So from their perspective, there's only `render` and `save`.

skops/card/_model_card.py
BenjaminBossan: You mean call `self.render` and then write the string directly to the file, instead of going through `card.save`?

adrinjalali: yeah that's a good question. I'm wondering if there _could_ be a difference between users calling save and saving the output of `render`, which wouldn't be ideal I think. WDYT?

BenjaminBossan: At the moment, both would amount to the same, but who knows what else could be added on the `modelcards` side in the future. Therefore, I would prefer users to use skops `Card.save`, which in turn should use `ModelCard.save`. But if a user wants to call `render` and save that, I guess we can't prevent them.

adrinjalali (commented):
LGTM. Merge conflicts though.

I'd wait maybe for @merveenoyan to have a final look.

BenjaminBossan:
Solved the merge conflict and wait for Merve's verdict.

adrinjalali:
Merging this one. @merveenoyan feel free to open an issue or a PR fixing things which need to be fixed when you're back :)

PR #98, opened 2022-08-12
TST Add tests for when pandas is not installed

BenjaminBossan:
Improves test coverage for the `try ... import pandas ...` parts.

Note: Instead of `try ... except ImportError`, we could also do `with contextlib.suppress(ImportError)` and skip the `except` part. I assume this was not used intentionally, so I left that code as is.

### DIFF
--- a/skops/hub_utils/tests/test_hf_hub.py
+++ b/skops/hub_utils/tests/test_hf_hub.py
@@ -4,7 +4,6 @@ import pickle
 import shutil
 import tempfile
 from pathlib import Path
-from unittest.mock import patch
 from uuid import uuid4
 
 import numpy as np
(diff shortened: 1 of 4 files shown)


BenjaminBossan:
Please review @skops-dev/maintainers

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -310,3 +311,26 @@ def test_get_column_names():
...
+def pandas_not_installed():
+    # patch import so that it raises an ImportError when trying to import
+    # pandas. This works because pandas is only imported lazily.
+    def mock_import(name, *args, **kwargs):
+        if name == "pandas":
+            raise ImportError
+        return __import__(name, *args, **kwargs)
+
+    with patch("builtins.__import__", side_effect=mock_import):
+        yield
```
adrinjalali, line 326: Aren't you using the same in the card tests? We could put this in a test utils kinda place and share between tests?

adrinjalali (commented):
Otherwise LGTM.

skops/hub_utils/tests/test_hf_hub.py
BenjaminBossan: Yes, I thought a bit of duplication is okay, but I now moved the fixture to `conftest.py`.

adrinjalali (approved):

PR #105, opened 2022-08-18
FEAT add method to call inference backend

adrinjalali:
This PR introduces a wrapper to make it convenient for users to get outputs for array like inputs.

I'm not sure about the name of the function.

I'll be adding tests.

@BenjaminBossan would you mind having a look and see if this looks okay as a start to you?

### DIFF
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -484,8 +484,10 @@ def download(
         shutil.rmtree(path=cached_folder)
 
 
-def get_output(repo_id: str, data: Any, token: Optional[str] = None) -> Any:
+def get_model_output(repo_id: str, data: Any, token: Optional[str] = None) -> Any:
     """Returns the output of the model using Hugging Face Hub's inference API.
+
+    See the :ref:`User Guide <hf_hub_inference>` for more details.
 
     Parameters
     ----------
@@ -494,11 +496,25 @@ def get_output(repo_id: str, data: Any, token: Optional[str] = None) -> Any:
         ``OWNER/REPO_NAME``.
 
     data: Any
-        The input to be given to the model.
+        The input to be given to the model. This can be a
+        :class:`pandas.DataFrame` or a :class:`numpy.ndarray`. If possible, you
+        should always pass a :class:`pandas.DataFrame` with correct column
+        names.
 
     token: str, optional
         The token to be used to call the inference API. Only required if the
         repository is private.
+
+    Returns
+    -------
+    output: numpy.ndarray
+        The output of the model.
+
+    Notes
+    -----
+    If there are warnings or exceptions during inference, this function raises
+    a :class:`RuntimeError` including the original errors and warnings
+    returned from the server.
     """
     model_info = HfApi().model_info(repo_id=repo_id, token=token)
     if not model_info.pipeline_tag:
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -8,10 +8,10 @@ import collections
 import json
 import shutil
 from pathlib import Path
-from typing import Any, List, MutableMapping, Union
+from typing import Any, List, MutableMapping, Optional, Union
 
 import numpy as np
-from huggingface_hub import HfApi, snapshot_download
+from huggingface_hub import HfApi, InferenceApi, snapshot_download
 from requests import HTTPError
 
 from ..utils.fixes import Literal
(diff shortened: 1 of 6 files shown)


adrinjalali:
Note: this cannot be tested or released until https://github.com/huggingface/huggingface_hub/pull/998 is released (unless we patch that variable in this library, but rather not to)

BenjaminBossan:
LGTM so far.

Regarding the naming, I have no strong opinion. Is there a reason to prefer "output" over "prediction"? At least when it comes to the planned features, we would always call `model.predict`, right? Not sure if we ever would want to have, say, `transform` or `decision_function`. Alternatively, we could use the word "inference", since we're using the inference API.

adrinjalali:
yeah that's why I didn't call it predict. I'm thinking  a feature we can add is to let users add a "predict_method" to the configuration file, and get `predict_proba` or `transform` instead of `predict` for instance.

BenjaminBossan:
In that case, it makes sense to not call it predict (except if that feature would be provided through separate functions).

adrinjalali:
I'm still not sure on that design though. But we'll get there when we start working on it.

adrinjalali:
The classifier test passes, the regression one fails, need to figure out why.

adrinjalali:
Seems like https://github.com/huggingface/api-inference-community/pull/83 is not deployed yet. Once that's done, the tests here should pass.

BenjaminBossan:
Please correct me if I misunderstand, but doesn't that check test if the prediction is 100% accurate? This may occur with classification but is almost impossible with regression.

adrinjalali:
so the issue was that I was comparing with the training data instead of the model output lol 🤦🏼 

Question: should we add this to the user guide? If we do, the user guide will take like 30 seconds at least more to run.

adrinjalali:
ready for review

skops/hub_utils/_hf_hub.py
```diff
@@ -482,3 +482,61 @@ def download(
...
+    except AttributeError:
+        # the input is not a pandas DataFrame
+        inputs = {f"x{i}": data[:, i] for i in range(data.shape[1])}
+        inputs = {"data": inputs}
+
+    res = InferenceApi(repo_id=repo_id, task=model_info.pipeline_tag, token=token)(
+        inputs=inputs
+    )
+
+    if isinstance(res, list):
```
BenjaminBossan, line 539: Can we rely on the assumption that if the output is a list, everything went well, else it didn't?

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -293,6 +312,68 @@ def test_push_download(
...
+        source=destination_path,
+        token=HF_HUB_TOKEN,
+        commit_message="test message",
+        create_remote=True,
+    )
+
+    X_test = data.data.head(5)
+    y_pred = model.predict(X_test)
+    output = get_model_output(repo_id, data=X_test, token=HF_HUB_TOKEN)
+    assert np.allclose(output, y_pred)
```
BenjaminBossan, line 370: nit: How about moving the `assert` to the end of the test, so that clean up is performed even if the assert fails? Even better would be a fixture or context manager that does the setup and cleanup but moving the assert would already help.

BenjaminBossan (commented):
Basically perfect for me. I have two minor comments, please take a look.

skops/hub_utils/_hf_hub.py
adrinjalali: Yes, on the inference side, we always raise when there's at least one warning or error: https://github.com/huggingface/api-inference-community/blob/5b9970bae8928019dd725f7f16832d9be753df3b/docker_images/sklearn/app/pipelines/tabular_classification.py#L55

skops/hub_utils/tests/test_hf_hub.py
adrinjalali: Didn't end up using fixtures for this cause I couldn't figure out how to do them easily, and the pytest docs are really sparse when it comes to using `request`.

skops/hub_utils/_hf_hub.py
BenjaminBossan: Okay, and we can be sure that if an error is raised, the API will never return a list. Ideally, I would wish for a response object with a status code but api inference doesn't seem to support that.

On the other hand, when the response _is ok_, we know it's a list because of how the `Pipeline`s are implemented. I guess this introduces some coupling, as future pipelines must always return a list as output too. For now it works, so we can proceed, but it doesn't feel very robust to me.

BenjaminBossan (approved):
LGTM.

Regarding the fixture, I agree that it would be overkill for this test.

If you want, I can merge, or do you want to wait for more reviews?

skops/hub_utils/_hf_hub.py
adrinjalali: The response comes as a json object, so if there are only values in it, it's a list. But I agree a more complex response or having warnings in the header would be better maybe. For now I'd say we can merge and we can change this if the logic on the backend changes. I don't think we can easily future proof this bit here.

PR #106, opened 2022-08-18
ENH Extend user guide to include classification report

### LINKED ISSUE #103: Add classification report to model card
We should expand the model card, but some expansions are achieved mostly by having a good example (at first at least).

from https://github.com/skops-dev/skops/issues/29#issuecomment-1216451647:

> [classification_report](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.classification_report.html)

> This is kinda covered with the `output_dict=True` argument + `add_table`, so we can consider it done (though an example would be nice to have for users).

This task can be achieved by adding this to our model card example.

BenjaminBossan:
Fixes #103 

## Description

It is currently already possible to add classification reports to the
model card, but it might not be completely obvious. Therefore, the model
card user guide now includes an example of how to add the classification
report.

Some of the texts in the user guide are rewritten to (hopefully) provide
more clarity and to remove some outdated info.

The model card for `plot_model_card.py` after the change in this PR can
be seen [here](https://huggingface.co/skops-ci/hf_hub_example-6ccd199d-47ce-4033-91b5-429131434fbe#classification-report).

## Implementation

As mentioned, the current code already supports adding the classification
report. However, it is unfortunately necessary to "massage" the report a bit
before it is shown in a useful way. First of all, `tabulate` cannot parse the
format of the output dict, so we need to go through a `DataFrame` first. Next, the
"accuracy" key has to be deleted, because it is only a single value, not a
collection, and would otherwise confuse pandas. And finally, the df needs to be
transposed and the index reset. Overall, I'm not sure of this is too hassle for
the user so that we should provide a utility function.

### DIFF
--- a/examples/plot_model_card.py
+++ b/examples/plot_model_card.py
@@ -160,7 +160,7 @@ clf_report = classification_report(
 # the correct format. This requires removing the "accuracy", which was added
 # above anyway.
 del clf_report["accuracy"]
-clf_report = pd.DataFrame(clf_report).T
+clf_report = pd.DataFrame(clf_report).T.reset_index()
 model_card.add_table(
     **{
         "Hyperparameter search results": cv_results,
--- a/examples/plot_model_card.py
+++ b/examples/plot_model_card.py
@@ -157,11 +157,14 @@ clf_report = classification_report(
     y_test, y_pred, output_dict=True, target_names=["malignant", "benign"]
 )
 # The classification report has to be transformed into a DataFrame first to have
-# the correct format.
+# the correct format. This requires removing the "accuracy", which was added
+# above anyway.
+del clf_report["accuracy"]
+clf_report = pd.DataFrame(clf_report).T
 model_card.add_table(
     **{
         "Hyperparameter search results": cv_results,
-        "Classification report": pd.DataFrame(clf_report),
+        "Classification report": clf_report,
     }
 )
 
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
@adrinjalali Ready for review.

adrinjalali:
This looks pretty good, but I'm wondering if we should put tables which have like more than N (maybe 10) rows in a details tag. The hyperparameter search table is rather long.

BenjaminBossan:
> This looks pretty good, but I'm wondering if we should put tables which have like more than N (maybe 10) rows in a details tag. The hyperparameter search table is rather long.

I had an idea regarding detail tags, maybe that would solve the issue (though in a separate PR): How about we add an additional argument to all `add_*` methods like `detail_tag=False`. If `True`, everything that is added is automatically enclosed in a detail tag. That would give the user more control over that, instead of setting arbitrary limits like 10 rows/columns. WDYT?

adrinjalali:
That sounds good to me, and having it in a separate PR sounds good too.

adrinjalali (approved):
I'm happy with this. What I have in mind is to have these in our examples, and see their usage, and then figure out a better API to make it easier for users to create a nice modelcard with minimal lines of code kinda thing. Shall I merge?

BenjaminBossan:
> That sounds good to me, and having it in a separate PR sounds good too.
Great, I'll create an issue.

> What I have in mind is to have these in our examples, and see their usage, and then figure out a better API to make it easier for users to create a nice modelcard with minimal lines of code kinda thing.

I assume this refers to my comment about how the classification report needs to be massaged to have the right format? If so, I agree.

> Shall I merge?
It can be merged from my side.

PR #108, opened 2022-08-19
ENH Add details tag in methods

### LINKED ISSUE #107: Add detail tag to new sections
Following up the discussion in #106:

> I had an idea regarding detail tags, maybe that would solve the issue (though in a separate PR): How about we add an additional argument to all `add_*` methods like `detail_tag=False`. If `True`, everything that is added is automatically enclosed in a detail tag.

Just to be clear, the section heading should be rendered normally but the section content would be folded.

BenjaminBossan:
Fixes #107

When passing `details_tag=True` for figures or tables, those are now
folded.

Note that this does not work for content added via the add method, since
that content relies on the existing template. Therefore, to change that,
the template has to be changed.

To see the results, look here:

* [Plot](https://huggingface.co/skops-ci/hf_hub_example-fcc9ddb0-1904-4bb6-ae49-e7fd1d2d589c#confusion-matrix)
* [Tables](https://huggingface.co/skops-ci/hf_hub_example-fcc9ddb0-1904-4bb6-ae49-e7fd1d2d589c#hyperparameter-search-results)

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -17,9 +17,8 @@ v0.2
   :pr:`94` by `Benjamin Bossan`_.
 - Make :meth:`skops.hub_utils.init` atomic. Now it doesn't leave a trace on the
   filesystem if it fails for some reason. :pr:`60` by `Adrin Jalali`_
-- When adding figures or tables, it's now possible to set ``details_tag=True``
-  to render the content inside a details tag (i.e. folded). :pr:`108` by
-  `Benjamin Bossan`_.
+- When adding figures or tables, it's now possible to set ``folded=True`` to
+  render the content inside a details tag. :pr:`108` by `Benjamin Bossan`_.
 
 v0.1
 ----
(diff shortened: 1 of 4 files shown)


BenjaminBossan:
@adrinjalali ready for review

adrinjalali (commented):
Thanks for the PR. I'm happy with the implementation, but I think the fact that the `details` tag does it, is an implementation detail. WDYT of calling the argument something like `expandable` instead?

BenjaminBossan:
> WDYT of calling the argument something like `expandable` instead?

Yes, we can rename the argument. I wanted to use something that would have 0% chance of being actually intended as a section title. If you think "expandable" fits the bill, we can take that. I also thought about "folded" (like "code folding"), just to throw something else out there.

adrinjalali:
yeah folded seems a better one to me.

adrinjalali (approved):

PR #109, opened 2022-08-22
ENH When model file is empty, warn before upload

### LINKED ISSUE #41: Warn users if the provided pickle file is empty
If users provide an empty pickle file, it's most probably by mistake, so we should warn them.

cc @BenjaminBossan

BenjaminBossan:
Fixes #41


adrinjalali (commented):
I'm not sure about the nits on the messages (but they're missing a dot at the end).

BenjaminBossan:
I committed your suggestions.

> but they're missing a dot at the end

I'm never sure about this one, error messages both with and without dot are used commonly.

adrinjalali (approved):

Issue #111, opened 2022-08-23
RFC Include generating file(s) in the repo


adrinjalali:
While thinking about https://github.com/skops-dev/skops/issues/110 I realized it'd be much easier to fix existing issues in repos if the file generating the model and the modelcard was included in the repo.

The question is, do we want to include the scripts generating them in the repo, and if yes, how.

cc @skops-dev/maintainers , @julien-c, @osanseviero , @lysandrejik

BenjaminBossan:
In general, I'm in favor of this, because it makes it easy to replicate everything. I would not go as far as to force this but it should be encouraged.

Maybe we can think of a standard. E.g. for models, there could be a `train.py` script that generates the model artifact, config, and model card when called (without arguments). The requirements file should be sufficient to run that script.

adrinjalali:
I agree that it shouldn't be a requirement, but I'm not sure of the API to let this happen.

BenjaminBossan:
> but I'm not sure of the API to let this happen.

You mean an API from the skops side to provide this functionality? I don't think it's really possible. My thought was rather to lead by example.

adrinjalali:
Yeah, I was thinking something like having an API which would take `__file__` and save it to the model folder, but I'm not sure how clean it is.

BenjaminBossan:
Hmm, wouldn't that result in users potentially uploading files that they may not have intended to be uploaded?

merveenoyan:
I think it makes sense and we should have it, just like they do with Space repositories (`app.py` file). I also agree with @BenjaminBossan that it might cause users to potentially upload files that they don't want in the repository. Also they might be putting a couple of API keys in their scripts or notebooks if it's generated by a notebook.
We could take the file during init and by default not have it inside. Maybe also add a warning for users to check if there's any variable that shouldn't be public.

adrinjalali:
So instead of adding it automatically, we could have a method like `hub_utils.add_file` which can also accept `__file__` as input value.

BenjaminBossan:
> So instead of adding it automatically, we could have a method like `hub_utils.add_file` which can also accept `__file__` as input value.

I'll take a look at that

adrinjalali:
I think we can close this one, and try to add generating files to existing repos, we should clean them up at some point anyway.

PR #123, opened 2022-08-31
FEA Utility function to add arbitrary files to the repo

BenjaminBossan:
Solves #111 

[This](https://huggingface.co/skops-ci/hf_hub_example-c57afd8c-0f9f-4dcc-9fd0-29325a03cf83/tree/main) is the outcome of using `plot_hf_hub.py` with one added line:

```python
hub_utils.add_files([__file__], dst=local_repo)
```

As can be seen, the script itself is now included in the upload.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -9,6 +9,11 @@ skops Changelog
     :depth: 1
     :local:
 
+v0.3
+----
+- Utility to function to add arbitrary files to be uploaded to the hub by using
+  :func:`.hub_utils.add_files`. :pr:`123` by `Benjamin Bossan`_.
+
 v0.2
 ----
 - Tables, e.g. cross-validation results, can now be added to model cards using
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -339,8 +339,7 @@ def add_files(*files: str | Path, dst: str | Path, exist_ok: bool = True) -> Non
     Parameters
     ----------
     *files : str or Path
-        The files to be added. If a file by that name already exists at target
-        location, it will be skipped.
+        The files to be added.
 
     dst : str or Path
         Path to the initialized repo, same as used during
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -9,6 +9,7 @@ import json
 import os
 import shutil
 import warnings
+from collections.abc import Sequence
 from pathlib import Path
 from typing import Any, List, MutableMapping, Optional, Union
 
@@ -325,6 +326,65 @@ def init(
     except Exception:
         shutil.rmtree(dst)
         raise
+
+
+def add_files(files: Sequence[str | Path], dst: str | Path) -> None:
+    """Add files to initialized repo.
+
+    After having called :func:`.hub_utils.init`, use this function to add
+    arbitrary files to be uploaded in addition to the model and model card.
+
+    In particular, it can be useful to upload the script itself that produces
+    those artifacts by calling ``hub_utils.add_files([__file__], dst=...)``.
+
+    Parameters
+    ----------
+    files : list of str or Path
+        The files to be added. If a file by that name already exists at target
+        location, it will be skipped.
+
+    dst : str or Path
+        Path to the initialized repo, same as used during
+        :func:`.hub_utils.init`.
+
+    Raises
+    ------
+    FileNotFoundError
+        When the target folder or the files to be added are not found.
+
+    TypeError
+        When the files are not passed correctly.
+
+    """
+    dst = Path(dst)
+    # check dst exists
+    if not dst.exists():
+        msg = f"Could not find '{dst}', did you run 'skops.hub_utils.init' first?"
+        raise FileNotFoundError(msg)
+
+    # validate input types
+    if isinstance(files, (str, Path)):
+        msg = (
+            "First argument must be a sequence of str or Path, got "
+            f"{type(files)} instead."
+        )
+        raise TypeError(msg)
+
+    src_files = [Path(file) for file in files]
+    # check that files exist
+    for file in src_files:
+        if not file.exists():
+            msg = f"File '{file}' could not be found."
+            raise FileNotFoundError(msg)
+
+    dst_files = [dst / Path(file).name for file in files]
+    for src_file, dst_file in zip(src_files, dst_files):
+        if dst_file.exists():
+            msg = f"File '{src_file.name}' already found at '{dst}', skipping."
+            warnings.warn(msg)
+            continue
+
+        shutil.copy2(src_file, dst_file)
 
 
 def update_env(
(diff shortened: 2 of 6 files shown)


BenjaminBossan:
@skops-dev/maintainers 

Could you help me figure out why the new tests are failing? The offending code seems to stem from this fixutre:

https://github.com/BenjaminBossan/skops/blob/adding-files/skops/hub_utils/tests/test_hf_hub.py#L497-L510

which results in the error `OSError: Model file '/tmp/.../model.pickle' does not exist.`

This seems odd to me because the tests pass locally and because I'm doing basically exactly the same as, e.g., [this test](https://github.com/skops-dev/skops/blob/ca853d4d49c64484e5dcd7c49e1bc03e68b1a0ba/skops/hub_utils/tests/test_hf_hub.py#L242-L254), which succeeds.

BenjaminBossan:
@skops-dev/maintainers The PR is ready for review now.

The error mentioned above was caused by `test_inference` doing some cleanup that would disrupt tests running later. We use some session-scoped fixtures, resulting in this dependency. My solution was to have `test_inference` run in a different directory in order for it not to interfere with other tests. Alternatively, we could reduce the scope of the used fixtures, but that would slow the tests down.

Regarding the failed codecov check: I have covered all new lines with tests, so this is a false positive. It seems that codecov thinks that my changes affect the coverage of `get_model_output` (see very last line):

https://app.codecov.io/gh/skops-dev/skops/compare/123/tree/skops/hub_utils/_hf_hub.py#D1L634

This is probably caused by this CI run not randomly failing when calling inference API, so it's a fluke.

docs/changes.rst
```diff
@@ -9,6 +9,11 @@ skops Changelog
     :depth: 1
     :local:
 
+v0.3
+----
+- Utility to function to add arbitrary files to be uploaded to the hub by using
```
```suggestion
- Utility function to add arbitrary files to be uploaded to the hub by using
```

skops/hub_utils/_hf_hub.py
```diff
@@ -327,6 +328,65 @@ def init(
         raise
 
 
+def add_files(files: Sequence[str | Path], dst: str | Path) -> None:
```
adrinjalali, line 331: we could also have the signature as:

```suggestion
def add_files(*files, dst):
```

makes it easier for people to just add many while adding one would be easier as:

```python
add_files(__file__, dst)
```

skops/hub_utils/_hf_hub.py
```diff
@@ -327,6 +328,65 @@ def init(
...
+        msg = f"Could not find '{dst}', did you run 'skops.hub_utils.init' first?"
+        raise FileNotFoundError(msg)
+
+    # validate input types
+    if isinstance(files, (str, Path)):
+        msg = (
+            "First argument must be a sequence of str or Path, got "
+            f"{type(files)} instead."
+        )
+        raise TypeError(msg)
```
adrinjalali, line 371: even if we stick with the current signature, I think we should accept a simple string or Path here.

skops/hub_utils/_hf_hub.py
```diff
@@ -327,6 +328,65 @@ def init(
...
+        if not file.exists():
+            msg = f"File '{file}' could not be found."
+            raise FileNotFoundError(msg)
+
+    dst_files = [dst / Path(file).name for file in files]
+    for src_file, dst_file in zip(src_files, dst_files):
+        if dst_file.exists():
+            msg = f"File '{src_file.name}' already found at '{dst}', skipping."
+            warnings.warn(msg)
+            continue
```
adrinjalali, line 385: we should have an argument defining the behavior on existing files.

adrinjalali (commented):
first look, haven't checked the tests yet.

BenjaminBossan:
@adrinjalali Yes, I was thinking about making files `*args` too, I changed it as suggested. Furthermore, I added an `exist_ok` argument (default `True`). If the file exists and `exist_ok=True`, I'll now overwrite it (instead of skipping like previously) and if `exist_ok=False` raise an error.

I was also thinking about adding directories but if we want that, let's do it in another PR.

skops/hub_utils/_hf_hub.py
```diff
@@ -327,6 +327,62 @@ def init(
...
+    arbitrary files to be uploaded in addition to the model and model card.
+
+    In particular, it can be useful to upload the script itself that produces
+    those artifacts by calling ``hub_utils.add_files([__file__], dst=...)``.
+
+    Parameters
+    ----------
+    *files : str or Path
+        The files to be added. If a file by that name already exists at target
+        location, it will be skipped.
```
adrinjalali, line 343:
```suggestion
        The files to be added.
```

skops/hub_utils/_hf_hub.py
```diff
@@ -327,6 +327,62 @@ def init(
...
+
+    dst : str or Path
+        Path to the initialized repo, same as used during
+        :func:`.hub_utils.init`.
+
+    exist_ok : bool (default=True)
+        Whether it's okay or not to add a file that already exists. If it's
+        okay, override the file, otherwise raise a ``FileExistsError``.
```
adrinjalali, line 351:
```suggestion
        Whether it's okay or not to add a file that already exists. If
        ``True``, override the files, otherwise raise a ``FileExistsError``.
```

skops/hub_utils/_hf_hub.py
```diff
@@ -327,6 +327,62 @@ def init(
...
+    ----------
+    *files : str or Path
+        The files to be added. If a file by that name already exists at target
+        location, it will be skipped.
+
+    dst : str or Path
+        Path to the initialized repo, same as used during
+        :func:`.hub_utils.init`.
+
+    exist_ok : bool (default=True)
```
adrinjalali, line 349: Probably set the default to ``False``, to avoid accidental overwrite?

skops/hub_utils/_hf_hub.py
```diff
@@ -327,6 +327,62 @@ def init(
...
+        Whether it's okay or not to add a file that already exists. If it's
+        okay, override the file, otherwise raise a ``FileExistsError``.
+
+    Raises
```
adrinjalali, line 353: nit: `copy2` might also raise due to permission issues, but I'm okay not listing them here.

adrinjalali (commented):
Otherwise LGTM.

skops/hub_utils/_hf_hub.py
BenjaminBossan: When the function is used elsewhere, we also don't mention that error. Not sure if we should document any possible error the function could raise or only those we explicitly raise in the function body?

skops/hub_utils/_hf_hub.py
BenjaminBossan: I think I slightly prefer the default to be `True`. That way, calling it multiple times is fine (idempotency).

BenjaminBossan: Not really sure if there is a Python convention for `exist_ok`: https://docs.python.org/3/library/pathlib.html#pathlib.Path.touch

Anyway, I changed it, as mentioned it was only a slight preference.

PR #124, opened 2022-09-01
MNT Remove redundant test in CI

BenjaminBossan:
As discussed, the "Tests" and "Doc Tests" tasks are redundant. Since CI
is already quite slow as is, we really don't want to run all the tests
twice. Therefore, the "Doc Tests" task is removed.

Ideally, we would have one task for only doctests and one for unit tests
but AFAICT, there is no option to _only_ run doctests.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -48,7 +48,7 @@ jobs:
       env:
         SUPER_SECRET: ${{ secrets.HF_HUB_TOKEN }}
       run: |
-        python -m pytest -s --cov=skops --cov-report=xml --doctest-modules -v skops/
+        python -m pytest -s -v --cov-report=xml skops/
 
     - name: Mypy
       run: mypy --config-file pyproject.toml skops
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -48,13 +48,7 @@ jobs:
       env:
         SUPER_SECRET: ${{ secrets.HF_HUB_TOKEN }}
       run: |
-        python -m pytest -s --cov=skops --cov-report=xml -v skops/
-
-    - name: Doc Tests
-      env:
-        SUPER_SECRET: ${{ secrets.HF_HUB_TOKEN }}
-      run: |
-        python -m pytest -s --cov=skops --doctest-modules skops/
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
@skops-dev/maintainers Ready for review (disregard false codecov complaint)

merveenoyan (approved):
Nice catch 🙂

.github/workflows/build-test.yml
```diff
@@ -48,13 +48,7 @@ jobs:
...
         SUPER_SECRET: ${{ secrets.HF_HUB_TOKEN }}
       run: |
-        python -m pytest -s --cov=skops --cov-report=xml -v skops/
-
-    - name: Doc Tests
-      env:
-        SUPER_SECRET: ${{ secrets.HF_HUB_TOKEN }}
-      run: |
-        python -m pytest -s --cov=skops --doctest-modules skops/
+        python -m pytest -s --cov=skops --cov-report=xml --doctest-modules -v skops/
```
adrinjalali, line 51: Aren't the options redundant with here?

https://github.com/skops-dev/skops/blob/ca853d4d49c64484e5dcd7c49e1bc03e68b1a0ba/pyproject.toml#L9-L17

BenjaminBossan: Yes they are. I thought about removing the options there but then was unsure for the following reason: Is the main purpose of pyproject.toml to be for local development or for CI? If it's for the former, and we make a change that should only affect local development, we don't want to accidentally affect CI too. Therefore, I chose to have the options explicitly in CI (I assume that explicitly setting arguments takes precedence over the config file). If you think this distinction is not relevant, I can remove all redundant options.

adrinjalali: Interesting point of view. To me the point was to have it easier for people to have a local env more similar to the CI, so that the same tests are run with less effort. I personally added them there because I wanted to have them in the CI (definitely true for `--doctest-modules`).

BenjaminBossan: This is definitely a decision we can make -- running locally should be as close as possible to CI and the pyproject.toml should be the source of truth for that. It's just good to make that assumption explicit, so that we can be aware of that for the future.

adrinjalali (approved):

PR #131, opened 2022-09-06
CI Reduce CI timeout from 60 to 15 min

BenjaminBossan:
After checking a couple of runs, none that took longer than 12 min ever
finished. Therefore, let's cap the timeout at 15 min for now.

Later, if/when we factor out the inference tests, we may decrease the
timeout even further.


BenjaminBossan:

adrinjalali (approved):
let's go with this and tune after if needed. Thanks @BenjaminBossan

Issue #126, opened 2022-09-05
Autogenerate Get Started section in the model card.


adrinjalali:
We should figure out a way to auto-generate that section so that users don't have to write code like this:

```python
get_started_code = (
    "import pickle\nwith open(pkl_filename, 'rb') as file:\n    clf = pickle.load(file)"
)
```

merveenoyan:
@adrinjalali they're pretty standard so getting filename and generating the part should suffice.

PR #132, opened 2022-09-06
DOC Autogeneration of getting started code

merveenoyan:
This PR solves #126.

I also made a fixture out of metadata creation from config file so that it can be used repeatedly. Let me know if this is not good.

### DIFF
--- a/skops/card/default_template.md
+++ b/skops/card/default_template.md
@@ -41,16 +41,9 @@ You can find the details about evaluation process and the evaluation results.
 
 Use the code below to get started with the model.
 
-<details>
-<summary> Click to expand </summary>
-
 ```python
 {{ get_started_code | default("[More Information Needed]", true)}}
 ```
-
-</details>
-
-
 
 
 # Model Card Authors


merveenoyan:
Model card created from this branch, the code is automatically generated: https://huggingface.co/merve/test-code-generation

skops/card/_model_card.py
```diff
@@ -380,6 +380,13 @@ def _generate_card(self) -> ModelCard:
         # add evaluation results
 
         template_sections = copy.deepcopy(self._template_sections)
+        if self.metadata:
+            model_file = self.metadata.to_dict().get("model_file", None)
```
BenjaminBossan, line 384: nit: not required

```suggestion
            model_file = self.metadata.to_dict().get("model_file")
```

BenjaminBossan (approved):
Overall LGTM.

One minor thing I wonder: Do we want to assume that users will always load the metadata from config or are there good reasons not to? If there are reasons not to, perhaps we should keep an example of manually adding the `get_started_code`.

skops/card/_model_card.py
merveenoyan: They can already add the section if they want to with `.add()`, it looks a bit ugly IMO that's why I removed it.

merveenoyan:
@BenjaminBossan I think it looks a bit ugly and our example already includes metadata from config file so that's why I wanted to remove it (it will be a duplication if we add that too).

BenjaminBossan:
Yes, I'm aware that in the example, it doesn't make sense. What I meant is to document it some other place so that users have a chance of discovering it (as is, the only way to discover the feature is by reading the template).

If we can assume, however, that the vast majority of users create the section automatically, then maybe it's not necessary.

merveenoyan:
@adrinjalali what do you suggest?

adrinjalali:
I think since w/o the config file people won't really would know what to do with the model file, it's safe to assume the config file is there, especially for an example in the skops library, which generates that file.

skops/card/_model_card.py
```diff
@@ -380,6 +380,13 @@ def _generate_card(self) -> ModelCard:
         # add evaluation results
 
         template_sections = copy.deepcopy(self._template_sections)
+        if self.metadata:
+            model_file = self.metadata.to_dict().get("model_file")
+            if model_file:
+                template_sections["get_started_code"] = (
+                    f"import pickle\nwith open({model_file}, 'rb') as file:\n    clf ="
+                    " pickle.load(file)"
```
adrinjalali, line 388: I think we can also add a `predict` example here, with the data available in the config file.

adrinjalali (commented):
otherwise LGTM

skops/card/_model_card.py
merveenoyan: Addressed. new version of card: https://huggingface.co/merve/test-code-generation

adrinjalali (commented):
Thanks, this LGTM. I'd suggest at the same time we also remove the `<details>` tag around it, this is something I'd say should be visible to users.

merveenoyan:
@adrinjalali done, new model card: https://huggingface.co/merve/test-code-generation

adrinjalali (approved):
That looks gorgeous, thanks for giving an example one on the Hug @merveenoyan , it really does help with the review.

PR #136, opened 2022-09-14
FEA update_env implementation

### LINKED ISSUE #133: implement existing empty hub_utils.update_env
I just noticed this method is still not implemented, but exists in the docs and the repo.

merveenoyan:
Closes #133 
This one replaces existing requirements with the new one user gives. Let me know if you'd rather have something that adds new dependencies on top of old requirements instead (or simply changes versions).

### DIFF
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -402,7 +402,14 @@ def update_env(
     -------
     None
     """
-    pass
+
+    with open(Path(path) / "config.json") as f:
+        config = json.load(f)
+
+    config["sklearn"]["environment"] = requirements
+
+    with open(Path(path) / "config.json", mode="w") as f:
+        json.dump(config, f, sort_keys=True, indent=4)
 
 
 def push(
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -210,8 +210,7 @@ def _create_config(
         else:
             raise ValueError("The data needs to be a list of strings.")
 
-    with open(Path(dst) / "config.json", mode="w") as f:
-        json.dump(config, f, sort_keys=True, indent=4)
+    dump_json(Path(dst) / "config.json", config)
 
 
 def _check_model_file(path: str | Path) -> Path:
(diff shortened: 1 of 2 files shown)


merveenoyan:
mypy is complaining: https://github.com/python/mypy/issues/13627

skops/hub_utils/_hf_hub.py
```diff
@@ -402,7 +402,14 @@ def update_env(
...
     """
-    pass
+
+    with open(Path(path) / "config.json") as f:
+        config = json.load(f)
+
+    config["sklearn"]["environment"] = requirements
+
+    with open(Path(path) / "config.json", mode="w") as f:
+        json.dump(config, f, sort_keys=True, indent=4)
```
BenjaminBossan, line 412: Just throwing an idea out there: Do we want to factor this line out into a small utility function, since it's used twice now (other being `_create_config`)?

skops/hub_utils/_hf_hub.py
```diff
@@ -402,7 +402,14 @@ def update_env(
     -------
     None
     """
-    pass
+
```
BenjaminBossan, line 405:
```suggestion
```

Also, can we remove the "Returns" section from the docstring, since nothing is returned?

BenjaminBossan (commented):
Looks good to me, thanks for taking this. I made two minor comments, but feel free to ignore them.

Regarding the mypy error, it sounds like they're working on a fix. Let's maybe just wait for the fix? I don't believe it's worth it to take extra measures here for an error that's (hopefully) going away very quickly.

skops/hub_utils/_hf_hub.py
merveenoyan: I just did!

skops/hub_utils/_hf_hub.py
merveenoyan: Done.

BenjaminBossan (approved):
LGTM. Maybe give this 1 week to see if the mypy issue is solved by then?

adrinjalali:
Is there an open issue on mypy side? If not, we can just add type: ignore there.

merveenoyan:
@adrinjalali there is: `https://github.com/python/mypy/issues/13627` but I don't want to reference their issue here :D

BenjaminBossan:
> but I don't want to reference their issue here :D

Note that pasting the link _did_ create a reference in that issue ^^ I think they changed something on GH to enable this "feature".

PR #137, opened 2022-09-15
CI pin python for mypy issue

adrinjalali:
This should fix the mypy issue for now.


BenjaminBossan (approved):
I would have hoped for a quick mypy release to address the error. But probably we should pin for now. We just have to make sure to observe the issue and unpin if it is resolved.

adrinjalali:
The issue doesn't even say the fix will be included in the coming release, it might come in the one after, so I think we should pin for now. I still haven't been sold on the idea of `mypy` lol.

Issue #104, opened 2022-08-17
Add feature importance to the model card


adrinjalali:
We should add some sort of a permutation importance to the model card example.

Later we can add this more as a functionality, but for now it can be just in the example with the existing methods we have.

ref: https://github.com/skops-dev/skops/issues/29#issuecomment-1216451647

PR #142, opened 2022-09-16
ENH permutation importance

merveenoyan:
I implemented permutation importance, will write tests if you like where it goes.

I have two problems:
- Are we okay with `matplotlib` being a dependency here? I'll look for something better. Same applies for `pandas`. 
- I'm looking for a more standard way of getting feature names. For some base estimators or pipelines it varies, from `feature_names_in_` to `get_feature_names_out()`.

The example script looks like below (I'll change layout of plot to fit inside the plot itself):
Ekran Resmi 2022-09-16 19 37 37


skops/card/_model_card.py
```diff
@@ -366,6 +368,33 @@ def add_metrics(self, **kwargs: str) -> "Card":
             self._eval_results[metric] = value
         return self
 
+    def add_feature_importances(self, feature_importances) -> "Card":
+        """Visualize permutation importance.
+
+        Parameters
+        ----------
+        self : object
```
BenjaminBossan, line 376: `self` does not need to be documented.

skops/card/_model_card.py
```diff
@@ -10,6 +10,8 @@
 from reprlib import Repr
 from typing import Any, Optional, Union
 
+import matplotlib.pyplot as plt
+import pandas as pd
```
BenjaminBossan, line 14: Here it is assumed that pandas is installed. However, pandas is not a strict dependency for skops. I think we specifically didn't want to add pandas because it is a "fat" dependency.

BenjaminBossan (changes requested):
I have some specific comments for this PR and then a more general discussion.

For this PR, I would like to see some tests added. Also, I have a few comments for some lines of code, please check.

For the more general discussion, I want to ask the following question: Do we want to design skops model cards such that we add a bunch of narrow methods such as `add_feature_importances` or do we just want to have general methods like `add_plot` and require the user to do the work inside the method themselves?

E.g. for adding cross-validation results, we have put the burden on the user to process the data and then use the generic `add_table` method:

https://github.com/skops-dev/skops/blob/cfea7cb5a741a86365eaa9bd4d56a86cc46cbea5/examples/plot_model_card.py#L151-L166

For consistency, this should either be its own method,`add_cv_results`, or `add_feature_importances` should not be its own method.

The arguments for _not_ providing specific methods are:

- Less work for us (though we should still provide examples)
- More flexibility for the user. E.g. here, the user cannot change the parameters of the plot, e.g. the what the title should be or what the name of the plot should be.
- Don't need to worry about pandas dependency etc.

The argument for providing the specific methods:

- Easier for the user if they don't want to modify the defaults.
- Easier to discover than just providing examples.

skops/card/_model_card.py
merveenoyan: Yes I thought of it as well, I will try to plot without it.

merveenoyan:
@BenjaminBossan I had a quick huddle last week with @adrinjalali where we discussed this and he told me to create a method to add feature importance graphs. I can remove this and add it to model card example instead.

merveenoyan:
@BenjaminBossan I sort of agree and think the dependencies are a bit much if we want to move forward. Would you like to wait for @adrinjalali to come back from his vacay to make a decision?

examples/plot_model_card.py
```diff
@@ -148,6 +149,9 @@
 disp.figure_.savefig(Path(local_repo) / "confusion_matrix.png")
 model_card.add_plot(**{"Confusion matrix": "confusion_matrix.png"})
 
+result = permutation_importance(model, X_test, y_test, n_repeats=10)
+model_card.add_feature_importances(result)
```
adrinjalali, line 153:
```suggestion
inportances = permutation_importance(model, X_test, y_test, n_repeats=10)
model_card.add_feature_importances(importances)
```

skops/card/_model_card.py
```diff
@@ -366,6 +368,33 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        importances = pd.DataFrame(
+            feature_importances.importances.T,
+            columns=self.model.feature_names_in_,
+        )
+        ax = importances.plot.box(vert=False, whis=10)
+        ax.set_title("Permutation Importances")
+        ax.set_xlabel("Decrease in accuracy score")
+        plt.savefig("feature_importances.png")
+        self.add_plot(**{"Feature Importances": "feature_importances.png"})
+        return self
```
adrinjalali, line 396: here pandas is only used to handle plotting, I think we can avoid this and create the plot ourself and remove the pandas import here.

skops/card/_model_card.py
```diff
@@ -366,6 +368,33 @@ def add_metrics(self, **kwargs: str) -> "Card":
             self._eval_results[metric] = value
         return self
 
+    def add_feature_importances(self, feature_importances) -> "Card":
+        """Visualize permutation importance.
```
adrinjalali, line 372: I think in terms of API, it would make sense to accept importances in a certain way, like accepting both a dict and a pandas dataframe, and then add it to the modelcard.

It doesn't have to be permutation importance.

adrinjalali (commented):
Thanks @merveenoyan

examples/plot_model_card.py
```diff
@@ -148,6 +149,9 @@
 disp.figure_.savefig(Path(local_repo) / "confusion_matrix.png")
 model_card.add_plot(**{"Confusion matrix": "confusion_matrix.png"})
 
+inportances = permutation_importance(model, X_test, y_test, n_repeats=10)
```
BenjaminBossan, line 152:
```suggestion
importances = permutation_importance(model, X_test, y_test, n_repeats=10)
```

skops/card/_model_card.py
```diff
@@ -373,6 +374,34 @@ def add_metrics(self, **kwargs: str) -> "Card":
             self._eval_results[metric] = value
         return self
 
+    def add_feature_importances(self, feature_importances, data) -> "Card":
```
BenjaminBossan, line 377: I wonder about the naming. The method uses "feature_importances", so intuitively, I'd think I can pass something like `est = RandomForestClassifier().fit(X, y); card.add_feature_importances(est.feature_importances_)` but this won't work. If it only works with permutation importances, should the method be renamed accordingly?

Alternatively, we could try to support passing `est.feature_importances_`, as it would be quite convenient, though the meaning of the plot would change.

skops/card/_model_card.py
```diff
@@ -373,6 +374,34 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        """Visualize permutation importance.
+
+        Parameters
+        ----------
+        feature_importances : sklearn.utils.Bunch
+            Output of sklearn.inspection.permutation_importance()
+
+        data :
```
BenjaminBossan, line 385: Here it looks like `data` has to be a pandas df. But in the end, only the column names are used, right? So how about we just pass the column names here? Also, I would make the argument optional, in case people only have a numpy array. Then the features would just be called `x0`, `x1`, etc. WDYT?

skops/card/_model_card.py
```diff
@@ -373,6 +374,34 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        """
+        sorted_importances_idx = feature_importances.importances_mean.argsort()
+        fig, ax = plt.subplots()
+        ax.boxplot(
+            x=feature_importances.importances[sorted_importances_idx].T,
+            labels=data.columns[sorted_importances_idx],
+            vert=False,
+        )
+        ax.set_title("Permutation Importances")
+        ax.set_xlabel("Decrease in Accuracy Score")
```
BenjaminBossan, line 400: "Accuracy" is not quite correct, right? It can be any score, really. So maybe just "Decrease in Score" or "Deterioration of Score"?

skops/card/_model_card.py
```diff
@@ -373,6 +374,34 @@ def add_metrics(self, **kwargs: str) -> "Card":
             self._eval_results[metric] = value
         return self
 
+    def add_feature_importances(self, feature_importances, data) -> "Card":
+        """Visualize permutation importance.
```
BenjaminBossan, line 378: I think the docstring should also mention that a `feature_importances.png` is created.

BenjaminBossan (commented):
I have a couple of comments, please take a look.

Also, I have a more general comment:

What do we do if a user wants to add more than 1 permutation importance graph? E.g. one for accuracy and one for recall? Right now, this would not work because the file name `feature_importances.png` would be re-used, so we would get the same graph twice. Also, the section name is hard-coded, so the two sections would have the exact same name. I think we should have a solution for this.

skops/card/_model_card.py
merveenoyan: I changed this to have columns instead.

skops/card/_model_card.py
merveenoyan: I edited it.

skops/card/_model_card.py
merveenoyan: I edited this. the `plot_name` is shared by both the title on plot (to avoid ambiguity) and the name of the plot on the markdown so people can pass what they want.

skops/card/_model_card.py
merveenoyan: I renamed it to be permutation importance specifically.

merveenoyan:
Fixes #104

skops/card/_model_card.py
```diff
@@ -373,6 +374,47 @@ def add_metrics(self, **kwargs: str) -> "Card":
             self._eval_results[metric] = value
         return self
 
+    def add_permutation_importances(
+        self, feature_importances, columns, plot_file, plot_name
```
BenjaminBossan, line 378: I think you forgot to set the default argument for `plot_file` and `plot_name`, which is why lines 414 and 415 are not triggered. However, instead of setting them to `None`, I would just put the actual values, i.e. `plot_file="feature_importances.png"` and `plot_name="Permutation importances"`.

skops/card/_model_card.py
```diff
@@ -373,6 +374,47 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        feature_importances : sklearn.utils.Bunch
+            Output of sklearn.inspection.permutation_importance()
+
+        columns :
+            Column names of the data used to generate importances.
+
+        plot_file :
+            Filename for the plot.
+
+        plot_name :
```
BenjaminBossan, line 393: Docstring is missing the type information.

skops/card/_model_card.py
```diff
@@ -373,6 +374,47 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+            Name of the plot.
+
+        Returns
+        -------
+        self : object
+            Card object.
+        """
+        sorted_importances_idx = feature_importances.importances_mean.argsort()
```
BenjaminBossan, line 401: Here and below, you still use the variable name `feature_importances`, which should be `permutation_importances`.

skops/card/tests/test_card.py
```diff
@@ -169,6 +170,19 @@ def test_add_plot(destination_path, model_card):
     assert "![fig1](fig1.png)" in model_card
 
 
+def test_permutation_importances(
```
BenjaminBossan, line 173: Can we get a test with 2 different permutation importance plots?

BenjaminBossan (changes requested):
Good work, by adding the new arguments, my concerns could be addressed. There are still a few improvements, please see my comments. Also, I think it would be a good idea to update the `plot_model_card.py` and perhaps also add a sentence to `docs/model_card.rst`.

merveenoyan:
> oh I love how testing on windows machines 🙂🙂🙂🙂🙂🙂

- said by no one, ever

```
C:\\\\Users\\\\RUNNER~1\\\\AppData\\\\Local\\\\Temp\\\\skops-test180xjo68\\\\importance.png

C:\\\\Users\\\\RUNNER~1\\\\AppData\\\\Local\\\\Temp\\\\skops-test180xjo68/importance.png
```

skops/card/_model_card.py
```diff
@@ -10,6 +10,7 @@
 from reprlib import Repr
 from typing import Any, Optional, Union
 
+import matplotlib.pyplot as plt
```
BenjaminBossan, line 13: I don't think we can do that, since matplotlib is not an install dependency. We could add it but I'd prefer not to. So for this feature to still work, we would need to make the import local (i.e. in the method itself), which is also not super nice. @adrinjalali what's your opinion?

skops/card/_model_card.py
```diff
@@ -373,6 +374,48 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        columns,
+        plot_file="permutation_importances.png",
+        plot_name="Permutation Importances",
+    ) -> "Card":
+        """Plots permutation importance and saves it to model card.
+
+        Parameters
+        ----------
+        permutation_importances : sklearn.utils.Bunch
+            Output of sklearn.inspection.permutation_importance()
```
BenjaminBossan, line 389:
```suggestion
            Output of :func:`sklearn.inspection.permutation_importance`.
```

skops/card/_model_card.py
```diff
@@ -373,6 +374,48 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        plot_name : str, or Path
+            Name of the plot.
+
+        Returns
+        -------
+        self : object
+            Card object.
+        """
+        sorted_importances_idx = permutation_importances.importances_mean.argsort()
+        fig, ax = plt.subplots()
```
BenjaminBossan, line 406:
```suggestion
        _, ax = plt.subplots()
```

BenjaminBossan (commented):
I only have some small comments, except the one about matplotlib, which might require some further discussion.

skops/card/_model_card.py
```diff
@@ -373,6 +374,48 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        plot_name="Permutation Importances",
+    ) -> "Card":
+        """Plots permutation importance and saves it to model card.
+
+        Parameters
+        ----------
+        permutation_importances : sklearn.utils.Bunch
+            Output of sklearn.inspection.permutation_importance()
+
+        columns : Index
```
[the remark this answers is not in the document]
BenjaminBossan: > wondering how str would work though?

`card.add_permutation_importances(..., columns=["my-feat0", "my-feat1", ...])`? This is assuming that the users know and have a name for their features.

skops/card/_model_card.py
BenjaminBossan: Yes, better than just failing when the import fails

skops/card/_model_card.py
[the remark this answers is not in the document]
merveenoyan: but they'll align it themselves with importances, do they know that too? 😅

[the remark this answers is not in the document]
BenjaminBossan: Do they? I thought this is what is achieved by this line:

`labels=columns[sorted_importances_idx]`

or do I miss something?

Btw., `labels` might be the better variable name than `columns`.

skops/card/_model_card.py
adrinjalali: Yes, we shouldn't import in the module level, we should just try to import in the method, and raise a meaningful error if it fails.

skops/card/_model_card.py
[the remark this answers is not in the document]
adrinjalali: I don't think we should ask for a pandas.Index. That's not really user friendly. I'd rather accept a list of strings, and check if the length of the list is the same as the number of columns we have in importances.

skops/card/_model_card.py
merveenoyan: I just edited it 🙂

skops/card/_model_card.py
```diff
@@ -402,6 +402,15 @@ def add_permutation_importances(
         self : object
             Card object.
         """
+        try:
+            metadata.version("matplotlib")
```
adrinjalali, line 406: instead, just try to import here rather than getting the version, and remove the import a few lines after.

skops/card/tests/test_card.py
```diff
@@ -169,6 +173,70 @@ def test_add_plot(destination_path, model_card):
...
+def test_matplotlib_dependency_error(
+    iris_estimator, iris_data, model_card, destination_path
+):
+    with mock.patch.dict(sys.modules):
+        sys.modules["matplotlib"] = None
+        X, y = iris_data
+        result = permutation_importance(
+            iris_estimator, X, y, n_repeats=10, random_state=42, n_jobs=2
+        )
+        with pytest.raises(ModuleNotFoundError):
```
BenjaminBossan, line 203: Ideally, also check the error message, to ensure that the specific error is indeed raised.

BenjaminBossan (commented):
Only a minor comment and what Adrin said, the rest looks good

skops/card/_model_card.py
```diff
@@ -373,6 +374,55 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        sorted_importances_idx = permutation_importances.importances_mean.argsort()
+        _, ax = plt.subplots()
+        ax.boxplot(
+            x=permutation_importances.importances[sorted_importances_idx].T,
+            labels=columns[sorted_importances_idx],
+            vert=False,
+        )
+        ax.set_title(plot_name)
+        ax.set_xlabel("Decrease in Score")
+        plt.savefig(plot_file)
```
BenjaminBossan, line 421: Not sure if we should care or not, but I wonder if we should check if a file by that name already exists and maybe warn the user? If they re-create the same plot, I guess it's annoying to get a warning. But what if they already happen to have a `permutation_importances.png` in the folder? Then we would just overwrite that. Really not sure...

BenjaminBossan (commented):
I just have a small comment below, but I think in general the feature should be ready to go. Before that, however, could you please merge the current main and check that the CI errors go away?

skops/card/_model_card.py
adrinjalali: yes, we should have a flag, where by default we raise it the file exists, and only overwrite if the flag is True: `overwrite=True/False`

skops/card/_model_card.py
```diff
@@ -375,6 +376,63 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        self : object
+            Card object.
+        """
+        try:
+            import matplotlib.pyplot as plt
+        except metadata.PackageNotFoundError:
+            raise ModuleNotFoundError(
+                "This feature requires matplotlib to be installed."
+            )
+        if str(plot_file) in self.render() and overwrite is False:
```
BenjaminBossan, line 417: Could you please explain why we need to check `str(plot_file) in self.render()`?

skops/card/tests/test_card.py
```diff
@@ -169,6 +172,126 @@ def test_add_plot(destination_path, model_card):
...
+        else:
+            return orig_import(name, *args)
+
+    X, y = iris_data
+    result = permutation_importance(
+        iris_estimator, X, y, n_repeats=10, random_state=42, n_jobs=2
+    )
+
+    with mock.patch("builtins.__import__", side_effect=mock_import):
+        with pytest.raises(ImportError, match="cannot import name*"):
```
BenjaminBossan, line 210: Hmm, I'm not sure if the test does what it's supposed to do. First, I don't understand why we need two `pytest.raises`. Second, I would have thought that we need to raise `ModuleNotFoundError` in the mock for matplotlib. Third the `except` in the actual function body appears to not be reached at all according to coverage:

> Added line #L414 was not covered by tests

Note that one problem with implementing this test properly could be that `patch` is applied too late. This can happen if matplotlib was already imported at the time that the patch is applied, since Python will not attempt to be imported again, if I'm not mistaken. One way to find this out is if the test behaves the same when run in isolation (i.e. `pytest ... -k test_matplotlib_dependency_error` vs running all tests).

BenjaminBossan (changes requested):
Thanks for addressing the previous comments. There is a merge conflict with main, could you please resolve it? I also have some questions, please take a look at my comments.

skops/card/_model_card.py
merveenoyan: none of the Card attributes point to the path to the repository where we could look for plot (or even if there's a plot file, it might not be in the model card) so I'm checking if it exists in the card itself. if so, we have to overwrite.

merveenoyan:
@BenjaminBossan I can't see merge conflict in the diff, I'd appreciate it if you could point out.

skops/card/_model_card.py
BenjaminBossan: Can we not check if `plot_file` exists on the file system, so basically `if Path(plot_file).exists() and not overwrite: ...`? Or am I missing something?

skops/card/tests/test_card.py
adrinjalali: we probably should just create a utility function like `import_or_raise`, and test that once, and in other methods just use that one.

merveenoyan:
I was hitting recursion with my solution so Adrin swapped it with pytest based one. (we couldn't find why mine wouldn't work)

merveenoyan:
I checked my local environment and how GHActions is setup yet I can't figure out why it doesn't raise the error here meanwhile it passes on my local (and Adrin's local too) I don't get it.

merveenoyan:
@BenjaminBossan can you take a look? Can't ask for a review again

merveenoyan:
@BenjaminBossan codecov is failing :/

BenjaminBossan:
@merveenoyan again, codecov worked early in the day and now starts failing...

skops/card/_model_card.py
```diff
@@ -203,7 +204,7 @@ def split_subsection_names(key: str) -> list[str]:
 
 
 def _getting_started_code(
-    file_name: str, model_format: Literal["pickle", "skops"], indent="    "
+    file_name: str, model_format: Literal["pickle", "skops"], indent="    "  # type: ignore
```
BenjaminBossan, line 207:
```suggestion
    file_name: str, model_format: Literal["pickle", "skops"], indent: str = "    "
```

skops/card/_model_card.py
```diff
@@ -445,7 +446,7 @@ def __init__(
         model,
         model_diagram: bool = True,
         metadata: ModelCardData | None = None,
-        template: Literal["skops"] | dict[str, str] | None = "skops",
+        template: Literal["skops"] | dict[str, str] | None = "skops",  # type: ignore
```
BenjaminBossan, line 449:
```suggestion
        template: Literal["skops"] | dict[str, str] | None = "skops",
```

skops/card/_model_card.py
```diff
@@ -848,7 +849,7 @@ def add_get_started_code(
         section: str | None = None,
         description: str | None = None,
         file_name: str | None = None,
-        model_format: Literal["pickle", "skops"] | None = None,
+        model_format: Literal["pickle", "skops"] | None = None,  # type: ignore
```
BenjaminBossan, line 852:
```suggestion
        model_format: Literal["pickle", "skops"] | None = None,
```

skops/card/_model_card.py
```diff
@@ -929,7 +930,7 @@ def _add_get_started_code(
         self,
         section: str,
         file_name: str,
-        model_format: Literal["pickle", "skops"],
+        model_format: Literal["pickle", "skops"],  # type: ignore
```
BenjaminBossan, line 933:
```suggestion
        model_format: Literal["pickle", "skops"],
```

skops/card/_model_card.py
```diff
@@ -1077,11 +1078,63 @@ def add_metrics(
...
         self._add_metrics(section, self._metrics, description=description)
         return self
 
+    def add_permutation_importances(
+        self,
+        permutation_importances,
+        columns,
+        plot_file="permutation_importances.png",
+        plot_name="Permutation Importances",
+        overwrite=False,
```
BenjaminBossan, line 1091:
```suggestion
        columns: Sequence[str],
        plot_file: str = "permutation_importances.png",
        plot_name: str = "Permutation Importances",
        overwrite: bool = False,
```

skops/card/_model_card.py
```diff
@@ -1077,11 +1078,63 @@ def add_metrics(
...
+        permutation_importances : sklearn.utils.Bunch
+            Output of :func:`sklearn.inspection.permutation_importance`.
+
+        columns : str, list or pandas.Index
+            Column names of the data used to generate importances.
+
+        plot_file : str
+            Filename for the plot.
+
+        plot_name : str, or Path
```
BenjaminBossan, line 1106:
```suggestion
        plot_name : str
```

skops/card/_model_card.py
```diff
@@ -1077,11 +1078,63 @@ def add_metrics(
...
+            Name of the plot.
+
+        overwrite : bool
```
BenjaminBossan, line 1109:
```suggestion
        overwrite : bool (default=False)
```

skops/card/_model_card.py
```diff
@@ -1077,11 +1078,63 @@ def add_metrics(
...
+            Whether to overwrite the permutation importance plot, if exists.
```
BenjaminBossan, line 1110:
```suggestion
            Whether to overwrite the permutation importance plot file, if a plot by that
            name already exists.
```

skops/card/tests/test_card.py
```diff
@@ -403,6 +404,105 @@ def test_add_twice(self, model_card):
         assert text1 == text2
 
 
+def test_add_plot(destination_path, model_card):
```
BenjaminBossan, line 407: You probably added this test because of a merge with main. However, it can be deleted, since the functionality is already tested here:

https://github.com/skops-dev/skops/blob/1fd7982a3e2eb7071729189f74f3e23c9ebb4569/skops/card/tests/test_card.py#L857

skops/card/tests/test_card.py
```diff
@@ -403,6 +404,105 @@ def test_add_twice(self, model_card):
...
+
+def test_multiple_permutation_importances(
+    iris_estimator, iris_data, model_card, destination_path
+):
+    X, y = iris_data
+    result = permutation_importance(
+        iris_estimator, X, y, n_repeats=10, random_state=42, n_jobs=2
+    )
+    model_card.add_permutation_importances(
+        result, X.columns, plot_file=(Path(destination_path) / "importance.png")
```
BenjaminBossan, line 442: Parentheses around plot file are not needed here, right?

skops/conftest.py
```diff
@@ -16,3 +18,26 @@ def mock_import(name, *args, **kwargs):
...
+
+    orig_import = builtins.__import__
+
+    def mock_import(name, *args, **kwargs):
+        if name == "matplotlib":
+            raise ImportError
+        return orig_import(name, *args, **kwargs)
+
+    with patch("builtins.__import__", side_effect=mock_import):
+        yield
```
BenjaminBossan, line 43: As discussed, after this `yield`, let's re-import matplotlib to restore the import cache.

BenjaminBossan (changes requested):
Overall, this LGTM and there is not much left to be changed. There are a few comments regarding docs, a test that can be deleted, and a minor addition to a fixture.

Then there are a bunch of comments about removing `# type: ignore`. I'm sure you added those because it was somehow necessary to avoid mypy issues, but I could not reproduce them. I cloned your branch and removed the comments and mypy was running fine -- both when called by pre-commit and when called directly.

Thus I'm sure the comments can be removed (some of them are not even from the code you wrote, so they are indeed not required or else the main branch would be failing). Could it maybe be that your local skops environment is somehow busted and that's why you get the issues?

merveenoyan:
@BenjaminBossan codecov fails 😅🥲

BenjaminBossan:
@merveenoyan Re-running solved the codecov issue. Not sure why codecov hates you specifically :D

Could you please merge the current main branch, as have now added CI for Python 3.11? Then I'll review again.

merveenoyan:
@BenjaminBossan merged

skops/conftest.py
```diff
@@ -16,3 +18,30 @@ def mock_import(name, *args, **kwargs):
 
     with patch("builtins.__import__", side_effect=mock_import):
         yield
+
+    import matplotlib  # noqa
```
BenjaminBossan, line 22: We don't need this line. First of all, it's pandas here, not matplotlib, second, we don't delete pandas from the import cache, so re-importing is not necessary.

skops/io/_audit.py
```diff
@@ -2,7 +2,7 @@
 
 import io
 from contextlib import contextmanager
-from typing import Any, Generator, Literal, Sequence, Type, Union
+from typing import Any, Generator, Literal, Sequence, Type, Union  # type: ignore
```
BenjaminBossan, line 5: Is the `# type: ignore` here  really necessary?

skops/utils/importutils.py
```diff
@@ -0,0 +1,29 @@
+from importlib import import_module
+
+
+def import_or_raise(module, feature_name):
+    """Raise error
```
BenjaminBossan, line 5: Let's add a better description here.

skops/utils/importutils.py
```diff
@@ -0,0 +1,29 @@
...
+    Raises
+    ------
+    ModuleNotFoundError
+        Is raised if a given module is not present in the environment
+    """
+    try:
+        module = import_module(module)
+    except ImportError as e:
+        package = module.split(".")[0]
+        raise ModuleNotFoundError(
```
BenjaminBossan, line 24: For my understanding: Why not a simple `ImportError`?

BenjaminBossan (changes requested):
Sorry, again I have a few small comments. But I swear we're getting really close :)

BenjaminBossan:
@merveenoyan Some tests are failing. This is mostly fixed after merging with current main branch. Could you please do that?

merveenoyan:
@BenjaminBossan should work now (magically ✨) but I realized pytest doesn't collect tests of parser for some reason on mac (assuming it's same on windows) (that's why nothing failed on my local, I guess)

I checked github workflow and looked for `needs` in case the workflow doesn't run it unless ubuntu tests pass (since I know ubuntu tests are usually fast so it's good to add such prerequisite, but here it's not the case). Then I decided to run the tests in isolation on my local both for whole file and specifically on the test that is failing, yet it refuses to collect on my local too! 🤯 

```
(py39) ➜  card git:(feature_importance) python3 -m pytest -sv tests/test_parser.py
================================================= test session starts =================================================
platform darwin -- Python 3.9.16, pytest-7.2.1, pluggy-1.0.0 -- /opt/anaconda3/envs/py39/bin/python3
cachedir: .pytest_cache
rootdir: /Users/mervenoyan/Desktop/skops/skops, configfile: pyproject.toml
plugins: flaky-3.7.0, cov-4.0.0, anyio-3.6.2
collected 0 items / 2 skipped  
```

do you know what's going on?

BenjaminBossan:
> I realized pytest doesn't collect tests of parser for some reason on mac (assuming it's same on windows) (that's why nothing failed on my local, I guess)

Probably it's just because you haven't installed pandoc locally? This line makes it so that tests are skipped when pandoc is not found:

https://github.com/skops-dev/skops/blob/d9b7c369e7998e0819ab63baea81f15a004d8418/skops/card/tests/test_parser.py#L14-L18

The reason why it runs on Ubuntu CI is:

https://github.com/skops-dev/skops/blob/d9b7c369e7998e0819ab63baea81f15a004d8418/.github/workflows/build-test.yml#L65-L67

I haven't added pandoc to the other builds because it's less trivial to install it on Mac and Windows and, as you mentioned, Ubuntu is the fastest.

BenjaminBossan (commented):
There are still a few `# type: ignore`s in here which I think can be removed, apart from that it's ready to be merged from my perspective.

BenjaminBossan (approved):
I think now we're good :tada:

Issue #118, opened 2022-08-26
Figure out CI issues


adrinjalali:
Since we've added inference tests, the CI almost always times out.

We should figure out why, starting by putting times in tests to see where the tests fail.

BenjaminBossan:
Issue opened here: https://github.com/huggingface/api-inference-community/issues/94

BenjaminBossan:
Let's close this for now after #212 has been merged.

PR #147, opened 2022-09-21
CI: Run inference tests only conditionally

BenjaminBossan:
As discussed [here](https://github.com/huggingface/api-inference-community/issues/94), only run the slow inference API tests on main or when the PR title contains the word "inference".

Addresses (though not really solves) #118. 

As always, these things are hard to test.

### DIFF
--- a/.github/workflows/inference-test-conditional.yml
+++ b/.github/workflows/inference-test-conditional.yml
@@ -1,6 +1,6 @@
 # Tests hitting the inference API are slow and prone to hanging, resulting in
-# timeouts. Therefore, we only run them on the main branch or if the PR title
-# contains the word "inference". See #118
+# timeouts. Therefore, we only run inference tests on the full test matrix if
+# the head commit message contains the word "inference". See #118
 
 name: inference-tests
 
@@ -13,15 +13,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  inference-pytest:
+  conditional-inference-pytest:
 
     runs-on: ${{ matrix.os }}
-    if: "(github.repository == 'skops-dev/skops') && ((github.ref == 'refs/heads/main') || contains(github.event.pull_request.title, 'inference'))"
+    if: "(github.repository == 'skops-dev/skops') && contains(github.event.head_commit.message, 'inference')"
     strategy:
       fail-fast: false  # need to see which ones fail
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python: ["3.7", "3.8", "3.9", "3.10.6"]
+        python: ["3.7", "3.8", "3.9", "3.10"]
 
     # Timeout: https://stackoverflow.com/a/59076067/4521646
     timeout-minutes: 15
@@ -41,18 +41,8 @@ jobs:
         pip list
       shell: bash
 
-    - name: Inference tests
+    - name: Conditional inference tests
       env:
         SUPER_SECRET: ${{ secrets.HF_HUB_TOKEN }}
       run: |
-        python -m pytest -s -v --cov-report=xml -m "inference" skops/
-
-    - name: Upload coverage to Codecov
-      uses: codecov/codecov-action@v2
-      with:
-        env_vars: OS,PYTHON
-        fail_ci_if_error: true
-        files: ./coverage.xml
-        flags: unittests
-        name: codecov-umbrella
-        verbose: true
+        python -m pytest -s -v -m "inference" skops/
--- a/.github/workflows/inference-test-partial.yml
+++ b/.github/workflows/inference-test-partial.yml
@@ -16,7 +16,7 @@ jobs:
   inference-pytest-partial:
 
     runs-on: ${{ matrix.os }}
-    if: "(github.repository == 'skops-dev/skops') && (!contains(github.event.head_commit.message, '[CI inference]'))"
+    if: ${{ (github.repository == 'skops-dev/skops') && (!contains(github.event.head_commit.message, '[CI inference]')) }}
     strategy:
       fail-fast: false  # need to see which ones fail
       matrix:
--- a/.github/workflows/inference-test-partial.yml
+++ b/.github/workflows/inference-test-partial.yml
@@ -1,6 +1,6 @@
 # Tests hitting the inference API are slow and prone to hanging, resulting in
-# timeouts. Therefore, we only run inference tests on the full test matrix if
-# the head commit message contains the word "inference". See #118
+# timeouts. Therefore, we only run a subset of settings with inference tests by
+# default. See #118
 
 name: inference-tests
 
@@ -13,15 +13,15 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  conditional-inference-pytest:
+  inference-pytest-partial:
 
     runs-on: ${{ matrix.os }}
-    if: "(github.repository == 'skops-dev/skops') && contains(github.event.head_commit.message, 'inference')"
+    if: "(github.repository == 'skops-dev/skops') && (!contains(github.event.head_commit.message, '[CI inference]'))"
     strategy:
       fail-fast: false  # need to see which ones fail
       matrix:
-        os: [ubuntu-latest, windows-latest, macos-latest]
-        python: ["3.7", "3.8", "3.9", "3.10"]
+        os: [ubuntu-latest]
+        python: ["3.10"]
 
     # Timeout: https://stackoverflow.com/a/59076067/4521646
     timeout-minutes: 15
@@ -41,7 +41,7 @@ jobs:
         pip list
       shell: bash
 
-    - name: Conditional inference tests
+    - name: Partial inference tests
       env:
         SUPER_SECRET: ${{ secrets.HF_HUB_TOKEN }}
       run: |
--- a/.github/workflows/inference-test-partial.yml
+++ b/.github/workflows/inference-test-partial.yml
@@ -2,7 +2,7 @@
 # timeouts. Therefore, we only run a subset of settings with inference tests by
 # default. See #118
 
-name: inference-tests
+name: inference-tests-partial
 
 on:
   - push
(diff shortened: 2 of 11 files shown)


BenjaminBossan:
To test the changes, I first created a PR containing the word "inference" and indeed all tests ran:

Screenshot from 2022-09-21 12-46-01

After that, I renamed the title to use "infe-rence" instead. Rerunning the CI didn't change anything (presumably the checks are not performed), so I had to push an empty commit. After that, indeed the inference tests were skipped:

Screenshot from 2022-09-21 12-50-14

This shows that we can now indeed skip the inference tests conditional on the PR title. I chose the PR title instead of commit message because the latter is harder to change after being pushed.

What I cannot test for obvious reasons is if the inference tests run indeed on the main branch, for this we would have to merge this PR.

adrinjalali:
I would go with commit message rather than PR title. For some commits we might want to test and for some we might skip. If we forget the tag in a commit, we can add an empty commit with the tag.

But generally, I think this alternate solution might be better:

- have a crone job which pushes to the hub repos like the ones generated by `generate.py`/`run.sh` on a daily basis.
- all inference tests would target those repos instead of creating a repo and testing immediately after.

We can even _pin_ those repos to make sure they're always up and running for tests to run quite fast.

WDYT?

BenjaminBossan:
> I would go with commit message rather than PR title. For some commits we might want to test and for some we might skip. If we forget the tag in a commit, we can add an empty commit with the tag.

Honestly, I don't think this comes up very often? And you can still change the PR title for that specific commit? This seems to be more flexible to me.

> But generally, I think this alternate solution might be better:
> 
> * have a crone job which pushes to the hub repos like the ones generated by `generate.py`/`run.sh` on a daily basis.
> * all inference tests would target those repos instead of creating a repo and testing immediately after.

Not sure if I understand 100%. Those repos would presumably be created based on the main branch? If so, if there is a skops change in the PR that breaks stuff already at the repo creation stage (say, a change in the metainfo), the CI wouldn't catch that on the PR. Or do I misunderstand?

Regardless of that, for the time being, we could still go with this PR and then remove the inference GH action once we have added said script.

> We can even _pin_ those repos to make sure they're always up and running for tests to run quite fast.

I didn't know about that possibility.

adrinjalali:
Ok, so what about this:

- make inference tests run only on one CI (linux+latest python?)
- add a commit tag which would trigger inference tests on all CI

This would also fix our codecov issue.

I prefer commit message tags since they are more of a norm than PR title triggers, and I think it's more intuitive to have CI triggers on commits and their messages rather than on commits + whatever's on the PR title. At least in sklearn we have quite a few of those CI triggers.

BenjaminBossan:
Okay, then let's proceed like this.

What is the "right" way to get the commit message? I found `github.event.head_commit.message` but couldn't find any use in sklearn, so I assume there is another way?

adrinjalali:
in sklearn we don't do these in gh actions, instead we do them in shell, and there we use `git`.

I'm impartial to whether doing it through GH action stuff or in shell.

.github/workflows/inference-test-conditional.yml
```diff
@@ -0,0 +1,48 @@
+# Tests hitting the inference API are slow and prone to hanging, resulting in
+# timeouts. Therefore, we only run inference tests on the full test matrix if
+# the head commit message contains the word "inference". See #118
```
adrinjalali, line 3: maybe something like `[CI inference]` instead?

.github/workflows/inference-test-conditional.yml
```diff
@@ -0,0 +1,48 @@
...
+on:
+  - push
+  - pull_request
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  conditional-inference-pytest:
```
adrinjalali, line 16: maybe `inference-pytest-full` and `inference-pytest-partial` for the other one?

.github/workflows/inference-test-unconditional.yml
```diff
@@ -0,0 +1,48 @@
...
+# default. See #118
+
+name: inference-tests
+
+on:
+  - push
+  - pull_request
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
```
adrinjalali, line 12: we can also disable this when `[CI inference]` is in the commit message, since the other job does this as well.

adrinjalali (approved):

BenjaminBossan:
@adrinjalali So it seems not to work correctly. First try, `inference-pytest-full` was skipped and `inference-pytest-partial` ran, as expected:

Screenshot from 2022-09-29 15-13-39

Second try, after an empty commit containing "[CI inference]", `inference-pytest-partial` ran (why?) and `inference-pytest-full` is not listed at all??

Screenshot from 2022-09-29 15-54-46

adrinjalali:
Your two new workflows have the same name @BenjaminBossan , I wouldn't be surprised if that's the root of the issue. Maybe try two different names?

BenjaminBossan:
> Your two new workflows have the same name

Good catch, I tried to have different names everywhere but missed the top level name :/

Unfortunately, this still doesn't work. First, without the name match, everything looks as expected:

Screenshot from 2022-09-29 17-25-17

But with the name match, `full` is still skipped and `partial` runs (ignore the other failing tests for now):

Screenshot from 2022-09-29 17-31-12

For reference, I used the logic described [here](https://dev.to/mliakos/conditional-github-action-based-on-commit-message-2l02) to check the commit message (but also confirmed in other sources).

.github/workflows/inference-test-partial.yml
```diff
@@ -0,0 +1,48 @@
...
+
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  inference-pytest-partial:
+
+    runs-on: ${{ matrix.os }}
+    if: "(github.repository == 'skops-dev/skops') && (!contains(github.event.head_commit.message, '[CI inference]'))"
```
adrinjalali, line 19: shouldn't this be

```suggestion
    if: ${{ (github.repository == 'skops-dev/skops') && (!contains(github.event.head_commit.message, '[CI inference]')) }}
```

ref: https://docs.github.com/en/actions/learn-github-actions/expressions#example-expression-in-an-if-conditional

adrinjalali (commented):
Maybe expression issue?

.github/workflows/inference-test-partial.yml
BenjaminBossan: I'll try that, fingers crossed. If this is really the correct syntax, does this line work correctly:

https://github.com/skops-dev/skops/blob/95a718afd41306c6523d55f110a27b553eb3f7e1/.github/workflows/build-test.yml#L15

?

BenjaminBossan:
@adrinjalali Still the same issue, "full" never runs and "partial" always runs. So nothing changed, which is not too surprising, as

> When you use expressions in an if conditional, you may omit the expression syntax (${{ }}) because GitHub automatically evaluates the if conditional as an expression.

[link](https://docs.github.com/en/actions/learn-github-actions/expressions#about-expressions)

adrinjalali:
At this point I'd suggest creating a new repo on GH and testing things there maybe? Just in case other things are interacting in a weird way with one another.

BenjaminBossan:
> At this point I'd suggest creating a new repo on GH and testing things there maybe? Just in case other things are interacting in a weird way with one another.

I did a bunch of testing. I could make "full" run by not matching on `[CI inference]` but `CI inference` -- I assume that brackets require some kind of escaping, but I didn't bother to find out how exactly, I think this would be good enough.

However, I could not figure out how _not_ to make partial run. For reference, here is the original conditional expression:

`if ${{ (github.repository == 'skops-dev/skops') && (!contains(github.event.head_commit.message, '[CI inference]')) }}`

Things I tried:

1. Not using bracktets
2. Not using upper case
3. Move `!` before parens
4. Remove parens
5. Swap position with `(github.repository == 'skops-dev/skops')`
6. Match on a completely different word without whitespace
7. Only do the `contains` condition, not checking repository

None of these helped, it was always triggered. The only way to skip it was to do `if ${{ false }}`.

I found a [blog post](https://blog.mergify.com/running-github-actions-only-on-certain-pull-requests/) that uses an expression with `!` + `contains`, which I copied verbatim and still no luck. I have no more ideas what else to test...

adrinjalali:
I tried in a new repo, and things seem to work as expected: https://github.com/adrinjalali/gh-action-test/tree/main

The last two commits show how it'd work.

Also, could this be of some help?

https://github.com/nektos/act

BenjaminBossan:
> I tried in a new repo, and things seem to work as expected

I invited you to my repo. It still doesn't work there, even though I copy-pasted your expression (with just the repo name changed): https://github.com/BenjaminBossan/mops/pull/2

The whole setup is a bit closer than your example to what we have in skops, but those differences should _theoretically_ not affect this expression.

adrinjalali:
Maybe merge the two workflows under the same one, just two different jobs?

Also, you could start with a minimal setup like the one I have in that repo, and see at what point things break maybe?

BenjaminBossan:
> Maybe merge the two workflows under the same one, just two different jobs?

I copied your exact workflow (only changed repo name) and still the exact same issue:

https://github.com/BenjaminBossan/mops/pull/2

(latest 2 commits).

Any global repo settings that you changed? Otherwise, I'm ready to throw in the towel.

adrinjalali:
No, just created a repo, and created a workflow in it, nothing special. This is really odd!

BenjaminBossan:
Okay, so where do we go from here? To recap, none of the suggested methods seem to work for getting the commit message for "pull_request" triggers.

adrinjalali:
potential related solution: https://github.com/adrinjalali/gh-action-test/pull/1

adrinjalali:
Fixed in #212

Issue #149, opened 2022-09-22
Secure persistence: Contiguity of arrays may change after loading


BenjaminBossan:
After loading objects that contain numpy arrays, it can happen that the contiguity of the array is not the same as it was before saving. 

The current testing does not catch this but we could add more checks [here](https://github.com/skops-dev/skops/blob/005bb526466fdcb6c1aa223b77cab112f2a0d625/skops/io/tests/test_persist.py#L286):

```python
                assert val1.flags["C_CONTIGUOUS"] is val2.flags["C_CONTIGUOUS"]
                assert val1.flags["F_CONTIGUOUS"] is val2.flags["F_CONTIGUOUS"]
```

A peculiar things is that some of the arrays causing the error seem to be _neither_ C, nor Fortran contiguous. And the loaded array is _both_ C and Fortran contiguous. One example where this occurs is in `cv_results_` of `GrahpicalLassoCV`.

At least for performance reasons, it would be better to conserve the contiguity of the arrays. One approach could be to store the contiguity of the array in the `state` and then restore it after loading, although I don't know how to get an array with neither contiguity.

As for the other flags, I don't know how important they are and if ensuring that their values are conserved is feasible or even desirable.

adrinjalali:
This should only be the case for object dtypes, since `npy` ensures the same contiguity.

We probably can do better in storing those numpy arrays anyway, something closer to the `npy` binary than what we have now.

BenjaminBossan:
> This should only be the case for object dtypes, since `npy` ensures the same contiguity.

I haven't checked that. Anyway, should we consider this to be a bug?

> We probably can do better in storing those numpy arrays anyway, something closer to the `npy` binary than what we have now.

With "those", do you mean object arrays? Not sure how easy that would be.

adrinjalali:
I think if the `npy` format is not keeping the contiguity, we can report it upstream, but for object arrays, we can find a better way to store them.

BenjaminBossan:
I checked and it seems the issue is not that the dtype was object. E.g., for `GraphicalLassoCV`, we get the error on `cv_results_`, because `split_*_score` values change contiguity from neither C nor F, to both C and F. The dtype is `float64` and our `get_state` and `get_instance` methods save and load via numpy.

Interestingly, for the `mean_test_score`, we get no error -- it's both C and F contiguous before and after loading. So probably `np.save` and `np.load` work fine if the array is either C or F contiguous or both, but not if it's neither.

[...]

PR #150, opened 2022-09-22
Avoid duplicating numpy arrays

### LINKED ISSUE #135: Secure persistence: Avoid duplicate values
Right now, values can be duplicated, e.g. when the same numpy array is referenced multiple times as an attribute. We should try to avoid it if easily possible.

For numpy arrays, we can achieve that by using the hash instead of a UUID as the file name:

```python
import joblib
def ndarray_get_state(obj, dst):
    res = {
        "__class__": obj.__class__.__name__,
        "__module__": inspect.getmodule(type(obj)).__name__,
    }

    hashed = joblib.hash(obj)
    try:
        f_name = f"{hashed}.npy"
        if not os.path.exists(f_name):
            with open(Path(dst) / f_name, "wb") as f:
                np.save(f, obj, allow_pickle=False)
        res["type"] = "numpy"
        res["file"] = f_name
    except ValueError:
        # object arrays cannot be saved with allow_pickle=False, therefore we
        # convert them to a list and store them as a json file.
        f_name = f"{hashed}.json"
        if not os.path.exists(f_name):
            with open(Path(dst) / f_name, "w") as f:
                f.write(json.dumps(obj.tolist()))
        res["type"] = "json"
        res["file"] = f_name

    return res

```

BenjaminBossan:
Fixes #135

## Description

Instead of using a UUID for the name of the numpy files, use their hash. That way, when the same array already exists, we can use it instead of storing another copy.

## Implementation

I started out with using `hashlib.sha256` instead of delegating hashing to joblib. However, this caused several problems:

- sha256 only worked for C-contiguous arrays
- the sha256 of `np.array(0)`, `np.array([0])`, `np.array([[0]])` etc. were all the same, which prevents us from restoring the arrays to their original shape

I did not try the other hashlib algorithms. Both drawbacks could be worked around (store the contiguity and shape and restore them after loading, but see #149), but since joblib is already a dependency, I just went with it.

The disadvantage of `joblib.hash` is that it uses either md5 or sha1, both of which are not cryptographically secure anymore. However, for this use case, that should not be a necessity, since the hash only serves as a cache key.

**Addendum**

While working on this, I discovered a minor bug where trying to store an object numpy array resulted in the creation of a broken `.npy` file being left over. This is because numpy tries to write to the file until it encounters an error and raises, but then doesn't clean up said file. Before this bugfix, we would include that broken file in the zip archive, although we wouldn't do anything with it. Now, no such file is being created. A test is added to test this.

### DIFF
--- a/skops/io/_numpy.py
+++ b/skops/io/_numpy.py
@@ -8,6 +8,7 @@ import numpy as np
 from ._general import function_get_instance
 from ._persist import get_instance, get_state
 from ._utils import _import_obj, get_module
+from .exceptions import UnsupportedTypeException
 
 
 def ndarray_get_state(obj, dst):
@@ -30,7 +31,7 @@ def ndarray_get_state(obj, dst):
                 f"numpy arrays of type {type(obj)} and dtype {obj.dtype} are not "
                 "supported yet"
             )
-            raise TypeError(msg) from exc
+            raise UnsupportedTypeException(msg) from exc
     else:
         # Object arrays cannot be saved with allow_pickle=False, therefore we
         # convert them to a list and recursively call get_state on it.
--- a/skops/io/_numpy.py
+++ b/skops/io/_numpy.py
@@ -1,7 +1,8 @@
 import io
+import os
 from pathlib import Path
-from uuid import uuid4
 
+import joblib
 import numpy as np
 
 from ._general import function_get_instance
@@ -16,11 +17,13 @@ def ndarray_get_state(obj, dst):
     }
 
     try:
-        f_name = f"{uuid4()}.npy"
-        with open(Path(dst) / f_name, "wb") as f:
-            np.save(f, obj, allow_pickle=False)
-            res["type"] = "numpy"
-            res["file"] = f_name
+        digest = joblib.hash(obj)
+        f_name = f"{digest}.npy"
+        if not os.path.exists(f_name):
+            with open(Path(dst) / f_name, "wb") as f:
+                np.save(f, obj, allow_pickle=False)
+        res["type"] = "numpy"
+        res["file"] = f_name
     except ValueError:
         # Object arrays cannot be saved with allow_pickle=False, therefore we
         # convert them to a list and recursively call get_state on it.
--- a/skops/io/_numpy.py
+++ b/skops/io/_numpy.py
@@ -16,15 +16,22 @@ def ndarray_get_state(obj, dst):
         "__module__": get_module(type(obj)),
     }
 
-    try:
-        digest = joblib.hash(obj)
-        f_name = f"{digest}.npy"
-        if not os.path.exists(f_name):
-            with open(Path(dst) / f_name, "wb") as f:
-                np.save(f, obj, allow_pickle=False)
-        res["type"] = "numpy"
-        res["file"] = f_name
-    except ValueError:
+    if obj.dtype != object:
+        try:
+            digest = joblib.hash(obj)
+            f_name = f"{digest}.npy"
+            if not os.path.exists(Path(dst) / f_name):
+                with open(Path(dst) / f_name, "wb") as f:
+                    np.save(f, obj, allow_pickle=False)
+            res["type"] = "numpy"
+            res["file"] = f_name
+        except ValueError as exc:
+            msg = (
+                f"numpy arrays of type {type(obj)} and dtype {obj.dtype} are not "
+                "supported yet"
+            )
+            raise TypeError(msg) from exc
+    else:
         # Object arrays cannot be saved with allow_pickle=False, therefore we
         # convert them to a list and recursively call get_state on it.
         if obj.dtype == object:
@@ -32,8 +39,6 @@ def ndarray_get_state(obj, dst):
             res["content"] = obj_serialized["content"]
             res["type"] = "json"
             res["shape"] = get_state(obj.shape, dst)
-        else:
-            raise TypeError(f"numpy arrays of dtype {obj.dtype} are not supported yet")
 
     return res
 
--- a/skops/io/tests/test_persist.py
+++ b/skops/io/tests/test_persist.py
@@ -650,3 +650,28 @@ def test_identical_numpy_arrays_not_duplicated(tmp_path):
     expected_files = 5
     num_files = len(files)
     assert num_files == expected_files
+
+
+class NumpyDtypeObjectEstimator(BaseEstimator):
+    """An estimator with a numpy array of dtype object"""
+
+    def fit(self, X, y=None, **fit_params):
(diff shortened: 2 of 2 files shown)


BenjaminBossan:

skops/io/_numpy.py
```diff
@@ -15,22 +16,29 @@ def ndarray_get_state(obj, dst):
...
+                    np.save(f, obj, allow_pickle=False)
             res["type"] = "numpy"
             res["file"] = f_name
-    except ValueError:
+        except ValueError as exc:
+            msg = (
+                f"numpy arrays of type {type(obj)} and dtype {obj.dtype} are not "
+                "supported yet"
+            )
+            raise TypeError(msg) from exc
```
adrinjalali, line 33: should this be our unsupported thingy error?

skops/io/_numpy.py
```diff
@@ -15,22 +16,29 @@ def ndarray_get_state(obj, dst):
         "__module__": get_module(type(obj)),
     }
 
-    try:
-        f_name = f"{uuid4()}.npy"
-        with open(Path(dst) / f_name, "wb") as f:
-            np.save(f, obj, allow_pickle=False)
+    if obj.dtype != object:
+        try:
+            digest = joblib.hash(obj)
```
adrinjalali, line 21: I think it'd be nicer to do `id(obj)` which checks memory address. It's clearly much faster, and in our usecase it does prevent persisting identical arrays. pickle does the same.

BenjaminBossan: The problem with `id(obj)` is that it wouldn't catch exact copies, would it? Is speed even a concern here?

skops/io/_numpy.py
BenjaminBossan: Done

skops/io/_numpy.py
adrinjalali: yes, it doesn't catch exact copies, but I think we're more worried about storing the same object several times rather than storing copies. I'd say creating a hash of a large matrix could be a bottleneck, and I think sticking with pickle convention here would be nicer.

BenjaminBossan: I did the change to `id(obj)` but then the tests failed. The reason is that lines like those:

```python
        self.vector = X[0]
        self.vector_2 = X[0]
```

produce different ids. Is this the behavior we want?

Also, I checked the timing for hashing:

```python
>>> x = np.random.random((10000, 1000))
>>> %timeit joblib.hash(x)
112 ms ± 2.55 ms per loop (mean ± std. dev. of 7 runs, 10 loops each)
>>> y = np.random.random((10, 10))
53.5 µs ± 89.1 ns per loop (mean ± std. dev. of 7 runs, 10,000 loops each)
```

So both a few big arrays and many small arrays shouldn't be a big problem.

BenjaminBossan: Yes, I'm aware, the question is if we want to store two different copies of the array when it was indexed, even though they're identical.

adrinjalali: Yeah I don't think that's something we really have to optimize for here.

adrinjalali:
I think I understand what's happening.

Saving trees involves calling their `__getstate__` and `__reduce__`, which create arrays representing the nodes and trees.

Pickle keeps a reference to the objects being saved in its `memo`, which means until pickling is finished, those objects are not GCed, and therefore new objects won't be on the same address, which is what cpython returns as the output of `id()`.

However, we get the output of those methods in our implementation, serialize them, and then move on w/o keeping a reference to them, and therefore allow GC to clean them up.

I considered having a `memo` in our implementation, but that would mean passing an extra state to our methods. Back then, I didn't want to do it since it'd complicate the signature of those functions and would make extending the persistence by third party libraries more complicated. But since we're moving more towards expecting third party estimators to implement `__getstate__`/`__setstate__` and expect them to return things we understand how to persist, we can actually move our functions all to a class like `_Pickler` does, and then we'd have access to these states. WDYT?

BenjaminBossan:
> Pickle keeps a reference to the objects being saved in its `memo`, which means until pickling is finished, those objects are not GCed, and therefore new objects won't be on the same address, which is what cpython returns as the output of `id()`.

That sounds very plausible, without a cache, `id` is just too unpredictable.

> But since we're moving more towards expecting third party estimators to implement `__getstate__`/`__setstate__` and expect them to return things we understand how to persist, we can actually move our functions all to a class like `_Pickler` does, and then we'd have access to these states. WDYT?

Yes, I expected something like this would become necessary sooner or later (my first contribution to persist already had this change of you remember ;)

Implementation-wise, we could do as you suggest or pass around a third argument to all the `get_*` functions. The latter would be okay for now but grants less flexibility if we ever need to pass even more arguments. Besides memo, I would imagine that there could be a future need to pass around some kind of config and the dispatcher (maybe turned into a proper class), and we don't want to pass around even more arguments.

Regarding this PR, should we bench it until we have a solution to the discussed problem?

adrinjalali:
> Regarding this PR, should we bench it until we have a solution to the discussed problem?

Agreed.

> Besides memo, I would imagine that there could be a future need to pass around some kind of config and the dispatcher (maybe turned into a proper class), and we don't want to pass around even more arguments.

I agree, and that's why I think it might be time to move them into a class, and just have `self` as our state everywhere.

would you like to tackle it? It's also going to help with the audit work I think.

BenjaminBossan:
> would you like to tackle it? It's also going to help with the audit work I think.

Yes, I can work on that.

BenjaminBossan:
> it might be time to move them into a class, and just have `self` as our state everywhere.

Okay, I took a look and it's not quite as easy. If we want to keep the tidy module structure we have right now, the functions cannot all live in the same class. We could make dummy or "mixin" classes in each module and then create a big class that collects all those individual classes. But that's kinda ugly and makes it more difficult to dynamically add new methods (if we need to do that after all). One other disadvantage is that we cannot use `singledispatch` but have to change to `singledispatchmethod` (haven't checked if that would cause any problems).

In the end, none of that is impossible but it doesn't look very elegant to me. So I wonder if adding a third argument (`persister` and `unpersister`/`loader`) to the `get_` functions might not be preferable after all. Or perhaps there is a better way that I didn't think of. WDYT @adrinjalali?

adrinjalali:
I wouldn't mind moving them all to the same class though. But I agree that might not look as nice.

I'd be okay with instead of passing `dst`, passing some sort of `config` or `state` or something. Afterall, `dst` and `trusted_source` are of the same nature, as in they're user inputs. Or should we pass a `config` and a `state`, one staying unchanged, the other one being modified by the methods?

adrinjalali:
Completed in #173

PR #152, opened 2022-09-26
Fix FutureWarnings for huggingface_hub>=0.10

BenjaminBossan:
For that Hub version, the use of the `token` argument produces a `FutureWarning`, leading our CI to fail. This PR fixes the warnings while leaving everything the same on the user side (i.e. they don't need to change the `token` argument).

## Implementation

As discussed, we will not change the argument name on the skops, preferring token to `use_auth_token`.

For the version checking, I went with the private `Version` class from [scipy](https://github.com/scipy/scipy/blob/main/scipy/_lib/_pep440.py), which follows PEP440. This follows the approach taken in skorch (see this issue: https://github.com/skorch-dev/skorch/issues/887), though we import it here instead of vendoring. Let me know if another approach is preferred.

### DIFF
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -12,10 +12,11 @@ import warnings
 from pathlib import Path
 from typing import Any, List, MutableMapping, Optional, Union
 
+import huggingface_hub
 import numpy as np
 from huggingface_hub import HfApi, InferenceApi, snapshot_download
 
-from ..utils.fixes import Literal
+from ..utils.fixes import Literal, Version
 
 SUPPORTED_TASKS = [
     "tabular-classification",
@@ -626,7 +627,13 @@ def get_model_output(repo_id: str, data: Any, token: Optional[str] = None) -> An
     Also note that if the model repo is private, the inference API would not be
     available.
     """
-    model_info = HfApi().model_info(repo_id=repo_id, token=token)
+    # TODO: remove once we drop support for Hub<0.10
+    if Version(huggingface_hub.__version__) < Version("0.10"):
+        extra_kwargs = {"token": token}
+    else:
+        extra_kwargs = {"use_auth_token": token}
+
+    model_info = HfApi().model_info(repo_id=repo_id, **extra_kwargs)
     if not model_info.pipeline_tag:
         raise ValueError(
             f"Repo {repo_id} has no pipeline tag. You should set a valid 'task' in"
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
@skops-dev/maintainers ready for review. Until this or another fix is merged, all our CI runs will fail, so this PR should have priority over the others.

CI only checks `huggingface_hub-0.10.0rc0` but locally I also tested with `huggingface_hub-0.9.1` and it succeeded. Theoretically, it would be best to add both versions to the test matrix if we plan on supporting them for a while, but that would multiply the number of test runs by another factor of 2, so not worth it IMO.

skops/hub_utils/_hf_hub.py
```diff
@@ -626,7 +627,13 @@ def get_model_output(repo_id: str, data: Any, token: Optional[str] = None) -> An
     Also note that if the model repo is private, the inference API would not be
     available.
     """
-    model_info = HfApi().model_info(repo_id=repo_id, token=token)
+    # TODO: remove once we drop support for Hub<0.10
+    if Version(huggingface_hub.__version__) <= Version("0.9.1"):
+        extra_kwargs = {"token": token}
+    else:
+        extra_kwargs = {"use_auth_token": token}
```
adrinjalali, line 634: rather have this in `fixes.py` as well.

adrinjalali: and we don't need to check for version, we try with `use_auth_token`, if it fails we do `token`.

adrinjalali (commented):
Thanks @BenjaminBossan

skops/hub_utils/_hf_hub.py
BenjaminBossan: Hmm, I understand that try...except allows us to skirt having version checking code. But having a bunch of

```python
try:
    client.foo(repo_id=repo_id, repo_type="model", token=token)
except TypeError:
    client.foo(repo_id=repo_id, repo_type="model", use_auth_token=token)
```

looks suboptimal to me. It litters our code in multiple places, instead of dealing with the issue once. And, just as a stupid example, what if Hub re-introduces the `token` argument but with a different functionality, it could lead to unwanted side-effects.

adrinjalali: We can have a `Client` class in `fixes.py` which does this, and import that instead, and that class exposes the methods we need in `skops`, and handles different versions there. The rest of the codebase would stay clean this way.

BenjaminBossan: That might be a solution, though a bit of a bigger undertaking with some open questions:

* private or public?
* automatically expose all methods or only those we use?
* handle versions with try...expect or version checks?
* also handle inference api?
* testing will probably not be quite trivial

As a quick fix for this issue here, we could also pin the `huggingface_hub` version and close this PR. WDYT?

adrinjalali: Since we don't have many users yet, I'm also okay with closing this, and then moving along with using modelcards from the new release, which would mean depending on the latest one anyway, and we won't have this issue. So happy to close (I'd leave that to you).

Issue #154, opened 2022-09-28
Bug: Table rendering in markdown broken when there is a line break in a repr


BenjaminBossan:
When tables have entries with line breaks, the table is not rendered correctly. Possibly, a solution is to replace the `\n` with `<br/>`.

PR #156, opened 2022-09-28
BUG Fix line breaks in markdown tables

BenjaminBossan:
Solves #154 

## Description

Markdown table entires shouldn't contain `"\n"` line breaks. Instead, those should be replaced by `"<br />"`. This is now done for both the hyperparameter table and for tables added through `card.add_table`.

An example of a fixed hyperparameter table: https://huggingface.co/skops-ci/hf_hub_example-a4b4abaf-d1c3-4300-a5ac-cc9e3c65bd39#hyperparameters

## Implementation

The bugfix has to be applied after calling `tabulate`, as the table entries can contain arbitrary objects, whose `repr` is only called by tabulate. That also means we have to be careful as the tabulated table string will contain valid `"\n"` line breaks that should be left untouched.

Unfortunately, the hyperparameter table doesn't use `card.add_table`, which is why the fix has to be applied in two places. IMO it would be better if it used that method under the hood but that would result in the hyperparameter table being placed in a different section (because of the inflexibility of the templating).

A minor naming issue I found: The method that actually creates the hyperparameter table is called `_extract_estimator_config`. I would have never guessed to be honest.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -16,6 +16,8 @@ v0.3
 - Add ``private`` as an optional argument to :meth:`.hub_utils.push` to
   optionally set the visibility status of a repo when pushing to the hub.
   :pr:`130` by `Adrin Jalali`_.
+- Fix a bug that resulted in markdown tables being rendered incorrectly if
+  entries contained line breaks by `Benjamin Bossan`_.
 
 v0.2
 ----
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -28,6 +28,19 @@ def wrap_as_details(text: str, folded: bool) -> str:
     return f"<details>\n<summary> Click to expand </summary>\n\n{text}\n\n</details>"
 
 
+def _clean_table(table: str) -> str:
+    # replace line breaks "\n" with html tag <br />, however, leave end-of-line
+    # line breaks (eol_lb) intact
+    eol_lb = "|\n"
+    placeholder = "$%!?"  # arbitrary sting that never appears naturally
+    table = (
+        table.replace(eol_lb, placeholder)
+        .replace("\n", "<br />")
+        .replace(placeholder, eol_lb)
+    )
+    return table
+
+
 @dataclass
 class PlotSection:
     """Adds a link to a figure to the model card"""
@@ -48,7 +61,7 @@ class PlotSection:
 class TableSection:
     """Adds a table to the model card"""
 
-    table: dict["str", list[Any]]
+    table: dict[str, list[Any]]
     folded: bool = False
 
     def __post_init__(self) -> None:
@@ -78,8 +91,8 @@ class TableSection:
         else:
             headers = self.table.keys()
 
-        table = tabulate(
-            self.table, tablefmt="github", headers=headers, showindex=False
+        table = _clean_table(
+            tabulate(self.table, tablefmt="github", headers=headers, showindex=False)
         )
         return wrap_as_details(table, folded=self.folded)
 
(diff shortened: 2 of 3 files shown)


BenjaminBossan:
Note: CI is still failing because of the breaking change in `huggingface_hub`.

merveenoyan:
@BenjaminBossan Can we just pin the version of HF hub used in CI for "`test`" separately until we get up to date with Hub changes? https://github.com/skops-dev/skops/blob/132925bb4368960b799df252692f1141c4bc838d/skops/_min_dependencies.py#L14

merveenoyan (approved):
LGTM, thanks a lot! can you please post the model cards (of the skorch example and another example that table already works) after this change here?

BenjaminBossan:
> Can we just pin the version of HF hub used in CI for "`test`" separately until we get up to date with Hub changes?

I suggested pinning as a quick fix until we get a proper fix [here](https://github.com/skops-dev/skops/pull/152#discussion_r982216081).

> can you please post the model cards

I linked [this](https://huggingface.co/skops-ci/hf_hub_example-a4b4abaf-d1c3-4300-a5ac-cc9e3c65bd39#hyperparameters) above, that should be sufficient, right?

docs/changes.rst
```diff
@@ -16,6 +16,8 @@ v0.3
 - Add ``private`` as an optional argument to :meth:`.hub_utils.push` to
   optionally set the visibility status of a repo when pushing to the hub.
   :pr:`130` by `Adrin Jalali`_.
+- Fix a bug that resulted in markdown tables being rendered incorrectly if
+  entries contained line breaks by `Benjamin Bossan`_.
```
adrinjalali, line 20: PR number missing here.

skops/card/_model_card.py
```diff
@@ -28,6 +28,19 @@ def wrap_as_details(text: str, folded: bool) -> str:
     return f"<details>\n<summary> Click to expand </summary>\n\n{text}\n\n</details>"
 
 
+def _clean_table(table: str) -> str:
+    # replace line breaks "\n" with html tag <br />, however, leave end-of-line
+    # line breaks (eol_lb) intact
+    eol_lb = "|\n"
+    placeholder = "$%!?"  # arbitrary sting that never appears naturally
```
adrinjalali, line 35: as a friend once said, equivalent to banging my head against the keyboard lol

BenjaminBossan: It's stupid :D If you have a better idea, please let me know

About the PR number: I fixed that. How do you add the correct number before you even created the PR? Do you just check the highest number so far and hope that until you actually open the PR, there hasn't been another PR in-between?

adrinjalali: I don't, I just add that as a commit after I create the PR :D

BenjaminBossan:
@skops-dev/maintainers I fixed the failing CI tests, so the PR is ready for re-review.

The error was that the markdown table had different padding between the expected and the actual outcome, e.g.:

```
expected:

| foo | bar |
|-----|-----|
|  1  |  2  |

result:

| foo |  bar  |
|-----|-------|
|  1  |   2   |
```

Neither Adrin nor I could reproduce the error locally, even on the same OS and Python version. As this padding is only relevant when reading the table in raw text but not for how it's rendered, I opted to remove it completely when comparing the resulting table with the expected table:

```
after processing:

| foo | bar |
|-|-|
| 1 | 2 |
```

I'm not super happy with this but it seemed like an okay solution. Alternatively, I thought about parsing the markdown table into a pandas DataFrame, but that seemed to be less rigorous. If you know any better solution, LMK.

adrinjalali (approved):

PR #160, opened 2022-09-29
TST Add checks for contiguity of loaded numpy arrays

### LINKED ISSUE #149: Secure persistence: Contiguity of arrays may change after loading

BenjaminBossan:
Resolves #149 

If an array is at least 2d, test if the contiguity is preserved. For some sklearn models, there are attributes whose contiguity changes after loading (they are curiously neither C nor F contiguous). Those are all 1d, however (sometimes after squeezing), so contiguity should not be an issue.

There is a known gap with 2d (or higher) object arrays, where contiguity is not always preserved. Since those don't seem to occur in sklearn, we don't fix that error for now.

### DIFF
--- a/skops/io/tests/test_persist.py
+++ b/skops/io/tests/test_persist.py
@@ -275,6 +275,11 @@ def _assert_vals_equal(val1, val2):
         assert sparse.issparse(val2) and ((val1 - val2).nnz == 0)
     elif isinstance(val1, (np.ndarray, np.generic)):
         if len(val1.dtype) == 0:
+            # for arrays with at least 2 dimensions, check that contiguity is
+            # preserved
+            if val1.squeeze().ndim > 1:
+                assert val1.flags["C_CONTIGUOUS"] is val2.flags["C_CONTIGUOUS"]
+                assert val1.flags["F_CONTIGUOUS"] is val2.flags["F_CONTIGUOUS"]
             if val1.dtype == object:
                 assert val2.dtype == object
                 assert val1.shape == val2.shape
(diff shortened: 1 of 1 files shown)


skops/io/tests/test_persist.py
```diff
@@ -275,6 +275,11 @@ def _assert_vals_equal(val1, val2):
         assert sparse.issparse(val2) and ((val1 - val2).nnz == 0)
     elif isinstance(val1, (np.ndarray, np.generic)):
         if len(val1.dtype) == 0:
+            # for arrays with at least 2 dimensions, check that contiguity is
+            # preserved
+            if val1.squeeze().ndim > 1:
+                assert val1.flags["C_CONTIGUOUS"] is val2.flags["C_CONTIGUOUS"]
+                assert val1.flags["F_CONTIGUOUS"] is val2.flags["F_CONTIGUOUS"]
```
adrinjalali, line 282: and all the tests pass! nice!

adrinjalali (approved):

PR #161, opened 2022-09-30
MNT Swap the names of _get_state <=> get_state & _get_instance <=> get_instance

BenjaminBossan:
The functions that should be actually used everywhere are `_get_state` and `_get_instance`, not `get_state` and `get_instance`. This is inconvenient and confusing. Therefore, swap out the names so that the functions being used everywhere now have the "public" name and the other ones the "private" name.

This was discussed [here](https://github.com/skops-dev/skops/pull/143#issue-1376742945).

### DIFF
--- a/skops/io/_persist.py
+++ b/skops/io/_persist.py
@@ -9,7 +9,7 @@ from zipfile import ZipFile
 
 import skops
 
-from ._utils import get_instance, get_state
+from ._utils import _get_instance, _get_state
 
 # For now, there is just one protocol version
 PROTOCOL = 0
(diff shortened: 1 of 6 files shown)


BenjaminBossan:

skops/io/_persist.py
```diff
@@ -54,7 +54,7 @@ def save(obj, file):
     """
     with tempfile.TemporaryDirectory() as dst:
         with open(Path(dst) / "schema.json", "w") as f:
-            state = get_state(obj, dst)
+            state = _get_state(obj, dst)
```
adrinjalali, line 57: why aren't we using `get_state` (with the new def) here?

BenjaminBossan: Yes, that should work, I did a simple search/replace here, I made the change now.

adrinjalali (approved):

PR #163, opened 2022-09-30
MNT Model card refactor to move plotting & table

### LINKED ISSUE #157: Create hyperparameter table and model diagram during render
RE: slack discussion we had with @BenjaminBossan I'm in favor of creating the table during rendering rather than initially creating the `Card` instance. I made it back then like that because we weren't having a separate function for rendering, and I wanted all cards to by default have this table and plot of the model as well. If the model is replaced by another the table will not match with the one that was created during creation of the card. 

- [ ] Move table
- [ ] Move plot

merveenoyan:
Resolves #157.
I moved plot & hyperparameter table to `_generate_card()` to keep `save()` atomic, given it's already called during `save()`. Does it sound good?

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -409,7 +409,7 @@ class Card:
                         " json.load(f)\n"
                         'clf.predict(pd.DataFrame.from_dict(config["sklearn"]["example_input"]))'
                     )
-        if self._model_diagram is True:
+        if self.model_diagram is True:
             model_plot: str | None = re.sub(
                 r"\n\s+", "", str(estimator_html_repr(self.model))
             )
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -241,15 +241,8 @@ class Card:
         metadata: Optional[CardData] = None,
     ) -> None:
         self.model = model
-        self._hyperparameter_table = self._extract_estimator_config()
-        # the spaces in the pipeline breaks markdown, so we replace them
+        self._model_diagram = model_diagram
         self._eval_results = {}  # type: ignore
-        if model_diagram is True:
-            self._model_plot: str | None = re.sub(
-                r"\n\s+", "", str(estimator_html_repr(model))
-            )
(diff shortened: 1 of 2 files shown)


skops/card/_model_card.py
```diff
@@ -380,6 +373,8 @@ def _generate_card(self) -> ModelCard:
         # add evaluation results
 
         template_sections = copy.deepcopy(self._template_sections)
+        self._hyperparameter_table = self._extract_estimator_config()
```
BenjaminBossan, line 376: Do we even need to assign `self._hyperparameter_table` anymore? We could just do:

`hyperparameter_table = self._extract_estimator_config()`

and below, inside `ModelCard.from_template`, replace `hyperparameter_table=self._hyperparameter_table,` with `hyperparameter_table=hyperparameter_table,`.

Same argument goes for `self._model_plot`.

BenjaminBossan (commented):
I have a comment that might simplify the code a bit, otherwise LGTM.

Regarding the testing, we could test exactly the case where `card.model` is reassigned, but I don't think it's strictly necessary.

skops/card/_model_card.py
merveenoyan: Addressed this.

BenjaminBossan (approved):

skops/card/_model_card.py
```diff
@@ -241,15 +241,8 @@ def __init__(
         metadata: Optional[CardData] = None,
     ) -> None:
         self.model = model
-        self._hyperparameter_table = self._extract_estimator_config()
-        # the spaces in the pipeline breaks markdown, so we replace them
+        self._model_diagram = model_diagram
```
adrinjalali, line 244: since this is a constructor parameter, it doesn't have to be private.

```suggestion
        self.model_diagram = model_diagram
```

adrinjalali (commented):
Otherwise LGTM

adrinjalali (approved):

Issue #96, opened 2022-08-11
Take model as argument to generate the card


merveenoyan:
About [this comment in this PR](https://github.com/huggingface/blog/pull/441/files#r943201153). Currently model card generation takes an argument called model which actually is a model path. We need to take model as well as model path, we can warn the user if both of them are missing.

adrinjalali:
I don't think we need to change anything about the args, the `model` arg can take an estimator instance, or a path.

p-mishra1:
@merveenoyan I created a PR regarding this #174 .

PR #174, opened 2022-10-05
load model in Card class

p-mishra1:
#96 added a function in the Card class to load the model if provided a file path. 
sorry if the implementation looks naive

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -16,6 +16,7 @@ from sklearn.utils import estimator_html_repr
 from tabulate import tabulate  # type: ignore
 
 import skops
+from skops.io import load
 
 # Repr attributes can be used to control the behavior of repr
 aRepr = Repr()
@@ -26,7 +27,10 @@ aRepr.maxstring = 79
 def wrap_as_details(text: str, folded: bool) -> str:
     if not folded:
         return text
-    return f"<details>\n<summary> Click to expand </summary>\n\n{text}\n\n</details>"
+    return (
+        "<details>\n<summary> Click to expand"
+        f" </summary>\n\n{text}\n\n</details>"
+    )
 
 
 def _clean_table(table: str) -> str:
@@ -93,7 +97,9 @@ class TableSection:
             headers = self.table.keys()
 
         table = _clean_table(
-            tabulate(self.table, tablefmt="github", headers=headers, showindex=False)
+            tabulate(
+                self.table, tablefmt="github", headers=headers, showindex=False
+            )
         )
         return wrap_as_details(table, folded=self.folded)
 
@@ -151,7 +157,9 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
     task = config.get("sklearn", {}).get("task", None)
     if task:
         card_data.tags += [task]
-    card_data.model_file = config.get("sklearn", {}).get("model", {}).get("file")
+    card_data.model_file = (
+        config.get("sklearn", {}).get("model", {}).get("file")
+    )
     example_input = config.get("sklearn", {}).get("example_input", None)
     # Documentation on what the widget expects:
     # https://huggingface.co/docs/hub/models-widgets-examples
@@ -173,7 +181,7 @@ class Card:
 
     Parameters
     ----------
-    model: estimator object
+    model: pathlib.path, str, or sklearn estimator object
         Model that will be documented.
 
     model_diagram: bool, default=True
@@ -255,7 +263,7 @@ class Card:
         model_diagram: bool = True,
         metadata: Optional[CardData] = None,
     ) -> None:
-        self.model = self._load_model(model)
+        self._model = model
         self.model_diagram = model_diagram
         self._eval_results = {}  # type: ignore
         self._template_sections: dict[str, str] = {}
@@ -303,11 +311,15 @@ class Card:
             Card object.
         """
         for plot_name, plot_path in kwargs.items():
-            section = PlotSection(alt_text=plot_name, path=plot_path, folded=folded)
+            section = PlotSection(
+                alt_text=plot_name, path=plot_path, folded=folded
+            )
             self._extra_sections.append((plot_name, section))
         return self
 
-    def add_table(self, folded: bool = False, **kwargs: dict["str", list[Any]]) -> Card:
+    def add_table(
+        self, folded: bool = False, **kwargs: dict["str", list[Any]]
+    ) -> Card:
         """Add a table to the model card.
 
         Add a table to the model card. This can be especially useful when you
@@ -374,29 +386,54 @@ class Card:
             self._eval_results[metric] = value
         return self
 
+    @property
+    def model(self):
+        model = self._load_model(self._model)
+        if model is not self._model:
+            self._model = model
+        return model
+
+    @model.setter
+    def model(self, model):
+        self._model = model
+
+    @model.deleter
+    def model(self):
+        del self._model
+
     def _load_model(self, model: Any) -> Any:
-        """Loads the model if provided a file path.
+        """Loads the model if provided a file path, if already a model instance,
+        return it unmodified.
 
         Parameters
         ----------
-        model : Any
-            Str or model instance.
+        model : pathlib.path, str, or sklearn estimator
+            Path/str or the actual model instance. If a Path or str, loads the model on first call.
 
         Returns
         -------
-        model : object 
+        model : object
             Model instance.
+
         """
-        if isinstance(model, str):
-            model_path = Path(model)
-            if not model_path.exists():
-                raise ValueError("Model file does not exist")
-            if model_path.suffix==".pkl":
-                model = joblib.load(model_path)
-            elif model_path.suffix==".skops":
-                model = skops.io.load(model_path)
-            else:
-                raise ValueError("Model Format not supported")
+        if not isinstance(model, (Path, str)):
+            return model
+
+        model_path = Path(model)
+        if not model_path.exists():
+            raise ValueError("Model file does not exist")
+
+        if model_path.suffix in (".pkl", ".pickle"):
+            model = joblib.load(model_path)
+        elif model_path.suffix == ".skops":
+            model = load(model_path)
+        else:
+            msg = (
+                f"Cannot interpret model suffix {model_path.suffix}, should be"
+                " '.pkl', '.pickle' or '.skops'"
+            )
+            raise ValueError(msg)
+
         return model
 
     def _generate_card(self) -> ModelCard:
@@ -429,18 +466,23 @@ class Card:
                     )
                 else:
                     template_sections["get_started_code"] = (
-                        "import joblib\nimport json\nimport pandas as pd\nclf ="
-                        f' joblib.load({model_file})\nwith open("config.json") as'
+                        "import joblib\nimport json\nimport pandas as"
+                        " pd\nclf ="
+                        f" joblib.load({model_file})\nwith"
+                        ' open("config.json") as'
                         " f:\n   "
                         " config ="
                         " json.load(f)\n"
                         'clf.predict(pd.DataFrame.from_dict(config["sklearn"]["example_input"]))'
                     )
         if self.model_diagram is True:
-            model_plot_div = re.sub(r"\n\s+", "", str(estimator_html_repr(self.model)))
+            model_plot_div = re.sub(
+                r"\n\s+", "", str(estimator_html_repr(self.model))
+            )
             if model_plot_div.count("sk-top-container") == 1:
                 model_plot_div = model_plot_div.replace(
-                    "sk-top-container", 'sk-top-container" style="overflow: auto;'
+                    "sk-top-container",
+                    'sk-top-container" style="overflow: auto;',
                 )
             model_plot: str | None = model_plot_div
         else:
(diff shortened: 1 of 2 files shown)


skops/card/_model_card.py
```diff
@@ -372,6 +372,26 @@ def add_metrics(self, **kwargs: str) -> "Card":
...
+        Returns
+        -------
+        model : object 
+            Model instance.
+        """
+        if isinstance(model, str):
+            model_path = Path(model)
+            if not model_path.exists():
+                raise ValueError("Model file does not exist")
+            model = skops.io.load(model)
```
merveenoyan, line 392: As of now, we support both pickle and a safer format called skops. For this to support both models, we need to also put a line that loads the model with joblib.

merveenoyan (changes requested):
Thanks a lot for contributing 🧡 could you please also write tests making sure that it would work for both a model, a model path containing a `pkl` file, and a model path containing `.skops` format?

skops/card/_model_card.py
merveenoyan, line 392: As of now, we also support pickle files. For this to work, it would be good to add a line that loads a pickle. skops.io.load() loads a model file with skops extension, which is essentially a .zip.

merveenoyan (changes requested):

adrinjalali:
> Right now, the model is eagerly loaded during __init__. Recently, we made a move to prevent this (see https://github.com/skops-dev/skops/pull/163). However, lazy loading could be costly because it would be done for each call to __repr__. Maybe we can store the actual model in a different attribute (say, _model) when it's called for the first time?

But we need to make sure we invalidate that whenever the `_model` changes.

BenjaminBossan:
> But we need to make sure we invalidate that whenever the `_model` changes.

This could be done with a setter but maybe we can come up with an easier solution?

p-mishra1:
> 1. We need to have a test for this. If you need help implementing the test, let us know.
Yes sure, there will be same test function but instead of model instance we have to pass a tmp model file (one for skops and one for pkl).
> 2. The docstring of the `Card` class needs to be updated.
> 3. Right now, the model is eagerly loaded during `__init__`. Recently, we made a move to prevent this (see #163). However, lazy loading could be costly because it would be done for each call to `__repr__`. Maybe we can store the _actual_ model in a different attribute (say, `_model`) when it's called for the first time?
On this as the same argument is supporting both model and model_path so in representation and in _generate_card we will be doing model loading there else we can do it in representation on 1st call and set `_model` to model instance but in that case both `model` and '_model' will have the instance when user will provide model instance to `Card` class.

p-mishra1:
Yes @BenjaminBossan i changed the functionality to skops file as well as pkl file and also mentioned some of the things about thr design choice. 
Waiting for input from you guys about that.

BenjaminBossan:
> > 1. We need to have a test for this. If you need help implementing the test, let us know.
> 
> Yes sure, there will be same test function but instead of model instance we have to pass a tmp model file (one for skops and one for pkl).

So here we would like to see unit tests for the new feature being added, similar to what we have [here](https://github.com/skops-dev/skops/blob/main/skops/hub_utils/tests/test_hf_hub.py). Would you please work on that? We can help if needed.

> On this as the same argument is supporting both model and model_path so in representation and in _generate_card we will be doing model loading.
> else we can do it in representation on 1st call and set `_model` to model instance but in that case both `model` and '_model' will have the instance when user will provide model instance to `Card` class.

So I prefer the second solution. When the model is first accessed, if it's a path, it is loaded, if it's already a model, then we're good. It should be stored on a private attribute like `_model`. Then the next time, the model is already loaded and we don't need to load it again. When the user changes the `model` attribute, we need to invalidate `_model`, since it could now be a different model.

BenjaminBossan (changes requested):
Thank you for working on this @p-mishra1 

This is progressing in the right direction, but there is still some work to do. Most significantly, the initial consideration is still not addressed:

> Right now, the model is eagerly loaded during `__init__`. Recently, we made a move to prevent this. However, lazy loading could be costly because it would be done for each call to `__repr__`. Maybe we can **store the actual model in a different attribute (say, `_model`)** when it's called for the first time?

Why is this important? Say, a user creates a card instance with `model_path0` and then later re-assigns the model to `model_path1`, the `_load_model` function will not be called. Then code further down the line will fail.

I created a minimal example that shows how this problem can be solved with `@property` and `setter`:

```python
import pickle
from typing import Any
from pathlib import Path
import joblib
from sklearn.linear_model import LinearRegression, LogisticRegression
from skops.io import load


def _load_model(model: Any) -> Any:
    """Loads the model if provided a file path, if already a model instance,
    return it unmodified.

    Parameters
    ----------
    model : pathlib.path, str, or sklearn estimator
        Path/str or the actual model instance. If a Path or str, loads the model on first call.

    Returns
    -------
    model : object
        Model instance.

    """
    if not isinstance(model, (Path, str)):
        return model

    model_path = Path(model)
    if not model_path.exists():
        raise ValueError("Model file does not exist")

    if model_path.suffix in (".pkl", ".pickle"):
        # I don't think there is any advantage of joblib.load over pickle.load
        model = joblib.load(model_path)
    elif model_path.suffix == ".skops":
        model = load(model_path)
    else:
        msg = f"Cannot interpret model suffix {model_path.suffix}, should be '.pkl', '.pickle' or '.skops'"
        raise ValueError(msg)

    return model


class Card:
    def __init__(self, model):
        self._model = model

    @property
    def model(self):
        model = _load_model(self._model)
        if model is not self._model:
            self._model = model
        return model

    @model.setter
    def model(self, model):
        self._model = model


class TestCardModelAttribute:
    def test_model_estimator(self):
        model0 = LinearRegression()

        card = Card(model0)
        assert card.model is model0

        # re-assigning the model works
        model1 = LogisticRegression()
        card.model = model1
        assert card.model is model1

        # re-assigning back to original works
        card.model = model0
        assert card.model is model0

    def test_model_is_str_pickle(self, tmp_path):
        model0 = LinearRegression(n_jobs=123)
        f_name0 = tmp_path / "lin_reg.pickle"
        with open(f_name0, "wb") as f:
            pickle.dump(model0, f)

        card = Card(f_name0)
        assert isinstance(card.model, LinearRegression)
        assert card.model.n_jobs == 123

        # re-assigning the model works
        model1 = LogisticRegression(n_jobs=456)
        f_name1 = tmp_path / "log_reg.pickle"
        with open(f_name1, "wb") as f:
            pickle.dump(model1, f)

        card.model = f_name1
        assert isinstance(card.model, LogisticRegression)
        assert card.model.n_jobs == 456

        # re-assigning back to original works
        card.model = f_name0
        assert isinstance(card.model, LinearRegression)
        assert card.model.n_jobs == 123
```

This example should work as is. As you can see, I simplified a lot, but it should demonstrate the general idea. Do you think you could refactor your code to work like this?

Also, as you can see, I added some specific tests for this feature. This is preferred over modifying all the existing tests to check both models passed as str or as actual estimators. So could you please undo your changes to the existing tests and instead add tests that are specific to this problem?

PS: We should also add the possibility to pass a `pathlib.Path`, not only `str`, since we generally support that throughout skops.

p-mishra1:
thanks @BenjaminBossan will update the changes

p-mishra1:
thanks @BenjaminBossan for the code snippet and detailed explanation i have added required changes, i also added a deleter for model. have a look.

BenjaminBossan (changes requested):
Thanks a lot for reworking this PR. We're getting closer to the finish line.

I did a light review of your PR and there are a couple of things that need to be changed.

## Formatting

Could you revert the formatting changes you made? It looks like some tool you use (your IDE maybe?) reformatted the files you worked on to use 79 or 80 characters per line. However, we use `black` with the default settings, which comes down to ~88 characters per line. If you undo these changes, the code will be much easier to review because the formatting changes distract from the real changes.

The best way to achieve this would be if you could first undo your formatting changes, then add `black` via a pre-commit hook. This step is explained in our [contribution pipeline](https://github.com/skops-dev/skops/blob/main/CONTRIBUTING.rst#setting-up-the-dev-environment) but the gist is that you should do something like:

```python
pip install pre-commit
pre-commit install
```

from within your Python environment. After doing this, each time when you commit, the code will be checked for `black` compliance.

## Testing

If I understand correctly, you have duplicated a number of tests. Examples are `test_plot_model_false_from_path` and `test_add_from_path`. The change you introduced was to have the card initialized from a file instead of passing the model directly, the rest is identical.

IMO, it is not necessary to duplicate all those tests, since the actual thing they're testing is not related to how the model was passed. Instead, it is sufficient to have tests that show that passing a model or a `str` or a `Path` do the same thing, and that modifying the `.model` attribute works (as in the last test class (which btw. should also test deleting, which you correctly added)). The duplicated tests can then be removed.

Another suggestion I have, which is not as clear cut, is to move `_load_model` out to be a standalone function and not a method on `Card`. Then add a few tests just for this function. Once we know that this function works well, testing the `Card` class will be much easier because we know that the model loading part is covered.

## Test fixtures

I think there is a misunderstanding regarding the use of pytest fixtures. Let's take this example:

```python
@pytest.mark.parametrize("suffix", [".pkl", ".pickle", ".skops"])
def test_save_model_card_from_path(destination_path, model_card_from_path):
    model_card_from_path.save(Path(destination_path) / "README.md")
    assert (Path(destination_path) / "README.md").exists()
```

Here you use the fixture `suffix` but your test doesn't make use of it. This means that the exact same test will be run 3 times, with no additional value. The correct way to use fixtures is to create the decorator (as you did), then add the fixture name as an argument to the function, and then to use that argument in the function (e.g. in `test_plot_model_false_from_path` you use the fixture correctly).

I hope this helps and if you have more questions feel free to ask us.

p-mishra1:
> ## Formatting
> Could you revert the formatting changes you made? It looks like some tool you use (your IDE maybe?) reformatted the files you worked on to use 79 or 80 characters per line. However, we use `black` with the default settings, which comes down to ~88 characters per line. If you undo these changes, the code will be much easier to review because the formatting changes distract from the real changes.
> 
> The best way to achieve this would be if you could first undo your formatting changes, then add `black` via a pre-commit hook. This step is explained in our [contribution pipeline](https://github.com/skops-dev/skops/blob/main/CONTRIBUTING.rst#setting-up-the-dev-environment) but the gist is that you should do something like:
> 
> ```python
> pip install pre-commit
> pre-commit install
> ```
> 
> from within your Python environment. After doing this, each time when you commit, the code will be checked for `black` compliance.


I used the contribution setup before raising the PR but i think my local black setup formatted the code before the being formatted by pre-commit.



> ## Testing
> If I understand correctly, you have duplicated a number of tests. Examples are `test_plot_model_false_from_path` and `test_add_from_path`. The change you introduced was to have the card initialized from a file instead of passing the model directly, the rest is identical.
> 
> IMO, it is not necessary to duplicate all those tests, since the actual thing they're testing is not related to how the model was passed. Instead, it is sufficient to have tests that show that passing a model or a `str` or a `Path` do the same thing, and that modifying the `.model` attribute works (as in the last test class (which btw. should also test deleting, which you correctly added)). The duplicated tests can then be removed.
> 
> Another suggestion I have, which is not as clear cut, is to move `_load_model` out to be a standalone function and not a method on `Card`. Then add a few tests just for this function. Once we know that this function works well, testing the `Card` class will be much easier because we know that the model loading part is covered.
> 

okay, I will remove the duplicate tests and make standalone _load_model function and write tests for funtion as well as passing path.

> ## Test fixtures
> I think there is a misunderstanding regarding the use of pytest fixtures. Let's take this example:
> 
> ```python
> @pytest.mark.parametrize("suffix", [".pkl", ".pickle", ".skops"])
> def test_save_model_card_from_path(destination_path, model_card_from_path):
>     model_card_from_path.save(Path(destination_path) / "README.md")
>     assert (Path(destination_path) / "README.md").exists()
> ```
> 
> Here you use the fixture `suffix` but your test doesn't make use of it. This means that the exact same test will be run 3 times, with no additional value. The correct way to use fixtures is to create the decorator (as you did), then add the fixture name as an argument to the function, and then to use that argument in the function (e.g. in `test_plot_model_false_from_path` you use the fixture correctly).

[Fixture parameterisation](https://stackoverflow.com/questions/18011902/pass-a-parameter-to-a-fixture-function#:~:text=However%2C%20although%20this,False%5D%20FAILED)
here the pytest fixture `model_card_from_path` has a param named `suffix`  and pytest is implicitly passing the values. have a look at the above stackoverflow link, so the pytest behaviour is correct(i checked the test locally and verfied the output and behaviour before the PR ).

BenjaminBossan:
> here the pytest fixture `model_card_from_path` has a param named `suffix` and pytest is implicitly passing the values. have a look at the above stackoverflow link, so the pytest behaviour is correct(i checked the test locally and verfied the output and behaviour before the PR ).

Ah I see, I didn't know about this new feature. It's kinda cool but also a bit magic. I wonder if it's really preferable over using `request.param` to parametrize a fixture.

adrinjalali:
Oh nice, finally pytest supporting this. Been waiting for it for ages. I personally find it more intuitive than using `request.param`. But everywhere this pattern is used, I'd leave a comment saying it's a parameter passed to the fixture for future developers to figure it out easily.

p-mishra1:
Created a fresh PR in #205 for this.

Issue #166, opened 2022-10-03
Make pytest fixtures atomic


merveenoyan:
Some of the pytest fixtures are doing one too many things in model card tests. It would be more convenient if they were atomic. 
e.g.
```python
@pytest.fixture
def model_card_metadata_from_config(destination_path):
    X, y = load_iris(return_X_y=True, as_frame=True)
    est = LogisticRegression(solver="liblinear").fit(X, y)
    pkl_file = tempfile.mkstemp(suffix=".pkl", prefix="skops-test")[1]
    with open(pkl_file, "wb") as f:
        pickle.dump(est, f)
    hub_utils.init(
        model=pkl_file,
        requirements=[f"scikit-learn=={sklearn.__version__}"],
        dst=destination_path,
        task="tabular-classification",
        data=X,
    )
    card = Card(
        est, model_diagram=True, metadata=metadata_from_config(destination_path)
    )
    yield card
```
I had to rewrite this to make it work for tests with serializing with skops format.

E-Aho:
Hey! I was thinking of picking this up

Might be useful to have a discussion about design decisions wrt testing on this but I'm happy to start on it :)

adrinjalali:
@E-Aho could you please elaborate on the "testing philosophy"?

E-Aho:
Ah sure, just things like how deep/thorough the tests should be, testing different edge cases/failure cases, that kind of thing.

Although I have reread the original post and this is more about adjusting already existing tests than writing new ones, so probably not needed on this issue.

adrinjalali:
Yes, this one is a refactoring issue. So you wouldn't be touching test coverage.

We try to have an almost complete coverage on patches.

PR #175, opened 2022-10-05
TST Refactor test fixture to be more atomic

E-Aho:
Small PR to split the responsibilities of the `model_card_metadata_from_config` fixture to be more atomic, with fixtures that

1) Fetch the Iris data
2) Creates an estimator from the Iris data
3) Pickles estimator in a temp file
4) Create a model card based on other fixtures

Notes:
* Could also now refactor the other places where the iris data is called to use the fixture, or could remove the Iris fetch fetch fixture all together and just call `load_iris` in each fixture.

Should resolve https://github.com/skops-dev/skops/issues/166

### DIFF
--- a/skops/card/tests/test_card.py
+++ b/skops/card/tests/test_card.py
@@ -47,7 +47,7 @@ def iris_estimator(iris_data):
 
 
 @pytest.fixture
-def pkl_file(iris_estimator):
+def iris_pkl_file(iris_estimator):
     pkl_file = tempfile.mkstemp(suffix=".pkl", prefix="skops-test")[1]
     with open(pkl_file, "wb") as f:
         pickle.dump(iris_estimator, f)
@@ -55,7 +55,7 @@ def pkl_file(iris_estimator):
 
 
 @pytest.fixture
-def skops_file(iris_estimator):
+def iris_skops_file(iris_estimator):
     skops_folder = tempfile.mkdtemp()
     model_name = "model.skops"
     skops_path = Path(skops_folder) / model_name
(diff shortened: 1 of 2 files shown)


merveenoyan (approved):
This looks very neat & clean, thank you ❤️🙂 
This is not obligatory but can you check if anything else could be refactored with the new fixtures?

E-Aho:
> This looks very neat & clean, thank you ❤️🙂 This is not obligatory but can you check if anything else could be refactored in the main tests, given that we have new fixtures?

Thanks ❤️ 😄 

I can definitely do that, I'll add another commit to this later today to refactor places that can now use these fixtures

E-Aho:
Updated the skops test that I think @merveenoyan was talking about in the original issue, all now use the fixtures so there's minimal set up in the tests themselves :)

merveenoyan:
@BenjaminBossan yes it does!

merveenoyan (approved):
Overall looks very good to me! @BenjaminBossan can you merge if you approve too?

skops/card/tests/test_card.py
```diff
@@ -34,23 +34,70 @@ def model_card(model_diagram=True):
...
+
+@pytest.fixture
+def iris_estimator(iris_data):
+    X, y = iris_data
     est = LogisticRegression(solver="liblinear").fit(X, y)
+    yield est
+
+
+@pytest.fixture
+def pkl_file(iris_estimator):
```
adrinjalali, line 50: maybe call this `iris_pkl_file`?

skops/card/tests/test_card.py
```diff
@@ -34,23 +34,70 @@ def model_card(model_diagram=True):
...
     pkl_file = tempfile.mkstemp(suffix=".pkl", prefix="skops-test")[1]
     with open(pkl_file, "wb") as f:
-        pickle.dump(est, f)
+        pickle.dump(iris_estimator, f)
+    yield pkl_file
+
+
+@pytest.fixture
+def skops_file(iris_estimator):
```
adrinjalali, line 58: same here 
```suggestion
def iris_skops_file(iris_estimator):
```

adrinjalali (commented):
otherwise LGTM

adrinjalali (approved):

Issue #164, opened 2022-10-02
Edit style of model plot to fit inside the model card 


merveenoyan:
Model plot sometimes overflows if there are too many preprocessors.

 <img width="1360" alt="Ekran Resmi 2022-10-03 00 35 47" src="https://user-images.githubusercontent.com/53175384/193479325-84604631-79cc-450f-9192-42674503a768.png">

See more detailed explanation below: 
![Ekran Resmi 2022-10-06 16 50 02](https://user-images.githubusercontent.com/53175384/194345479-2cb7034c-5835-46b0-b9c0-d31023fe8d95.png)

adrinjalali:
How is this a good first issue? I have no idea how to fix this lol

merveenoyan:
@adrinjalali if you know front-end 😂 I think the top level container's class tag needs to include width and that's it. (not sure though)

rushic24:
Where is it overflowing in the screenshot?

merveenoyan:
@rushic24 the model plot above is overflowing from the model card (even covering the spaces that use this model section). We need to edit the top level container to have a width that would fit in model class. Can you handle it? This is easy if you know front-end well.

merveenoyan:
@rushic24 you can edit the resulting plot, no?

rushic24:
That's what I was going to check, can I have the overflow card snippet?

rushic24:
Ok just found the resulting plot is editable, but I still feel this would be better if fixed in sklearn. I can try asking there and implement it here if they don't agree.

rushic24:
It appears ok on a wide monitor LOL, any idea how can I replicate the bug on my system?
![image](https://user-images.githubusercontent.com/6279035/194342872-4a028ea3-576e-45a7-ace5-72f5000f78fe.png)

merveenoyan:
@rushic24 I wanted to show the overflowing part below 👇 it happens on your monitor as well. The left part is model card, it's a markdown file. right part belongs to widget and links to spaces.
![Ekran Resmi 2022-10-06 16 50 02](https://user-images.githubusercontent.com/53175384/194345282-88ea4eef-5e78-4707-82b8-300ab04464ca.png)

rushic24:
Ahh now I get it, then this isn't a sklearn issue.

merveenoyan:
@rushic24 sorry for not communicating it well enough above, it looks a little cryptic 🥲

rushic24:
[...]

PR #178, opened 2022-10-08
FIX Edit style of model plot to fit in the page

rushic24:
# How it will look:
Readme example here :- https://huggingface.co/rushic24/TestPlaygroundSkops

Fixes https://github.com/skops-dev/skops/issues/164

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -410,9 +410,12 @@ class Card:
                         'clf.predict(pd.DataFrame.from_dict(config["sklearn"]["example_input"]))'
                     )
         if self.model_diagram is True:
-            model_plot: str | None = re.sub(
+            model_plot_div = re.sub(
                 r"\n\s+", "", str(estimator_html_repr(self.model))
             )
+            assert model_plot_div.count("sk-top-container") == 1
+            model_plot_div = model_plot_div.replace("sk-top-container", 'sk-top-container" style="overflow: auto;')
+            model_plot: str | None = model_plot_div
         else:
             model_plot = None
         template_sections["eval_results"] = tabulate(
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -410,11 +410,11 @@ class Card:
                         'clf.predict(pd.DataFrame.from_dict(config["sklearn"]["example_input"]))'
                     )
         if self.model_diagram is True:
-            model_plot_div = re.sub(
-                r"\n\s+", "", str(estimator_html_repr(self.model))
-            )
-            assert model_plot_div.count("sk-top-container") == 1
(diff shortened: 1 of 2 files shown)


skops/card/_model_card.py
```diff
@@ -410,9 +410,12 @@ def _generate_card(self) -> ModelCard:
                         'clf.predict(pd.DataFrame.from_dict(config["sklearn"]["example_input"]))'
                     )
         if self.model_diagram is True:
-            model_plot: str | None = re.sub(
+            model_plot_div = re.sub(
                 r"\n\s+", "", str(estimator_html_repr(self.model))
             )
+            assert model_plot_div.count("sk-top-container") == 1
```
adrinjalali, line 416: production code should never have `assert` in it. Instead, you can check the value, and handle the case where it doesn't satisfy the condition.

adrinjalali (commented):
Thanks for the PR @rushic24. We don't have a testing framework for these issues, but could you please create a model repo with this change so that we can see the change in action?

adrinjalali:
You also need to apply linting (`black` and `isort`, although only `black` is relevant in this case).

You can enable them in your pre-commit hooks as explain here: https://github.com/skops-dev/skops/blob/main/CONTRIBUTING.rst#using-condamamba

merveenoyan:
(x-posting from the issue) @rushic24 given this is about model card README in 🤗 Hub it would be nice to have a model with a lot of preprocessors in the pipeline (given it's making the plot bigger) inside a model repository that we can look at.

skops/card/_model_card.py
```diff
@@ -410,9 +410,12 @@ def _generate_card(self) -> ModelCard:
...
+            model_plot_div = model_plot_div.replace("sk-top-container", 'sk-top-container" style="overflow: auto;')
```
merveenoyan, line 417: Neat 🙂 
Maybe a test that asserts this overflow style is in model card would be nice! (you can look from some model card tests for inspiration) If you need any help in writing tests please let us know hug🤗

merveenoyan (commented):
Thanks a lot for working on it 🤗

merveenoyan:
@rushic24 Lastly, can you post two repositories, one that has a big pipeline plot and another one that has a small one? (given we can't check how Hub looks on our side 😅 just to make sure both cases are fine!)

rushic24:
Added in the Readme below https://huggingface.co/rushic24/TestPlaygroundSkops
Are the macos tests failing serious issue?

adrinjalali:
It looks a bit narrow to me, as in, the container seems to have more space on the right side that we could use, but I'm not sure how easy it is to fix that and if it's not, I'm happy with the solution here.

merveenoyan:
@adrinjalali it ends where the right most boundary of other components end, no? Added a photo below: image
I think it's good.

adrinjalali (approved):

Issue #180, opened 2022-10-10
Persistence: save to and load from str/bytes


BenjaminBossan:
It would be nice to have the equivalent of `pickle.dumps` / `pickle.loads` or `json.dumps` / `json.loads` for `skops.save` / `skops.load`. That way, saving and loading does not necessarily require dealing with files. It could also facilitate testing and sending data over the wire.

adrinjalali:
That makes sense, then we probably should also rename `save` to `dump` for consistency. I'm +1 on this.

BenjaminBossan:
> then we probably should also rename `save` to `dump`

We can do that but I never liked the name "dump", it kinda has a negative connotation, doesn't it? Numpy also uses `save` and `load` but API-wise, we're closer to `pickle` and `json`, so I'm not sure.

adrinjalali:
Since this should ideally be a replacement for `pickle`, I think following that API would make more sense.

PR #182, opened 2022-10-10
ENH Add dumps and loads functions

BenjaminBossan:
Resolves issue #180. Quoting:

It would be nice to have the equivalent of `pickle.dumps` / `pickle.loads` or `json.dumps` / `json.loads` for `skops.save` / `skops.load`. That way, saving and loading does not necessarily require dealing with files. It could also facilitate testing and sending data over the wire.

## Implementation

This is still very much WIP but I wanted to get some early feedback by @skops-dev/maintainers. The idea is to reuse as much code as possible.

The main changes are around saving numpy arrays and sparse matrices, which now have to deal with a third option. The solution I chose involves a rather ugly roundtrip from bytes > base 64 encoded bytes > string and back, since json can't deal with bytes. If there is a better solution, please let me know. Also, since we don't save to file, there is no memoization step involved.

Regarding the tests, I have modified them so that they now run once with `save`/`load` and once with `dumps`/`loads`. This of course doubles the time required to run those tests, so in the future we may only run one setting (`dumps` and `loads`) and then have a handful of tests that specifically test `save`/`load`.

### DIFF
--- a/skops/io/_numpy.py
+++ b/skops/io/_numpy.py
@@ -1,6 +1,5 @@
 from __future__ import annotations
 
-import base64
 import io
 from typing import Any
 
@@ -31,7 +30,8 @@ def ndarray_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
             np.save(data_buffer, obj)
             obj_id = save_state.memoize(obj)  # TODO: useless
             f_name = f"{obj_id}.npy"
-            save_state.zip_file.writestr(f_name, data_buffer.getbuffer())
+            if f_name not in save_state.zip_file.namelist():
+                save_state.zip_file.writestr(f_name, data_buffer.getbuffer())
             res.update(type="numpy", file=f_name)
     except ValueError:
         # Couldn't save the numpy array with either method
@@ -45,16 +45,6 @@ def ndarray_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
 
 
 def ndarray_get_instance(state, src):
-    if state["type"] == "base64":
-        # TODO
-        b64 = state["content"].encode("utf-8")
-        b = io.BytesIO(base64.b64decode(b64))
-        val = np.load(b)
-        if state["__class__"] != "ndarray":
-            cls = _import_obj(state["__module__"], state["__class__"])
-            val = cls(val)
-        return val
-
     # Dealing with a regular numpy array, where dtype != object
     if state["type"] == "numpy":
         val = np.load(io.BytesIO(src.read(state["file"])), allow_pickle=False)
--- a/skops/io/_numpy.py
+++ b/skops/io/_numpy.py
@@ -1,5 +1,6 @@
 from __future__ import annotations
 
+import base64
 import io
 from typing import Any
 
@@ -25,6 +26,14 @@ def ndarray_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
             res["content"] = obj_serialized["content"]
             res["type"] = "json"
             res["shape"] = get_state(obj.shape, save_state)
+        elif save_state.path is None:
+            # using skops.io.dumps, don't write to file
+            f = io.BytesIO()
+            np.save(f, obj)
+            b = f.getvalue()
+            b64 = base64.b64encode(b)
+            s = b64.decode("utf-8")
+            res.update(content=s, type="base64")
         else:
             # Memoize the object and then check if it's file name (containing
             # the object id) already exists. If it does, there is no need to
@@ -49,6 +58,16 @@ def ndarray_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
 
 
 def ndarray_get_instance(state, src):
+    if state["type"] == "base64":
+        # TODO
+        b64 = state["content"].encode("utf-8")
+        b = io.BytesIO(base64.b64decode(b64))
+        val = np.load(b)
+        if state["__class__"] != "ndarray":
+            cls = _import_obj(state["__module__"], state["__class__"])
+            val = cls(val)
+        return val
+
     # Dealing with a regular numpy array, where dtype != object
     if state["type"] == "numpy":
         val = np.load(io.BytesIO(src.read(state["file"])), allow_pickle=False)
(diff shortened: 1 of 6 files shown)


adrinjalali:
Instead of this, I think we should refactor to pass a `zipfile` instead of a path as `dst`, and then use a `ButesIO` as the `file` arg to the `ZipFile`, and then we have everything in memory. This would also probably solve our windows deleting/moving file issue.

adrinjalali:
Also, I think the same PR should rename `save` to `dump`

BenjaminBossan:
> Instead of this, I think we should refactor to pass a `zipfile` instead of a path as `dst`, and then use a `ButesIO` as the `file` arg to the `ZipFile`, and then we have everything in memory. This would also probably solve our windows deleting/moving file issue.

I'm not sure I quite understand the suggestion. I assume that you want to avoid having a third code path for `dumps` in functions like `ndarray_get_state`, instead having `np.save` just write to an in-memory byte stream. I attempted something like this first and didn't succeed, but maybe I just overlooked a possible solution.

> Also, I think the same PR should rename `save` to `dump`

I currently have `save = dump` in `persist.py`. I'd like to avoid renaming in this PR to keep the diff noise down, would do it in the next PR once this is merged. WDYT?

adrinjalali:
I tried a prototype, and seems like this is how we could write in memory, and at the end save things to file:

```py
import base64
from ast import Bytes
from io import BytesIO
from zipfile import ZipFile

import numpy as np

X = np.random.rand(5, 5)

# This part happens in dumps
buffer = BytesIO()
with ZipFile(buffer, "w") as zip:
	# This part happens in ndarray_getstate(data, save_state)
    data_buffer = BytesIO()
    np.save(data_buffer, X)
    zip.writestr("test.npy", data=data_buffer.getbuffer())

# back to dumps
print(base64.b64encode(buffer.getbuffer()))

# and this in dump
with open("/tmp/test.skops", "wb") as f:
    f.write(buffer.getbuffer())
```

Does this help?

> I currently have `save = dump` in `persist.py`. I'd like to avoid renaming in this PR to keep the diff noise down, would do it in the next PR once this is merged. WDYT?

I could live with that 👍🏼

BenjaminBossan:
> Does this help?

Okay, I see now what you mean. I managed to cobble something together along those lines that makes the all tests pass (except for `test_identical_numpy_arrays_not_duplicated`). It is more elegant for sure than the solution I came up with.

That being said, I think there are some disadvantages to this approach:

1. Saving the numpy arrays will always create a copy of the data in memory. That means that if the estimator has some big numpy arrays, users could run out of memory (this will always be true for `dumps` but not for the current implementation of `dump`).
2. Because of that, there is also no way to avoid copies of the same array, so the whole memoization part doesn't work.
3. With the current approach, the output of `dumps` is a kinda human readable jsonified string, with only numpy arrays being bytes. With the new approach, the output as just a blob. I'm not sure how important readability is here.

I think none of the disadvantages are so severe that we couldn't live with them, but the tradeoff is worth discussing. Or maybe we can find a third way that combines the advantages of both.

BenjaminBossan:
@adrinjalali I cleaned up and pushed the changes (still very much WIP). This should make it easier to compare the implementations.

skops/io/_numpy.py
```diff
@@ -26,16 +27,11 @@ def ndarray_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
...
             res["shape"] = get_state(obj.shape, save_state)
         else:
-            # Memoize the object and then check if it's file name (containing
-            # the object id) already exists. If it does, there is no need to
-            # save the object again. Memoizitation is necessary since for
-            # ephemeral objects, the same id might otherwise be reused.
-            obj_id = save_state.memoize(obj)
+            data_buffer = io.BytesIO()
+            np.save(data_buffer, obj)
+            obj_id = save_state.memoize(obj)  # TODO: useless
```
adrinjalali, line 32: I don't understand why this is now useless.

skops/io/_numpy.py
```diff
@@ -26,16 +27,11 @@ def ndarray_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
             res["type"] = "json"
             res["shape"] = get_state(obj.shape, save_state)
         else:
-            # Memoize the object and then check if it's file name (containing
-            # the object id) already exists. If it does, there is no need to
-            # save the object again. Memoizitation is necessary since for
-            # ephemeral objects, the same id might otherwise be reused.
-            obj_id = save_state.memoize(obj)
+            data_buffer = io.BytesIO()
+            np.save(data_buffer, obj)
```
adrinjalali, line 31: this creates a duplicate, and we can fix it, but I think it's okay for the first version to do it this way and later optimize for memory.

skops/io/_numpy.py
```diff
@@ -49,6 +45,16 @@ def ndarray_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
...
 def ndarray_get_instance(state, src):
+    if state["type"] == "base64":
+        # TODO
+        b64 = state["content"].encode("utf-8")
+        b = io.BytesIO(base64.b64decode(b64))
+        val = np.load(b)
+        if state["__class__"] != "ndarray":
+            cls = _import_obj(state["__module__"], state["__class__"])
+            val = cls(val)
+        return val
```
adrinjalali, line 56: why would we ever save in base64?

adrinjalali (commented):
overall it looks quite good to me.

skops/io/_numpy.py
BenjaminBossan: found a simple way, working on it

skops/io/_numpy.py
BenjaminBossan: It's useless because no memoization takes place, but I'm working on it.

skops/io/_numpy.py
BenjaminBossan: This was a leftover from earlier, I'll remove it.

BenjaminBossan:
@adrinjalali I think I fixed all remaining issues, please review again. After this is merged, I can do the `save` => `dump` renaming.

adrinjalali (commented):
Otherwise LGTM. The codecov might be legit, but I'll ignore it for now and fix later.

adrinjalali (approved):

Issue #184, opened 2022-10-11
Bug: loading certain scipy functions fails


BenjaminBossan:
Certain scipy functions cannot be loaded correctly. Here is a test to reproduce:

```python
from scipy import stats

def test_scipy_stats(tmp_path):
    estimator = FunctionTransformer(func=stats.zipf)
    save_load_round(estimator, tmp_path / "file.skops")
```

`zipf` is only one of a few failing scipy functions that I tested. The error message is

> AttributeError: module 'builtins' has no attribute 'method'

with a really long, unhelpful stacktrace. However, the problem already occurs during the generation of the `state`, which looks really weird:

<details>
<summary> Click to expand </summary>

```json
{
  "__class__": "FunctionTransformer",
  "__module__": "sklearn.preprocessing._function_transformer",
  "content": {
    "__class__": "dict",
    "__module__": "builtins",
    "content": {
      "func": {
        "__class__": "zipf_gen",
        "__module__": "scipy.stats._discrete_distns",
        "content": {
          "__class__": "dict",
          "__module__": "builtins",
          "content": {
            "_stats_has_moments": {
              "__class__": "str",
              "__module__": "builtins",
              "content": "true",
              "is_json": true
            },
            "_random_state": {
              "__class__": "RandomState",
              "__module__": "numpy.random.mtrand",
              "content": {
                "__class__": "dict",
                "__module__": "builtins",
                "content": {
                  "bit_generator": {
                    "__class__": "str",
                    "__module__": "builtins",
                    "content": "\"MT19937\"",
                    "is_json": true
                  },
                  "state": {
                    "__class__": "dict",
                    "__module__": "builtins",
                    "content": {
                      "key": {
                        "__class__": "ndarray",
                        "__module__": "numpy",
                        "type": "numpy",
                        "file": "140209309201776.npy"
                      },
                      "pos": {
                        "__class__": "str",
                        "__module__": "builtins",
                        "content": "335",
                        "is_json": true
                      }
                    },
                    "key_types": {
                      "__class__": "list",
                      "__module__": "builtins",
                      "content": [...]
                    }
                  },
                  "has_gauss": {
                    "__class__": "str",
                    "__module__": "builtins",
                    "content": "0",
                    "is_json": true
                  },
                  "gauss": {
                    "__class__": "str",
                    "__module__": "builtins",
                    "content": "0.0",
                    "is_json": true
                  }
                },
                "key_types": {
                  "__class__": "list",
                  "__module__": "builtins",
[...]

PR #195, opened 2022-10-17
Add Method type in dispatch calls

E-Aho:
Fixes https://github.com/skops-dev/skops/issues/184

PR which adds MethodType and related functions to the skops.io dispatch functions.

I can expand on the test if more cases/checks are needed, but for now this shows the functionality works as expected.

### DIFF
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -33,11 +33,11 @@ def dict_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def dict_get_instance(state, src):
+def dict_get_instance(state, src, load_state):
     content = gettype(state)()
-    key_types = get_instance(state["key_types"], src)
+    key_types = get_instance(state["key_types"], src, load_state)
     for k_type, item in zip(key_types, state["content"].items()):
-        content[k_type(item[0])] = get_instance(item[1], src)
+        content[k_type(item[0])] = get_instance(item[1], src, load_state)
     return content
 
 
@@ -54,10 +54,10 @@ def list_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def list_get_instance(state, src):
+def list_get_instance(state, src, load_state):
     content = gettype(state)()
     for value in state["content"]:
-        content.append(get_instance(value, src))
+        content.append(get_instance(value, src, load_state))
     return content
 
 
@@ -72,7 +72,7 @@ def tuple_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def tuple_get_instance(state, src):
+def tuple_get_instance(state, src, load_state):
     # Returns a tuple or a namedtuple instance.
     def isnamedtuple(t):
         # This is needed since namedtuples need to have the args when
@@ -86,7 +86,7 @@ def tuple_get_instance(state, src):
         return all(type(n) == str for n in f)
 
     cls = gettype(state)
-    content = tuple(get_instance(value, src) for value in state["content"])
+    content = tuple(get_instance(value, src, load_state) for value in state["content"])
 
     if isnamedtuple(cls):
         return cls(*content)
@@ -106,7 +106,7 @@ def function_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def function_get_instance(state, src):
+def function_get_instance(state, src, load_state):
     loaded = _import_obj(state["content"]["module_path"], state["content"]["function"])
     return loaded
 
@@ -127,12 +127,12 @@ def partial_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def partial_get_instance(state, src):
+def partial_get_instance(state, src, load_state):
     content = state["content"]
-    func = get_instance(content["func"], src)
-    args = get_instance(content["args"], src)
-    kwds = get_instance(content["kwds"], src)
-    namespace = get_instance(content["namespace"], src)
+    func = get_instance(content["func"], src, load_state)
+    args = get_instance(content["args"], src, load_state)
+    kwds = get_instance(content["kwds"], src, load_state)
+    namespace = get_instance(content["namespace"], src, load_state)
     instance = partial(func, *args, **kwds)  # always use partial, not a subclass
     instance.__setstate__((func, args, kwds, namespace))
     return instance
@@ -153,7 +153,7 @@ def type_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def type_get_instance(state, src):
+def type_get_instance(state, src, load_state):
     loaded = _import_obj(state["content"]["__module__"], state["content"]["__class__"])
     return loaded
 
@@ -172,7 +172,7 @@ def slice_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def slice_get_instance(state, src):
+def slice_get_instance(state, src, load_state):
     start = state["content"]["start"]
     stop = state["content"]["stop"]
     step = state["content"]["step"]
@@ -190,6 +190,7 @@ def object_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
             "__class__": "str",
             "__module__": "builtins",
             "__loader__": "none",
+            "__id__": id(obj),
             "content": obj_str,
             "is_json": True,
         }
@@ -200,6 +201,7 @@ def object_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
         "__class__": obj.__class__.__name__,
         "__module__": get_module(type(obj)),
         "__loader__": "object_get_instance",
+        "__id__": id(obj),
     }
 
     # __getstate__ takes priority over __dict__, and if non exist, we only save
@@ -218,7 +220,7 @@ def object_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def object_get_instance(state, src):
+def object_get_instance(state, src, load_state):
     if state.get("is_json", False):
         return json.loads(state["content"])
 
@@ -233,7 +235,7 @@ def object_get_instance(state, src):
     if not content:  # nothing more to do
         return instance
 
-    attrs = get_instance(content, src)
+    attrs = get_instance(content, src, load_state)
     if hasattr(instance, "__setstate__"):
         instance.__setstate__(attrs)
     else:
@@ -253,15 +255,20 @@ def method_get_state(obj: Any, save_state: SaveState):
         "__loader__": "method_get_instance",
         "content": {
             "func": obj.__func__.__name__,
-            "obj": get_state(obj.__self__, save_state),
+            "__id__": id(obj.__self__),
         },
     }
-
+    root_obj = obj.__self__
+    obj_state = save_state.get_memoized_state(root_obj)
+    if obj_state is None:
+        obj_state = get_state(root_obj, save_state)
+        save_state.store_state(root_obj, obj_state)
+    res["content"]["obj"] = obj_state
     return res
 
 
-def method_get_instance(state, src):
-    loaded_obj = object_get_instance(state["content"]["obj"], src)
+def method_get_instance(state, src, load_state):
+    loaded_obj = get_instance(state["content"]["obj"], src, load_state)
     method = getattr(loaded_obj, state["content"]["func"])
     return method
 
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import json
 from functools import partial
-from types import FunctionType
+from types import FunctionType, MethodType
 from typing import Any
 
 import numpy as np
@@ -98,6 +98,7 @@ def function_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
             "function": obj.__name__,
         },
     }
+
     return res
 
 
@@ -232,6 +233,26 @@ def object_get_instance(state, src):
     return instance
 
 
+def method_get_state(obj: Any, save_state: SaveState):
+    res = {
+        "__class__": obj.__class__.__name__,
+        "__module__": get_module(obj),
+        "content": {
+            "func": obj.__func__.__name__,
+            "obj": obj.__self__.__class__.__name__,
+            "module_path": get_module(obj.__self__.__class__),
+        },
+    }
+    return res
+
+
+def method_get_instance(state, src):
+    # TODO: init with attrs
+    loaded_obj = _import_obj(state["content"]["module_path"], state["content"]["obj"])()
+    method = getattr(loaded_obj, "_entropy")
+    return method
+
+
 def unsupported_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     raise UnsupportedTypeException(obj)
 
@@ -243,6 +264,7 @@ GET_STATE_DISPATCH_FUNCTIONS = [
     (tuple, tuple_get_state),
     (slice, slice_get_state),
     (FunctionType, function_get_state),
+    (MethodType, method_get_state),
     (partial, partial_get_state),
     (type, type_get_state),
     (object, object_get_state),
@@ -254,6 +276,7 @@ GET_INSTANCE_DISPATCH_FUNCTIONS = [
     (tuple, tuple_get_instance),
     (slice, slice_get_instance),
     (FunctionType, function_get_instance),
+    (MethodType, method_get_instance),
     (partial, partial_get_instance),
     (type, type_get_instance),
     (object, object_get_instance),
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -239,16 +239,15 @@ def method_get_state(obj: Any, save_state: SaveState):
         "__module__": get_module(obj),
         "content": {
             "func": obj.__func__.__name__,
-            "obj": obj.__self__.__class__.__name__,
-            "module_path": get_module(obj.__self__.__class__),
+            "obj": object_get_state(obj.__self__, save_state),
         },
     }
+
     return res
 
 
 def method_get_instance(state, src):
-    # TODO: init with attrs
-    loaded_obj = _import_obj(state["content"]["module_path"], state["content"]["obj"])()
+    loaded_obj = object_get_instance(state["content"]["obj"], src)
     method = getattr(loaded_obj, state["content"]["func"])
     return method
 
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -33,11 +33,11 @@ def dict_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def dict_get_instance(state, src, load_state):
+def dict_get_instance(state, src):
     content = gettype(state)()
-    key_types = get_instance(state["key_types"], src, load_state)
+    key_types = get_instance(state["key_types"], src)
     for k_type, item in zip(key_types, state["content"].items()):
-        content[k_type(item[0])] = get_instance(item[1], src, load_state)
+        content[k_type(item[0])] = get_instance(item[1], src)
     return content
 
 
@@ -54,10 +54,10 @@ def list_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def list_get_instance(state, src, load_state):
+def list_get_instance(state, src):
     content = gettype(state)()
     for value in state["content"]:
-        content.append(get_instance(value, src, load_state))
+        content.append(get_instance(value, src))
     return content
 
 
@@ -72,7 +72,7 @@ def tuple_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def tuple_get_instance(state, src, load_state):
+def tuple_get_instance(state, src):
     # Returns a tuple or a namedtuple instance.
     def isnamedtuple(t):
         # This is needed since namedtuples need to have the args when
@@ -86,7 +86,7 @@ def tuple_get_instance(state, src, load_state):
         return all(type(n) == str for n in f)
 
     cls = gettype(state)
-    content = tuple(get_instance(value, src, load_state) for value in state["content"])
+    content = tuple(get_instance(value, src) for value in state["content"])
 
     if isnamedtuple(cls):
         return cls(*content)
@@ -106,7 +106,7 @@ def function_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def function_get_instance(state, src, load_state):
+def function_get_instance(state, src):
     loaded = _import_obj(state["content"]["module_path"], state["content"]["function"])
     return loaded
 
@@ -127,12 +127,12 @@ def partial_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def partial_get_instance(state, src, load_state):
+def partial_get_instance(state, src):
     content = state["content"]
-    func = get_instance(content["func"], src, load_state)
-    args = get_instance(content["args"], src, load_state)
-    kwds = get_instance(content["kwds"], src, load_state)
-    namespace = get_instance(content["namespace"], src, load_state)
+    func = get_instance(content["func"], src)
+    args = get_instance(content["args"], src)
+    kwds = get_instance(content["kwds"], src)
+    namespace = get_instance(content["namespace"], src)
     instance = partial(func, *args, **kwds)  # always use partial, not a subclass
     instance.__setstate__((func, args, kwds, namespace))
     return instance
@@ -153,7 +153,7 @@ def type_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def type_get_instance(state, src, load_state):
+def type_get_instance(state, src):
     loaded = _import_obj(state["content"]["__module__"], state["content"]["__class__"])
     return loaded
 
@@ -172,7 +172,7 @@ def slice_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
     return res
 
 
-def slice_get_instance(state, src, load_state):
+def slice_get_instance(state, src):
     start = state["content"]["start"]
     stop = state["content"]["stop"]
     step = state["content"]["step"]
@@ -190,7 +190,6 @@ def object_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
             "__class__": "str",
             "__module__": "builtins",
             "__loader__": "none",
-            "__id__": id(obj),
             "content": obj_str,
             "is_json": True,
         }
@@ -201,7 +200,6 @@ def object_get_state(obj: Any, save_state: SaveState) -> dict[str, Any]:
         "__class__": obj.__class__.__name__,
         "__module__": get_module(type(obj)),
         "__loader__": "object_get_instance",
-        "__id__": id(obj),
     }
 
     # __getstate__ takes priority over __dict__, and if non exist, we only save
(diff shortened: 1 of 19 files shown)


E-Aho:
NB: The added test is just a placeholder to show that it resolves #184

BenjaminBossan:
Well done figuring out where the error occurs.

Before we continue, I really want to understand what is going on here and if the suggested solution is the best one.

So first, it's true that we don't have any special case for objects whose attributes are methods bound _to another object_. E.g. this fails for now:

```python
class Foo(BaseEstimator):
    def fit(self, X, y=None, **fit_params):
        self.bound_method_ = FunctionTransformer().transform
        return self
    def transform(self, X):
        return self.bound_method_(X)

def test_bound_method(tmp_path):
    estimator = Foo().fit(None)
    f_name = tmp_path / "file.skops"
    loaded = save_load_round(estimator, f_name)
```

gives:

```
    def object_get_instance(state, src):
        if state.get("is_json", False):
            return json.loads(state["content"])
    
        cls = gettype(state)
>       instance = cls.__new__(cls)
E       TypeError: method expected 2 arguments, got 0
```

I think we should add a test along these lines to our test suite.

How does this relate to the original issue? From what I can tell, the problem stems from these lines:

https://github.com/scipy/scipy/blob/1023d9207fdc1430a8ba196f1a1616ac3c264acf/scipy/stats/_distn_infrastructure.py#L3183-L3186

The `vecentropy` attribute is the one causing the issue. It is a `numpy.vectorize` of a bound method, but it's not actually bound to another instance, it's bound to itself! Still, it is recognized as a bound method, hence dispatches to the function that @E-Aho implemented.

What is strange to me is that the same issue does not apply to `self._cdfvec`, even though it's treated the same as `vecentropy`.

Another thing that surprised me is the following: I removed the whole `vecentropy` object from the state (I did this by going inside of `dict_get_state` and if the key is `"vecentropy"`, I just skip it). I would have thought that this cannot work, but it does, the test passes without requiring `method_get_state/instance`, and the `vecentropy` attribute exists after loading.

I dug into this and I think it's because `_attach_methods` is called during `__setstate__`:

https://github.com/scipy/scipy/blob/1023d9207fdc1430a8ba196f1a1616ac3c264acf/scipy/stats/_distn_infrastructure.py#L636

If that's true, we should be able to drop those methods entirely from the state and let scipy restore them during `__setstate__`. WDYT?

Regardless of that, as mentioned at the top, we should have a way to save and load methods bound to other instances.

I also wondered whether we can add a special case for the `numpy.vectorize` class, similar to what we have with `partial`. Maybe that can solve the initial problem, but it wouldn't solve the general bound method issue.

E-Aho:
> If that's true, we should be able to drop those methods entirely from the state and let scipy restore them during __setstate__. WDYT?

I fully agree, I ran into this as well while testing, but with "dummy functions" that returned nothing, a few other tests that checked on the dispatch functions started failing too.

I think in this case, it could definitely be solved by dropping these methods and adding a special condition to check for this during loading.

However, in a general case, I think there still needs to be a solution for bound methods more generally, which I think would be along the lines of this PR. E.g:

```
    class MethodHolder:
        def __init__(self, variant):
            if variant == "sqrt":
                self._method = np.sqrt
            elif variant == "log":
                self._method = np.log
            else:
                self._method = np.exp

        def func(self, x):
            return  self._method(x)

    a = MethodHolder("log")
    b = a.func
 
```

I feel like something along these lines isn't unlikely to exist somewhere in SKLearn/SciPy (given we've already found one that happens to get bounded during init), and it would probably be a good idea to provide support for serializing it correctly, which shouldn't be too hard to do.

BenjaminBossan:
> I think there still needs to be a solution for bound methods more generally, which I think would be along the lines of this PR

:+1: This is exactly what I meant when I wrote

> we should have a way to save and load methods bound to other instances

We have to be precise with the wording here, we already persist objects with bound methods without issue, the problem are attributes that are methods bound to _another_ object than the one we're persisting.

Once we have this general issue solved, if it doesn't solve the scipy issue, we can think about a more specific solution there.

E-Aho:
> We have to be precise with the wording here, we already persist objects with bound methods without issue, the problem are attributes that are methods bound to _another_ object than the one we're persisting.

Ah sorry, you're entirely correct. I miswrote it in my comment, but I was meaning to show that if you used that `b` in a different object or function, then serialized *that*, that would be where this bug exists (to my understanding).

e.g if we had 
```
    a = MethodHolder("log")
    b = a.func
    c = SomeOtherObject(func=b)
```

This would happen I believe.

I think one other place this might occur for users is when trying to serialize something that can take a function as an input (e.g [DistanceMetrics with a user defined distance](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.DistanceMetric.html?highlight=pyfunc#)). 

If that function is bound to a different object, it wouldn't work right now AFAIK

BenjaminBossan:
> And tried to serialise `c`, this would happen I believe.

Yes, I think it would fail. Is there a qualitative difference to the other example I posted:

> ```python
> class Foo(BaseEstimator):
>     def fit(self, X, y=None, **fit_params):
>         self.bound_method_ = FunctionTransformer().transform
>         return self
>     def transform(self, X):
>         return self.bound_method_(X)
> ```

BenjaminBossan:
> CI runner is failing with a 404 error on uploading to CodeCov thinking

Most of the time, it's just some flaky behavior that can be fixed by rerunning the CI.

E-Aho:
Should be ready for review when you have a moment @BenjaminBossan :) let me know if you have any thoughts!

skops/io/_general.py
```diff
@@ -232,6 +233,25 @@ def object_get_instance(state, src):
     return instance
 
 
+def method_get_state(obj: Any, save_state: SaveState):
```
BenjaminBossan, line 236: Since it's not quite obvious, I would like to have a comment here that this is for methods bound to other instances, for future reference.

skops/io/_general.py
```diff
@@ -232,6 +233,25 @@ def object_get_instance(state, src):
...
+    res = {
+        "__class__": obj.__class__.__name__,
+        "__module__": get_module(obj),
+        "content": {
+            "func": obj.__func__.__name__,
+            "obj": object_get_state(obj.__self__, save_state),
```
BenjaminBossan, line 242: Here you explicitly call `object_get_state`. I wonder if it would not be better to call `get_state` here. We don't know the type of the object that the method is bound to, it might need special treatment. Quickly looking at the existing special types we have, it seems unlikely that any method would be attached to them, but who knows if that might change in the future.

skops/io/tests/test_persist.py
```diff
@@ -804,3 +804,30 @@ def test_loads_from_str():
     msg = "Can't load skops format from string, pass bytes"
     with pytest.raises(TypeError, match=msg):
         loads("this is a string")
+
+
+class _BoundMethodHolder:
+    """Used to test the ability to serialise and deserialize bound methods"""
+
+    def __init__(self, variant: str):
```
BenjaminBossan, line 812: I appreciate that you want to provide a somewhat realistic example here. But I think overall it detracts from what the test is actually testing, since a lot of the code is not actually used. This makes it harder for future readers of the code to understand what happens. Thus I would prefer to reduce the example to the minimum necessary. Maybe a comment or two in the test could also help.

skops/io/tests/test_persist.py
```diff
@@ -804,3 +804,30 @@ def test_loads_from_str():
     msg = "Can't load skops format from string, pass bytes"
     with pytest.raises(TypeError, match=msg):
         loads("this is a string")
+
+
+class _BoundMethodHolder:
+    """Used to test the ability to serialise and deserialize bound methods"""
```
BenjaminBossan, line 810:
```suggestion
    """Used to test the ability to serialize and deserialize bound methods"""
```

Either British or American but not both ;) It seems we went with American in this repo.

skops/io/tests/test_persist.py
```diff
@@ -804,3 +804,30 @@ def test_loads_from_str():
...
+def test_for_serialized_bound_method_works_as_expected(tmp_path):
+    obj = _BoundMethodHolder(variant="log")
+    bound_function = obj.apply
+    transformer = FunctionTransformer(func=bound_function)
+
+    loaded_transformer = save_load_round(transformer, tmp_path / "file.skops")
+    loaded_bound_function = loaded_transformer.func.__self__.chosen_function
+
+    assert loaded_transformer.func.__name__ == bound_function.__name__
+    assert loaded_bound_function == obj.chosen_function
```
BenjaminBossan, line 833: Should we add `assert_params_equal(loaded_transformer.__dict__, transformer.__dict__)` for good measure?

BenjaminBossan (changes requested):
Overall this looks good, thanks a lot for investigating and proposing a solution. I made a few comments, please have a look.

And just to be clear, the initial issue reported in #184 is not solved yet, right? When I test it, I get an error (infinite recursion). Since we changed the skope (haha) of this PR, that's fine, I just want to make sure that this is expected.

skops/io/tests/test_persist.py
E-Aho: Good point, I can make this change and add some comments. 

I know most of the tests in this repo are just at the base level of the `test_...` file, would it be worth adding in a `TestBoundMethodSerialization` to go over edge cases, like the infinite loop caused by the original issue?

skops/io/tests/test_persist.py
E-Aho: I think this would make a lot of sense to add, especially if moving towards a more clear test as in your above comment

skops/io/_general.py
E-Aho: Hmm, maybe. This could also be involved in the infinite recursion problem...

I'll play with this now, and double check I understand the differences between what happens when using `get_state` vs `object_get_state`

skops/io/tests/test_persist.py
BenjaminBossan: Personally I don't mind a class containing multiple related tests, I've added a few of those myself. Since they were accepted, I assume the other maintainers don't mind either ;)

E-Aho:
> Overall this looks good, thanks a lot for investigating and proposing a solution. I made a few comments, please have a look.

Thanks for the comments! And it was a fun problem to investigate :)

 
> And just to be clear, the initial issue reported in #184 is not solved yet, right? When I test it, I get an error (infinite recursion). Since we changed the skope (haha) of this PR, that's fine, I just want to make sure that this is expected.

😨 

I definitely changed my focus from the original issue to a more general case, but I'm surprised it hasn't addressed the original issue, given it was working with just empty function calls for `method_get_state` and `method_get_instance`...

I'll look into this and see if there's a way to address the original bug here too. I don't love the idea of merging in a bug fix if it introduces a new bug, especially one like an infinite recursion 😨

skops/io/_general.py
BenjaminBossan: > I'll play with this now, and double check I understand the differences between what happens when using `get_state` vs `object_get_state`

The main difference that I is when the method is bound to a type that we have a special dispatch function for. Say, the method is bound to a numpy array (which is probably never going to happen in reality). With the current implementation, `object_get_state` is hard-coded, so `ndarray_get_state` is not called for the object, resulting in a failure.

E-Aho:
So after some digging, I think the problem with the original issue being an infinite loop is as follows:

1) We try to persist `zipf` as an object
2) We get the states of the attributes of `zipf` which includes `self.vecentropy`, which is just `numpy.vectorize(self._entropy)`
3) We try to persist the bound method, `self._entropy`, which depends on the state of the original zipf object, so try to persist state`zipf`
4) goto 1)

This results in the loop. 

I don't *think* this is a common pattern, so could add a special case for any `rv_discrete` objects, but I'm gonna try to see if I can think of a way to work around this for general cases of circular bound methods like this.

BenjaminBossan:
Great that you figured out the cause. I think for this PR, it's not necessary to solve the whole circular reference problem, that might be best left for another day or else the PR will become too big.

As for potential solutions, we might be able to use the memoization mechanism that we already added [here](https://github.com/skops-dev/skops/blob/e4f820aa5bac5ec96060df7628abdbb677c93910/skops/io/_utils.py#L246). There is an example of its usage for [numpy arrays](https://github.com/skops-dev/skops/blob/e4f820aa5bac5ec96060df7628abdbb677c93910/skops/io/_numpy.py#L31-L39) (which I now notice could be optimized...).

E-Aho:
Sounds good, I can raise an issue once this gets merged in to track the issue that's occuring with `zipf` right now 👍

skops/io/_general.py
```diff
@@ -232,6 +233,28 @@ def object_get_instance(state, src):
     return instance
 
 
+def method_get_state(obj: Any, save_state: SaveState):
+    # This method is used to persist methods bound to a different instance
```
BenjaminBossan, line 237: I know I insisted on the distinction of methods of _other_ instances but I now see that what I said is a bit misleading. A better distinction would be: methods that are stored as instance attributes. This can refer to methods on other objects (which is probably the most common case) but also methods stored on the same object (as with zipf). Could you please amend the comment to reflect that? Sorry for the inconvenience I caused.

skops/io/_utils.py
```diff
@@ -51,6 +51,7 @@ def dispatch(instance):  # CHANGED: variable name cls->instance
         # CHANGED: check if we deal with a state dict, in which case we use it
         # to resolve the correct class. Otherwise, just use the class of the
         # instance.
+
```
BenjaminBossan, line 54: Not necessary:

```suggestion
```

skops/io/tests/test_persist.py
```diff
@@ -804,3 +804,75 @@ def test_loads_from_str():
...
+class _BoundMethodHolder:
+    """Used to test the ability to serialize and deserialize bound methods"""
+
+    def __init__(self, object_state: str):
+        self.object_state = (
+            object_state  # Initialize with some state to make sure state is persisted
+        )
+        self.chosen_function = (
+            np.log  # bind some method to this object, could be any persistable function
+        )
```
BenjaminBossan, line 818:
```suggestion
        # Initialize with some state to make sure state is persisted
        self.object_state = object_state
        # bind some method to this object, could be any persistable function
        self.chosen_function = np.log
```

May be more readable, WDYT?

skops/io/tests/test_persist.py
```diff
@@ -804,3 +804,75 @@ def test_loads_from_str():
...
+        return self.chosen_function(x)
+
+
+class TestPersistingBoundMethods:
+    @staticmethod
+    def assert_transformer_persisted_correctly(
+        loaded_transformer: FunctionTransformer,
+        original_transformer: FunctionTransformer,
+    ):
+        """Checks that a persisted and original transformer are equivalent, including the func passed to it
```
BenjaminBossan, line 830: Could you please set a length limit inside the docstring (say, 79 or 88)? I know we haven't specified this anywhere, but I think it helps with readability and having a consistent look. Same applies to the method below.

skops/io/tests/test_persist.py
```diff
@@ -804,3 +804,75 @@ def test_loads_from_str():
...
+        bound_function = obj.bound_method
+        transformer = FunctionTransformer(func=bound_function)
+
+        loaded_transformer = save_load_round(transformer, tmp_path / "file.skops")
+        loaded_obj = loaded_transformer.func.__self__
+
+        self.assert_transformer_persisted_correctly(loaded_transformer, transformer)
+        self.assert_bound_method_holder_persisted_correctly(obj, loaded_obj)
+
+    def test_when_method_is_not_set_during_init_works_as_expected(self, tmp_path):
```
BenjaminBossan, line 863: Maybe I misunderstand, but the name of the test doesn't seem to be reflect what's actually happening. The comment below is correct though. Would you please find a more fitting name?

skops/io/tests/test_persist.py
```diff
@@ -804,3 +804,75 @@ def test_loads_from_str():
...
+
+        assert_params_equal(original_obj.__dict__, loaded_obj.__dict__)
+
+    def test_for_base_case_returns_as_expected(self, tmp_path):
+        initial_state = "This is an arbitrary state"
+        obj = _BoundMethodHolder(object_state=initial_state)
+        bound_function = obj.bound_method
+        transformer = FunctionTransformer(func=bound_function)
+
+        loaded_transformer = save_load_round(transformer, tmp_path / "file.skops")
```
BenjaminBossan, line 857: Just FYI, we added `dumps` and `loads` now, so you can also do `loaded_transformer = loads(dumps(transformer))`. There will be a refactor to change this throughout all tests (#198) so it's okay to keep it like this.

BenjaminBossan (changes requested):
Overall, this looks quite good to me, thanks for your work. I still have a few minor comments, please take a look.

> I can raise an issue once this gets merged in to track the issue that's occuring with zipf right now

Or we can just leave the original issue open.

skops/io/tests/test_persist.py
E-Aho: I 100% agree

I originally had these as just normal inline comments, but I think they got auto-linted by the pre-commit hook to these odd multiline bracket statements

skops/io/tests/test_persist.py
E-Aho: Oh nice! I might as well get ahead of it nonetheless 👍

skops/io/_general.py
E-Aho: No need to apologise! I think you've hit on an important distinction. Let me know if the new wording is clearer :)

E-Aho:
Huh, looks like given the changes in #200, neither FunctionType nor MethodType ever get returned by `gettype`.

BenjaminBossan:
> Huh, looks like given the changes in #200, neither FunctionType nor MethodType ever get returned by `gettype`.

Ah yes, this is not because `gettype` was changed in any way, this is because we no longer call `gettype` to determine kind of `get_instance` function to dispatch to.

E-Aho:
@BenjaminBossan if you have a minute could you rerun the CI?

BenjaminBossan:
> if you have a minute could you rerun the CI?

Sure, but feel free to ignore those "Error: Codecov: Failed to properly upload" messages. For some reason, those errors occur more frequently since the past week or so, but they're unrelated to anything actually being wrong.

E-Aho:
Oh for sure, I just wanted to check Codecov was actually happy now, the MethodType return not getting tested was making it upset

BenjaminBossan:
> I just wanted to check Codecov was actually happy now, the MethodType return not getting tested was making it upset

For sure, but you can take any of the passing CI runs to check that, no need to wait for each one to confirm.

E-Aho:
@BenjaminBossan, let me know if you have a moment to check the changes since your last review :)

skops/io/_general.py
```diff
@@ -242,6 +242,30 @@ def object_get_instance(state, src):
...
+            "func": obj.__func__.__name__,
+            "obj": get_state(obj.__self__, save_state),
+        },
+    }
+
+    return res
+
+
+def method_get_instance(state, src):
+    loaded_obj = object_get_instance(state["content"]["obj"], src)
```
BenjaminBossan, line 264:
```suggestion
    loaded_obj = get_instance(state["content"]["obj"], src)
```

Sorry, I missed this earlier, but I think this should be just plain `get_instance`, right? Analogous change to the one discussed `method_get_state` in an earlier review. Apart from that this LGTM.

I would like @adrinjalali to also take a look since this is very important to get right.

E-Aho: I think you’re right, happy to commit this as long as the other maintainers agree

skops/io/_general.py
```diff
@@ -242,6 +242,30 @@ def object_get_instance(state, src):
...
+    # dependent on a specific instance of an object.
+    # It stores the state of the object the method is bound to,
+    # and prepares both to be persisted.
+    res = {
+        "__class__": obj.__class__.__name__,
+        "__module__": get_module(obj),
+        "__loader__": "method_get_instance",
+        "content": {
+            "func": obj.__func__.__name__,
+            "obj": get_state(obj.__self__, save_state),
```
adrinjalali, line 256: I think here we have to put the object in the memo, not for efficiency or recursive reference issues, but for correctness:

```py
a = A()
f = a.f
g = a.g
c = C(f, g)
```

Dumping `c` will dump `a` twice, which means after load there will be two instances of `A()`. Now if `f` and `g` have side effects on `A`, in the above example, `c` is working on the same object, but after load it'd be working on two separate objects.

E-Aho: This was I think the root of the recursive calls occuring with the `zipf` object mentioned earlier.

I can get a version of this being memoized tonight 👍

BenjaminBossan: Good catch Adrin. Let's also add a test for this.

Just to note that the existing memoization doesn't really support checking if an object was already encountered and retrieving this memoized object, as it wasn't necessary so far. So the memoization mechanism would have to be extended to support this. But I was expecting it would become necessary sooner or later.

BenjaminBossan: I don't have any good ideas. First thing that came to mind was storing a "breadcrumb" to the actual object in the memo dict and then if in `get_state_*` we find that we already saved that object's state, instead of saving it again, we just save the breadcrumb.

One disadvantage that comes to mind would be that during loading, we would have to load the objects in the exact same order, otherwise the breadcrumb would lead nowhere. There might be other issues, I haven't investigated deeply.

What was your idea?

E-Aho: I've now implemented, so feel free to check out the new code in this branch, but essentially:

1) Store states of objects as normal in the `get_state`, but also store `__id__` of the original objects (this does mean objects that exist multiple times in the tree have their state stored multiple times, but I couldn't think of a tidy way to solve this without storing the state in multiple places)

2) Add a `LoadState` to get_instance calls to track `__id__` of objects as they get loaded

3) Check if an object with a given `__id__` has been instantiated already, if it has, return a pointer to that object instead of creating a new one.

I think storing in a JSON format makes it very hard to nicely store the state of an object that is referenced multiple times. Because like you said, we don't know which part of the tree will be loaded first, each time the object is referenced needs to be able to get the state. 

Perhaps a way to do this would be to do something along the lines of:

1) Store objects as normal, and memoize with `__id__`
2) If an object tries to get persisted a second times, remove state from original object, and leave `__id__` and `__is_memoized__` as a flag for each instance
3) At end of dump procedure, store each object that's been memoized like (2) in a `multi_instance_object` field outside normal tree, at top of JSON object
4) When loading, first check for any `multi_instance_object`s, and load them into the LoadState
5) Any object with the `_is_memoized_` flag just references the object in the LoadState

This would work I think, but it would be a pretty big shift from current save/load procedure...

E-Aho: But on the plus side, the MVP simple version is passing a new test checking the pattern @adrinjalali mentioned above, with only one instance of `A()` 🥳

E-Aho: I'll try to implement something that stores each object once done tomorrow, let me know if you think the procedure I described in steps 1-5 above sounds reasonable

BenjaminBossan: Wow, pretty big change. As always, big thanks @E-Aho for your excellent work.

I haven't done an in depth review, so just putting my thoughts out here.

1. I would have preferred this change to be its own PR.

I know that there is this potential bug that @adrinjalali mentioned, but maybe it's better to separate the memoization into its own PR and just add an `xfail`ing test for the bug? I think it's uncommon enough that we can survive with the main branch having this bug for a while.

2. A different signature for `get_instance_*` functions

Instead of adding a 3rd argument, it would be better if the `LoadState` holds the `src` as an additional attribute.

3. Deduplication is only done for method types

IIUC, the new memoization/dedup mechanism is only used for method types right now, is that right? I assume this is to keep the scope of the PR smaller, but if the whole mechanism is moved to a separate PR, I would like to see it applied for all types. That may mean that we should move it out of the individual `get_state_*` functions if possible.

4. Test too complex?

Maybe I'm missing something, but `test_works_when_given_multiple_bound_methods_attached_to_single_instance` seems to be doing more work than necessary to test for the bug. Instead of introducing `_MultipleMethodHolder`, would it not be sufficient to do something like this?

```python
foo = SomeObject()
estimator = FunctionTransformer(func=foo.bar, inverse_func=foo.baz)
```

Also, would this not be more correct?

```python
- assert loaded_obj.func1.__self__ == loaded_obj.func2.__self__
+ assert loaded_obj.func1.__self__ is loaded_obj.func2.__self__
```

5. Possible conflicts with Adrin's audit PR

@adrinjalali is working on auditing in #204. Since this PR introduces important changes to `get_instance`, there may be conflicts which may require coordination. Adrin can hopefully give more insights.

adrinjalali: I don't mind this work to move forward (memoization on load should be its own PR as @BenjaminBossan mentions though). I can handle the merge conflicts on my PR.

BenjaminBossan: > Perhaps a way to do this would be to do something along the lines of:
> ...
> 3. At end of dump procedure, store each object that's been memoized like (2) in a `multi_instance_object` field outside normal tree, at top of JSON object

Hmm. It feels wrong to me to have two classes of objects, those stored in the "tree" structure and those stored outside, in this new field. Yes, it would be nice to be able to avoid the dependency on the order in which we load objects, but maybe not at the cost of this added complexity.

E-Aho: @BenjaminBossan thanks for putting up with this PR through it's so far many iterations! Has been a fun problem to think through.

> 1) I would have preferred this change to be its own PR.

Totally agree that loading memoization should be a seperate PR, it feels quite far out of scope for "Add method type in dispatch calls". I originally kept it in this branch because it was addressing an issue with the new functionality, but I'll revert to before the latest change, and open a new issue to track loading states for single objects.

> 2) A different signature for get_instance_* functions

Definitely much tidier to keep `get_instance`, also keeps closer to the pattern used in `SaveState`.

> 3) Deduplication is only done for method types
You are right as to why I only addressed `method` deduping, i.e to fix the problem Adrin mentioned, but if we were to implement this functionality in general, I agree it should cover any type.

> 4) Test too complex?
That would also work as a test, yes. I introduced the `_MultipleBoundMethodHolder` because to me it was initially clearer what was being tested, but you've changed my mind and I would be happy to use a test closer to what you suggested.

E-Aho:
Ok, I've reverted this back to prior to the LoadState work.

I'll open an issue once this merges in and try to keep working on a `LoadState` that can load single instances across any object type :)

skops/io/tests/test_persist.py
```diff
@@ -792,6 +792,104 @@ def test_get_instance_unknown_type_error_msg():
...
+
+        transformer = FunctionTransformer(func=bound_function)
+
+        loaded_transformer = loads(dumps(transformer))
+        loaded_obj = loaded_transformer.func.__self__
+
+        self.assert_transformer_persisted_correctly(loaded_transformer, transformer)
+        self.assert_bound_method_holder_persisted_correctly(obj, loaded_obj)
+
+    @pytest.mark.xfail(reason="Can't load one obj referenced multiple times")
```
BenjaminBossan, line 870: The reason is not quite correct, is it? It can load the object, it just loads it multiple times.

BenjaminBossan (approved):
Great work, it's really appreciated. I think now the PR is very clean and readable. Thanks a lot.

I only have a minor comment about one of the `xfail`ing tests, which is not really a blocker, since I'm sure we can remove it soon :)

@adrinjalali please also take one more look.

adrinjalali (approved):
Other than @BenjaminBossan 's comment, LGTM.

skops/io/tests/test_persist.py
E-Aho: Yeah, I tried rewriting this a few times but it always seemed really clunky to say “can’t load an object as a single instance if referenced multiple times”, but it’s probably better to be clunky than vague here

E-Aho:
Awesome, thanks for all the support guys!

I've reworded the `xfail` reason to be a bit clearer, I'm happy to merge this in if you're happy @BenjaminBossan

BenjaminBossan:
@E-Aho Excellent, thanks, really great job.

Will you be working on the deduplication issue?

E-Aho:
> @E-Aho Excellent, thanks, really great job.
> 
> Will you be working on the deduplication issue?

🤗 Thanks for all the support, it really was a huge help.

And yes, I'll keep working on the dupe issue, I'll raise an issue later today once I've got time, and start working on implementing a proper *LoadState* based on the comments in this PR.

PR #196, opened 2022-10-18
CI Run different sklearn versions in CI

### LINKED ISSUE #192: Consider running CI on different sklearn versions
Follow up to this discussion: https://github.com/skops-dev/skops/pull/188#issuecomment-1280600073

We want to support multiple sklearn versions. Sklearn is an integral part of this library so we should ensure that skops works well with all supported versions. However, our CI only tests one version at the moment and it's unlikely that developers would test multiple versions locally. Therefore, it's easy to overlook errors caused by specific versions.

To help with this, CI should run on different sklearn versions. To not explode our test matrix, we may tie the sklearn version to the Python version, just as an example:

- Python 3.7 + sklearn 0.24
- Python 3.8 + sklearn 1.0
- Python 3.9 + sklearn 1.1
- Python 3.10 + sklearn main?

BenjaminBossan:
Resolves #192

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -10,12 +10,15 @@ concurrency:
 
 jobs:
   pytest:
+
     runs-on: ${{ matrix.os }}
     if: "github.repository == 'skops-dev/skops'"
     strategy:
       fail-fast: false  # need to see which ones fail
       matrix:
-        os: [ubuntu-latest]  # FIXME debug: , windows-latest, macos-latest]
+        os: [ubuntu-latest, windows-latest, macos-latest]
+        python: ["3.7", "3.8", "3.9", "3.10"]
+        # this is to make the CI run on different sklearn versions
         include:
           - python: "3.7"
             sklearn_version: "0.24"
@@ -41,7 +44,10 @@ jobs:
         pip install .[docs,tests]
         pip install black=="22.6.0" isort=="5.10.1" mypy=="0.981"
         pip uninstall scikit-learn
-        pip install "scikit-learn==${{ sklearn_version }}"
+        if [ ${{ matrix.sklearn_version }} == "nightly" ];
+          then pip install --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple scikit-learn;
+          else pip install "scikit-learn==${{ matrix.sklearn_version }}";
+        fi
         python --version
         pip --version
         pip list
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -21,11 +21,11 @@ jobs:
         # this is to make the CI run on different sklearn versions
         include:
           - python: "3.7"
-            sklearn_version: "0.24"
+            sklearn_version: "0.24.0"
           - python: "3.8"
-            sklearn_version: "1.0"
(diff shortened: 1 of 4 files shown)


adrinjalali:
the distutil warning we need to ignore. It's a ton of work on the sklearn side to migrate away from it, but that's not something we can do anything about.

The first one we can also ignore, but I'm not sure why it's only on MacOS. @ogrisel @thomasjpfan, anything we should care about here, or in sklearn?

ogrisel:
I don't understand either. Both use the same version of scipy 1.9.2 and this problem should have been fixed in scikit-learn dev in https://github.com/scikit-learn/scikit-learn/pull/23633.

Maybe the scikit-learn dev wheels are not the same on all the platforms? Since we do not include the git commit hash in the version name, it's hard to tell.

ogrisel:
Actually they have, it just that for macos we have older files such as `scikit_learn-1.2.dev0-cp310-cp310-macosx_10_13_x86_64.whl` that are favored by pip over the new and more backward compat files such as `scikit_learn-1.2.dev0-cp310-cp310-macosx_10_9_x86_64.whl`:

image


Let me delete the old macos files.

ogrisel:
Done. Feel free to push an empty commit to this PR to trigger the CI again and confirm that this fixed the problem.

BenjaminBossan:
Yes, it was fixed, big thanks @ogrisel

.github/workflows/build-test.yml
```diff
@@ -33,6 +43,11 @@ jobs:
       run: |
         pip install .[docs,tests]
         pip install black=="22.6.0" isort=="5.10.1" mypy=="0.981"
+        pip uninstall --yes scikit-learn
+        if [ ${{ matrix.sklearn_version }} == "nightly" ];
+          then pip install --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple scikit-learn;
+          else pip install "scikit-learn>=${{ matrix.sklearn_version }}";
```
adrinjalali, line 49: shouldn't this be

```suggestion
          else pip install "scikit-learn=${{ matrix.sklearn_version }}";
```

?

BenjaminBossan:
@adrinjalali I added the skip for the inference tests when using a dev sklearn version

.github/workflows/build-test.yml
adrinjalali: Right, then maybe this is better?

```shell
$ pip install scikit-learn~=1.0.0
Collecting scikit-learn~=1.0.0
  Downloading scikit_learn-1.0.2-cp310-cp310-manylinux_2_17_x86_64.manylinux2014_x86_64.whl (26.5 MB)
```

https://peps.python.org/pep-0440/#compatible-release

adrinjalali (commented):
Awesome, seems like it was easier than we thought!

BenjaminBossan:
@adrinjalali TIL about `~=`. I made the change and it looks good except for `scikit-learn~=1.0`, which results in:

```
Collecting scikit-learn~=1.0
  Using cached scikit_learn-1.1.2-cp38-cp38-win_amd64.whl (7.3 MB)
```

([here](https://github.com/skops-dev/skops/actions/runs/3287955478/jobs/5417716311) e.g.)

Any idea?

PR #199, opened 2022-10-21
MNT Rename skops.io.save to skops.io.dump

BenjaminBossan:
As discussed in #182.

Also added a small snippet to docs about using `dumps` and `loads`.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -30,27 +30,32 @@ the `Python docs
     code during unpickling. Never unpickle data that could have come from an
     untrusted source, or that could have been tampered with.
 
-In contrast to ``pickle``, the :func:`skops.io.save` and :func:`skops.io.load`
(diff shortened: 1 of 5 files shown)


BenjaminBossan:

I grep'ed through the repo, so hopefully I didn't miss any uses of `save`. However, there are probably examples outside this repo that use `save`, maybe @merveenoyan could confirm? Those would break right now. We could leave a vestige `save = dump` but I'd rather only have one way before the release. WDYT?

(re codecov, maybe I'm using it wrong but I just don't see where the -0.01% come from)

adrinjalali (approved):
The codecov trigger is because we're removing a previously covered line here, therefore the coverage percentage goes down, I think

LGTM.

BenjaminBossan:
> The codecov trigger is because we're removing a previously covered line here, therefore the coverage percentage goes down, I think

Ah okay, now I get it, still debatable if this should count as a failure.

PR #200, opened 2022-10-21
Refactor: get_instance method saved in state

### LINKED ISSUE #197: Refactor get_instance functions to not use dispatch
This is more of a reminder for myself after discussion with Adrin.

Currently, during the dispatch of the `get_instance` functions, the class stored in the `state` is being loaded to determine which function to dispatch to. This is bad because module loading can be dangerous. We will add auditing but this is intended to be on the level of `get_instance` itself, not for the dispatch mechanism.

The solution proposed by Adrin is to store which `get_instance` method should be used as part of the state, therefore allowing us to avoid having to figure that out via the dispatcher. As a side effect, this change should (hopefully) allow us to avoid having a custom `singledispatch` implementation.

BenjaminBossan:
Resolves #197

## Description

Currently, during the dispatch of the `get_instance` functions, the class stored in the state is being loaded to determine which function to dispatch to. This is bad because module loading can be dangerous. We will add auditing but it is intended to be on the level of `get_instance` itself, not for the dispatch mechanism.

In this PR, the state returned by `get_state` functions is augmented with the name of the get_instance method required to load the object. This way, we can look up the correct method based on the state and don't need to use the modified `singledispatch` mechanism, thus avoiding loading modules during dispatching.

## Implementation

Whereas for `get_state`, we still rely in `singledispatch`, for `get_instance` we now use a simple dictionary that looks up the function based on its name (which in turn is stored in the state). The dictionary, going by the name of `GET_INSTANCE_MAPPING`, is populated similarly to how the `get_instance` functions were registered previously with `singledispatch`.

There was an issue with circular imports (e.g. `get_instance > GET_INSTANCE_MAPPING > ndarray_get_instance > get_instance`), hence the `get_instance` function was moved to its own module, `_dispatch.py` (better name suggestions are welcome).

For some types, we now need extra get_state functions because they have specific `get_instance` methods. So e.g. `sgd_loss_get_state` just wraps `reduce_get_state` and adds `sgd_loss_get_instance` as its loader.

## Coincidental changes

Since we no longer have to inspect the contents of state to determine the function to dispatch to for `get_instance`, we can fall back to the Python implementation of `singledispatch` instead of rolling our own. This side effect is a big win.

The function `Tree_get_instance` was renamed to `tree_get_instance` for consistency.

In the `debug_dispatch_functions`, there was some code from a time when the state was allowed not to be a dict (json-serializable objects). Now we always have a dict, so this dead code was removed.

Also, this fixture was elevated to module-level scope, since it only needs to run once.

### DIFF
--- a/skops/io/_dispatch.py
+++ b/skops/io/_dispatch.py
@@ -1,5 +1,7 @@
 #!/usr/bin/env python3
 
+from __future__ import annotations
+
 import json
 from typing import Any, Callable
 from zipfile import ZipFile
--- a/skops/io/_dispatch.py
+++ b/skops/io/_dispatch.py
@@ -1,5 +1,3 @@
-#!/usr/bin/env python3
-
 from __future__ import annotations
 
 import json
--- a/skops/io/_dispatch.py
+++ b/skops/io/_dispatch.py
@@ -3,10 +3,8 @@
 from __future__ import annotations
 
 import json
-from typing import Any, Callable
-from zipfile import ZipFile
 
-GET_INSTANCE_MAPPING: dict[str, Callable[[dict[str, Any], ZipFile], Any]] = {}
+GET_INSTANCE_MAPPING = {}  # type: ignore
 
 
 def get_instance(state, src):
@@ -17,7 +15,8 @@ def get_instance(state, src):
     try:
         get_instance_func = GET_INSTANCE_MAPPING[state["__loader__"]]
     except KeyError:
+        type_name = f"{state['__module__']}.{state['__class__']}"
         raise TypeError(
-            f"Creating an instance of type {type(state)} is not supported yet"
+            f" Can't find loader {state['__loader__']} for type {type_name}."
         )
     return get_instance_func(state, src)
--- a/skops/io/_dispatch.py
+++ b/skops/io/_dispatch.py
@@ -1,0 +1,21 @@
+#!/usr/bin/env python3
+
+import json
+from typing import Any, Callable
+from zipfile import ZipFile
+
+GET_INSTANCE_MAPPING: dict[str, Callable[[dict[str, Any], ZipFile], Any]] = {}
+
+
+def get_instance(state, src):
+    """Create instance based on the state, using json if possible"""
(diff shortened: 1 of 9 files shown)


BenjaminBossan:

BenjaminBossan:
Addendum: I noticed (thanks codecov) that I had accidentally not registered `bunch_get_state`. Therefore, `bunch_get_instance` also was not used. Still, the tests passed. At first, I thought we might not be testing correctly but `Bunch` attributes are indeed loaded as `Bunch` objects. Therefore, it seems we can remove the special treatment for `Bunch`.

I'm not exactly sure which change we did that resulted in this, since I'm pretty sure it used to be necessary, but it's not anymore. Therefore, I removed the corresponding code.

skops/io/_dispatch.py
```diff
@@ -0,0 +1,23 @@
...
+def get_instance(state, src):
+    """Create instance based on the state, using json if possible"""
+    if state.get("is_json"):
+        return json.loads(state["content"])
+
+    try:
+        get_instance_func = GET_INSTANCE_MAPPING[state["__loader__"]]
+    except KeyError:
+        raise TypeError(
+            f"Creating an instance of type {type(state)} is not supported yet"
```
adrinjalali, line 21: type of state is a dict. We should say something like:

```
f" Can't find loader {state["__loader__"]} for type {state[__module__]}.{state["__class__"]}."

File: skops/io/_sklearn.py [hunk 2]
```diff
@@ -21,10 +21,10 @@
     SquaredLoss,
 )
 from sklearn.tree._tree import Tree
-from sklearn.utils import Bunch
```
adrinjalali, line 24: this is because we load and create the class of the object whatever it was now in `dict_get_instance` instead of creating a `dict`.

adrinjalali (commented):
This looks really nice!

BenjaminBossan:
@adrinjalali I addressed your comments, also added a test for the error message.

adrinjalali (approved):
I'm mostly happy about the code we're removing here lol

PR #202, opened 2022-10-24
Refactor: clean persistence tests to avoid duplicate testing

### LINKED ISSUE #198: Clean up persistence tests to make them run faster
Just dumping some quotes since I'm lazy:

> BB: Regarding the tests, I have modified them so that they now run once with save/load and once with dumps/loads. This of course doubles the time required to run those tests, so in the future we may only run one setting (dumps and loads) and then have a handful of tests that specifically test save/load. [#182]

> AJ: Yeah I saw the code and thought the same. I think testing for a few cases that dump and load do the same as dumps and loads should be enough, and the rest we just do in memory.

BenjaminBossan:
Resolves #198 

After introducing `dumps` and `loads`, we ran all persistence tests twice, once with those functions and once with `dump` and `load`. This is inefficient, since they're basically testing the same thing but running 400+ tests twice is costly. With this PR, we only use the more efficient `dumps` and `loads`. On top of that, there is one test that very specifically tests `dump`  and `load`.

### DIFF
--- a/skops/io/tests/test_persist.py
+++ b/skops/io/tests/test_persist.py
@@ -6,6 +6,7 @@ import sys
 import warnings
 from collections import Counter
 from functools import partial, wraps
+from pathlib import Path
 from zipfile import ZipFile
 
 import numpy as np
@@ -51,7 +52,7 @@ from sklearn.utils.estimator_checks import (
 )
 
 import skops
-from skops.io import dumps, loads
+from skops.io import dump, dumps, load, loads
 from skops.io._dispatch import GET_INSTANCE_MAPPING, get_instance
 from skops.io._sklearn import UNSUPPORTED_TYPES
 from skops.io._utils import _get_state, get_state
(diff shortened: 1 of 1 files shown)


BenjaminBossan:

skops/io/tests/test_persist.py
```diff
@@ -807,3 +789,44 @@ def test_get_instance_unknown_type_error_msg():
...
+    """Estimator with np array, np scalar, and sparse matrix attribute"""
+
+    def fit(self, X, y=None):
+        self.numpy_array = np.zeros(3)
+        self.numpy_scalar = np.ones(1)[0]
+        self.sparse_matrix = sparse.csr_matrix(np.arange(3))
+        return self
+
+
+def test_dump_to_and_load_from_disk(tmp_path):
```
adrinjalali, line 804: maybe as an alternative, or another test, we could save/load from disk and memory, and compare the objects' hashes to make sure they're the same?

adrinjalali (commented):
Thanks, otherwise LGTM.

skops/io/tests/test_persist.py
BenjaminBossan: Yes, I can add another test. To be sure I understand correctly, you mean comparing the hashes of both objects in memory, one obtained through `dump/load` and one for `dumps/loads`? If so, what method would you use to hash them, `joblib.hash`?

adrinjalali: yes, exactly.

BenjaminBossan: done

adrinjalali (approved):

Issue #72, opened 2022-08-03
Programmatically add sections to model card


merveenoyan:
Currently `add` only supports injecting values to already existing sections in a template, meanwhile add_plot can append new sections at the end of the model card and write plots. `add` should also support adding sections and subsections to the model card, without appending if possible. e.g. for default template, if user wants to add a subsection to training procedure (which is in the middle of the card) they should be able to do that without modifying the template themselves.

adrinjalali:
We could also document how users can do this by adding markdown markers. As in, if the user adds this as the content of a section:

```
this is the main content.

## Subsection

this is the subsection content.
```

it will be rendered as a subsection.

rushic24:
Can anyone assign this to me

BenjaminBossan:
@rushic24 Before working on this, let's first discuss how to actually implement this feature, as it's not trivial. If you already have an idea, please describe it here.

E-Aho:
Is it neccesary for `add` to take multiple sections at a time? 

If not, we could refactor the method to take in a single section at a time, and add an optional "parent" parameter, that would allow a workflow like:

1) If section in template, add as normal
2) If parent is none, add at end 
3) If parent, add under parent

This would need a bit of a refactor for the render method and the way `self._template_sections` stores sections, but should be intuitive for the user

BenjaminBossan:
> Is it neccesary for `add` to take multiple sections at a time?

I wouldn't change it, honestly, as I believe it's quite useful as is.

Perhaps there is a way, however, to keep it and still add the feature. OTOH:

```python
card = Card(...)
card.add({
    "Model description/Training Procedure/New section": "some text",
})
```

where "Model description" and "Training Procedure" are existing sections and "New section" is added as a subsection. This still leaves a few open questions, but it's maybe worth exploring.

Going from API to implementation, this is the bigger concern for me though. I don't think the current way we handle model cards lends itself well for this type of dynamic additions. Trying to insert sections with regex looks too brittle to me. I couldn't yet come up with a good solution that works with the current model card implementation.

rushic24:
Instead of string we can enable markdown rendering for card values
like

```
card = Card(...)
card.add({
    "Model description/Training Procedure/New section": """
    
    # Test
    sdjksdksdjsjks
    ## Subsection
    """,
})
```

BenjaminBossan:
> Instead of string we can enable markdown rendering for card values

Maybe I don't quite understand your suggestion, but your example should already work right now.

[...]

PR #203, opened 2022-10-25
Alternative Card implementation

BenjaminBossan:
This is an RFC for a suggested solution for the issues discussed on #72

## Description

The proposed model card implementation would allow to dynamically add sections or overwrite them.

This is not a complete implementation but already covers most of the features we already have and then some.

On top of these features, it would be possible to add more features like creating a default `Card` with placeholders, just like the existing template, or the possibility to delete existing sections or to retrieve the result of a certain section.

## Implementation

The underlying data structure consists of a dict and a `Section` `dataclass`.

All data is stored in a `_data` attribute with the type `dict[str, Section]`. The `dataclass` hold the section contents, i.e. the section title, the section content, and subsections, which again have the same type. It's thus recursive data structure (basically a tree). Section title and dict key are identical, which is mostly for convenience.

With this refactor, there are no separate data containers anymore for eval results, template sections, extra sections, etc. They are all treated the same.

IMHO, this greatly simplifies the code overall. The only complex function that's left is the one needed to traverse the tree holding the data, and even that is just 14 LOC.

## Demo

**updated**

To see how the new class can be used, take a look at the main function. The resulting Card can be seen here:

https://huggingface.co/skops-ci/hf_hub_example-e02eee42-5c99-4924-879e-95a20d795d49

The README file looks like this:

https://huggingface.co/skops-ci/hf_hub_example-e02eee42-5c99-4924-879e-95a20d795d49/raw/main/README.md


BenjaminBossan:
RFC @skops-dev/maintainers @E-Aho 

Don't hesitate to tell me if you think this totally sucks ;)

skops/card/_card_alternative.py
```diff
@@ -0,0 +1,361 @@
...
+        # add a placeholder for figures
+        card.add(Plots="")
+
+        # add arbitrary sections, overwrite them, etc.
+        card.add(hi="howdy")
+        card.add(**{"parent section/child section": "child content"})
+        card.add(**{"foo": "bar", "spam": "eggs"})
+        # change content of "hi" section
+        card.add(**{"hi/german": "guten tag", "hi/french": "salut"})
+        card.add(**{"very/deeply/nested/section": "but why?"})
```
E-Aho, line 317: Love this usage pattern ❤️ feels much easier to add to programatically

skops/card/_card_alternative.py
```diff
@@ -0,0 +1,361 @@
...
+        raise NotImplementedError
+
+    def __init__(
+        self, model, model_diagram: bool = True, metadata: CardData | None = None
+    ):
+        self.model = model
+        self.model_diagram = model_diagram
+        self.metadata = metadata or CardData()
+
+        self._data: dict[str, Section] = {}
```
E-Aho, line 64: One thought: Might be worth using an `OrderedDict` for this (and for the `subsections` field in the `Section` class.)

That way it would allow us to have more control over the ordering of the output. I know that as of 3.7, the order of the keys in a `dict` is just the order of insertion, but it might be nice to offer users the ability to insert between existing sections.

skops/card/_card_alternative.py
```diff
@@ -0,0 +1,361 @@
...
+            section = TableSection(table=val, folded=folded)
+            self._add_single(key, section)
+        return self
+
+    def add_metrics(self, **kwargs: int | float) -> "Card":
+        self._metrics.update(kwargs)
+        self._add_metrics(self._metrics)
+        return self
+
+    def _add_metrics(self, metrics: dict[str, float | int]) -> None:
```
E-Aho, line 154: One thought I had (feel free to challenge me or tell me it's a bad idea):

If we have a lot of these helper `_add_...` sections, it might be nice to split this class into two classes, one with the core functionality methods (e.g `_strip_blank` or __repr__`) and a subclass with the default formatting/helper functions, which a user might want to tweak themselves.

That way, it would be a lot easier for a user to make their *own* `MyCard` class that has different formatting/ extra methods for their own models, similar to how they might create their own template in the old .md world.

E-Aho (commented):
Overall I think this is great! In my opinion it looks like it would be a lot more flexible than using .md templates.

I had a couple things I thought might be nice additions, but overall I really like this implementation :)

skops/card/_card_alternative.py
adrinjalali: Do we need to support that? I see it as "the card is rendered in the order you're adding things" kinda API, especially since we don't expose this `_data` to the users as a public API.

skops/card/_card_alternative.py
```diff
@@ -0,0 +1,361 @@
...
+
+
+class Formattable(Protocol):
+    def format(self) -> str:
+        ...
+
+
+class Card:
+    @classmethod
+    def make_default(
```
adrinjalali, line 51: when is this called? In the constructor? We should have something which adds all the missing recommended sections by default.

skops/card/_card_alternative.py
```diff
@@ -0,0 +1,361 @@
...
+    ):
+        self.model = model
+        self.model_diagram = model_diagram
+        self.metadata = metadata or CardData()
+
+        self._data: dict[str, Section] = {}
+        self._metrics: dict[str, float | int] = {}
+        self._reset()
+
+    def _reset(self) -> None:
```
adrinjalali, line 68: from the code I don't see a _reset_ happening here. As in, nothing seems to be cleared.

skops/card/_card_alternative.py
BenjaminBossan: Yes, one weakness of the chosen data structure is that it's hard to change the order of sections once they are created. However, I'm not sure how much `OrderedDict` would help. I haven't used it in quite a while, but from the docs, the only additional methods it seems to have are `move_to_end` and `popitem(last=...)`, both of which don't help that much? Ideally, we would have something like:

```python
card = Card(...)
card.add(section_a="content a", section_c="content c")
card.insert(section_b="content b", after="section_a")
```

or something. Not sure if that's easier to implement with `OrderedDict` than with normal dicts.

Alternatively, we could implement the underlying tree as a nested list, which gives us more flexibility to modify the order:

```
>>> card._data
[
  ("section_a", "content a"),
  ("section_b", [
    "content b",
    ("section_b0", "content_b0"),
    ("section_b1", "content_b1"),
  ],
  ...,
]
```

The disadvantage is that selecting items will be more difficult an inefficient.

skops/card/_card_alternative.py
adrinjalali: I think instead of that they can pass something `Formatable` to `add`, and that's much easier to implement for them, isn't it?

adrinjalali (commented):
I quite like the idea. I wonder how much work it'd be to load an `.md` file back to this object.

skops/card/_card_alternative.py
BenjaminBossan: > That way, it would be a lot easier for a user to make their own MyCard class that has different formatting/ extra methods for their own models, similar to how they might create their own template in the old .md world.

Let me try to understand this better. Right now, if a user would like to make small modifications to the existing templates, I think this can be easily accommodated with the added dynamism of the new class. If they want to have a new default template that they can share with others, it's not easily possible, they would have to share the code (a script) that generates this new template.

What you suggest is that it should be possible to create a subclass of `Card` that starts with the new default. I like the idea. If that requires to split the class, I'm not sure, since pre-filling the card is not implemented yet (this is what I hoped to achieve with the `make_default` class method). If splitting makes it easier for the user, then I'm all for it, I would have to see it in action.

Another issue is if we want to enable completely different output formats. Just as an example, what if a user wants to create an `rst` or `org` file instead of `md`, or another `md` flavor? Right now, the `md` format is hard-coded. If we could split the class so that the "formatting engine" is decoupled, it would be a nice gain. But maybe we'll just never support other formats and users have to use `pandoc` or similar tools to achieve this.

skops/card/_card_alternative.py
BenjaminBossan: The idea is that if we want to have a model card with the defaults pre-filled as they currently are, we would call `card = Card.make_default(...)`. Of course, that would be bc breaking if `__init__` doesn't pre-fill. It's more intuitive IMO for `__init__` to create an empty card and for pre-filled ones to use class methods, but we can also change that.

skops/card/_card_alternative.py
BenjaminBossan: Ah yes, it's not used yet. The idea is that if I override, say, the model, we can internally call `reset` and more sure that everything that relates to the change is updated.

skops/card/_card_alternative.py
E-Aho: > Do we need to support that? I see it as "the card is rendered in the order you're adding things" kinda API, especially since we don't expose this _data to the users as a public API.

I think it would be fine to not support changing ordering for an MVP, but it would be at least a nice-to-have down the road imo. Especially if the default `Card` class users handle has a fixed order for the defaults. Users might want to insert things between other sections.

E-Aho: > The disadvantage is that selecting items will be more difficult an inefficient.

I may have been remembering the abilities of OrderedDicts with rose tinted glasses... 
Another alternative that might be a little simpler would be to hold a `subsection_order` attribute in each `Section` that is just a list with the names of children, and renders them in that order.

So if we rendered, as above:
```
("section_b", [
    "content b",
    ("section_b0", "content_b0"),
    ("section_b1", "content_b1"),
  ],
)
```
Section b would just have an attribute like `[__content__, section_b0, section_b1]`

We could implement ease-of-use things like support for `.add(..., after="foo")` as above as well

On efficiency, while it's always good to think about, I don't think we need to get too deep into optimizations with something like this. I'd be very surprised if anyone had even 30 sections total, and at that kind of scale it's not really a limitation imo.

adrinjalali: > The disadvantage is that selecting items will be more difficult an inefficient.

We can always implement `__getitem__`.

BenjaminBossan: > Another alternative that might be a little simpler would be to hold a subsection_order attribute in each Section that is just a list with the names of children, and renders them in that order.

This would be a kind of index. I think if possible, we should avoid it so that we don't leave room for possible bugs where the content and the "index" are out of sync.

> On efficiency, while it's always good to think about, I don't think we need to get too deep into optimizations with something like this.

Probably you're right, the main concern should be usability and simple code. But you never know how users can surprise you. In skorch, we ran into an issue because a user was training a neural net for tens of thousands of epochs, which we didn't expect...

> We can always implement `__getitem__`.

Yes, I thought about that, it just feels "wrong" to have that method when the access is O(n) and not O(1).

skops/card/_card_alternative.py
merveenoyan: is this going to simply erase README.md as it's still what we use to version models and create a new one? just to make sure we're on same page.

skops/card/_card_alternative.py
merveenoyan: I feel like nested list is a bit overkill (and selection should be less efficient as you said). I'm in favor of implementation as is with a regular dict.

merveenoyan (commented):
For very basic details I'm in favor of keeping this PR as is and not changing any data structures as it looks quite simple. It was quite overwhelming to catch up with this and review though, I will never take time off from work or go to far east again 😂😂

skops/card/_card_alternative.py
BenjaminBossan: No, this will not erase the `README.md`. The intent was just that if a user re-assigns the `model`, in the setter, we can call `self._reset()` and every part of the model card that needs updating will be updated. I will do the proper implementation once #205 is merged, I hope it becomes more obvious then.

merveenoyan: I feel like it makes sense to update the parts that weren't previously written by the user (hyperparameters, plot and so on) it was a pain point with `modelcards` as well.

adrinjalali (commented):
I didn't dig deep into everything here, but overall it looks great. Genrally:

- I think quite a few of the private methods could use a docstring to make it more clear what they do.
- The tests are changed quite a bit. I can't tell which ones are improvements to existing tests and which ones are due to breaking changes. I don't mind breaking changes when it comes to the rendered doc as a result of this PR, but it'd be nice to have a rather small-ish change in the tests compared to what we have now.
- It's not clear to me what the user can do when they `select` a section, and why they'd do that.

skops/card/_model_card.py
```diff
@@ -162,13 +189,106 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
     return card_data
 
 
+def split_subsection_names(key: str) -> list[str]:
```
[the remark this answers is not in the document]
BenjaminBossan: Yes, I think it should be possible somehow, probably involving non-capturing groups (`(?:...)`) where we somehow discard `"\\/"` and only keep `"/"`. But I couldn't get to work with split after some trying. I feel like the regex solution would be more complicated in the end.

skops/card/_model_card.py
```diff
@@ -215,19 +331,22 @@ class Card:
...
+    >>> model_card = Card(model)
     >>> model_card.metadata.license = "mit"
     >>> y_pred = model.predict(X)
     >>> model_card.add_metrics(**{
     ...     "accuracy": accuracy_score(y, y_pred),
     ...     "f1 score": f1_score(y, y_pred, average="micro"),
     ... })
     Card(
-      model=LogisticRegression(random_state=0, solver='liblinear'),
+      model=LogisticRegression(random_state=0, solver='liblinear')
```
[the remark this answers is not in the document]
BenjaminBossan: yeah, no good reason :)

skops/card/_model_card.py
```diff
@@ -239,49 +358,295 @@ class Card:
     >>> tmp_path = Path(tempfile.mkdtemp(prefix="skops-"))
     >>> disp.figure_.savefig(tmp_path / "confusion_matrix.png")
     ...
-    >>> model_card.add_plot(confusion_matrix="confusion_matrix.png")
+    >>> model_card.add_plot(**{
+    ...     "Model description/Confusion Matrix": tmp_path / "confusion_matrix.png"
+    ... })
+    Card(
```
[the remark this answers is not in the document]
BenjaminBossan: right, it's kinda noisy

skops/card/_model_card.py
```diff
@@ -239,49 +358,295 @@ class Card:
...
-        for section, value in kwargs.items():
-            self._template_sections[section] = value
+        for key, val in kwargs.items():
+            self._add_single(key, val)
         return self
 
-    def add_plot(self, folded=False, **kwargs: str) -> "Card":
+    def _select(
+        self, subsection_names: Sequence[str], create: bool = True
+    ) -> dict[str, Section]:
```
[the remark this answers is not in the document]
BenjaminBossan: :+1:

skops/card/_model_card.py
```diff
@@ -239,49 +358,295 @@ class Card:
...
+            *subsection_names, leaf_node_name = key
+
+        if not leaf_node_name:
+            msg = f"Section name cannot be empty but got '{key}'"
+            raise KeyError(msg)
+
+        parent_section = self._select(subsection_names, create=False)
+        return parent_section[leaf_node_name]
+
+    def delete(self, key: str | Sequence[str]) -> None:
```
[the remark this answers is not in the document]
BenjaminBossan: See response elsewhere.

BenjaminBossan (commented):
> I think quite a few of the private methods could use a docstring to make it more clear what they do.

Yeah, I focused on public methods, will add docstrings to the private ones too.

> The tests are changed quite a bit. I can't tell which ones are improvements to existing tests and which ones are due to breaking changes. I don't mind breaking changes when it comes to the rendered doc as a result of this PR, but it'd be nice to have a rather small-ish change in the tests compared to what we have now.

None of the changes are breaking, except that we don't automatically put some stuff in specific sections anymore, instead giving users the control over the sections. In a previous version of this PR, I even had a test that showed that old and new implementation provide identical results:

https://github.com/BenjaminBossan/skops/blob/e7a2ae3e0e56738423962df53dd4535a11da33ea/skops/card/tests/test_card_alternative.py#L828-L958

However, as we decided to replace the old implementation, the test also had to go.

Regarding how the tests are changed, they test for the most part exactly what has been tested before, but they are more strict. Let's illustrate with an example. Below is the old test for adding a plot:

```python
def test_add_plot(destination_path, model_card):
    plt.plot([4, 5, 6, 7])
    plt.savefig(Path(destination_path) / "fig1.png")
    model_card = model_card.add_plot(fig1="fig1.png").render()
    assert "![fig1](fig1.png)" in model_card
```

Here is the new test:

```python
def test_add_plot(destination_path, model_card):
    plt.plot([4, 5, 6, 7])
    plt.savefig(Path(destination_path) / "fig1.png")
    model_card = model_card.add_plot(fig1="fig1.png")
    plot_content = model_card.select("fig1").content.format()
    assert plot_content == "![fig1](fig1.png)"
```

As you can see, they are fairly similar. However, the old test would succeed if _anywhere_ in the model card, there is a string `"fig1"`. The new test specifically tests that this string is in the section we indicated. A similar pattern can be observed with the other tests.

I know that the diff view does not really help in identifying these changes, maybe it's better to put old and new tests side-by-side and going from top to bottom.

> It's not clear to me what the user can do when they select a section, and why they'd do that.

The main reason I added `select` and `delete` was to be complete, so if a user wants to change or delete a section, they get the possibility (allowing all CRUD operations basically).

E.g. select allows you to do stuff like:

```python
card = Card(...)
card.add_plot(**{"Plots/My awesome plot": "myfig.png"})
pl = card.select("Plots/My awesome plot")
pl.title = "My new title"
```

and it will be reflected in the model card. Without select, you'd need to recreate the whole section. In effect, this means that building the model card as a whole can be a lot more interactive, leaving room for mistakes and fixing them.

Another benefit of `select` is that we can use it for testing more precisely (see comment above).

I will address some of your minor comments and then provide a new update once the permutation importances and model loading from str are merged, to integrate those features as well.

examples/plot_model_card.py
```diff
@@ -160,8 +165,8 @@
 model_card.add_table(
     folded=True,
     **{
-        "Hyperparameter search results": cv_results,
-        "Classification report": clf_report,
+        "Model description/Evaluation Results/Hyperparameter search results": cv_results,
+        "Model description/Evaluation Results/Classification report": clf_report,
```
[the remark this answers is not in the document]
BenjaminBossan: Okay, so it seems this template has a different structure to what is used in the [skops template](https://raw.githubusercontent.com/skops-dev/skops/main/skops/card/default_template.md).

E.g. metrics go into

`# Evaluation > ## Testing Data, Factors & Metrics > ### Metrics`

in the hub template, and into

`# Model description > ## Evaluation Results`

in the skops template. I'm impartial as to how to structure the default template, in this implementation I tried to  keep it the same as the previous skops model card.

skops/card/_model_card.py
```diff
@@ -1,27 +1,54 @@
...
-from huggingface_hub import CardData, ModelCard
+from huggingface_hub import CardData
 from sklearn.utils import estimator_html_repr
 from tabulate import tabulate  # type: ignore
 
-import skops
+if sys.version_info >= (3, 8):
+    from typing import Protocol
+else:
+    from typing_extensions import Protocol
```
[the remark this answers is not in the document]
BenjaminBossan: The problem when putting this into `fixes.py` and importing it is that for some reason, mypy doesn't understand the protocol anymore. E.g. it'll claim:

> skops/card/_model_card.py:694: error: Argument 2 to "_add_single" of "Card" has incompatible type "PlotSection"; expected "Union[Formattable, str]"

even though `PlotSection` conforms to the `Formattable` protocol. Now we could put a `# type: ignore` on this method, but then we might as well not use a protocol at all, if it is never checked.

Yes, it is annoying that this doesn't work. However, in this case I'd be in favor of keeping it as is because it's a temporary fix anyway. As soon as we can stop supporting Python 3.7, all these gymnastics can go away (official EOL for Python 3.7 is 27 June 2023).

examples/plot_model_card.py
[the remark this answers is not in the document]
adrinjalali: If we're changing it, I'd make it the same as the template we have in `huggingface_hub`

skops/card/_model_card.py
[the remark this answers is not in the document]
adrinjalali: ok, then we can add a comment here and in `fixes.py` to remember it.

examples/plot_model_card.py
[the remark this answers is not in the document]
BenjaminBossan: Happy to change it to conform to the hub card. Since the skops default template was already there before I joined, I just assumed it was consciously changed to better fit the needs for sklearn models

skops/card/_templates.py
```diff
@@ -0,0 +1,144 @@
...
+        "You can contact the model card authors through following channels:\n"
+        f"{CONTENT_PLACEHOLDER}"
+    ),
+    "Citation": (
+        "Below you can find information related to citation.\n\n**BibTeX:**\n```\n"
+        f"{CONTENT_PLACEHOLDER}\n```"
+    ),
+}
+
+HUB_TEMPLATE = {
```
adrinjalali, line 36: spending more time on this, I don't see an easy way to make it suitable for usual tabular usecases. For instance, the environmental impact is pretty much zero for most models.

However, one can do NLP/image usecases and have massive models as a part of a scikit-learn pipeline. We can start with a minimal model card template and improve it then.

BenjaminBossan: So what is your suggestion here? To trim the `HUB_TEMPLATE` down to only contain those sections that make most sense for sklearn models? Or create a new template?

adrinjalali: my suggestion is to go back to support only one template, the one we already have, and change it/improve it/add more templates in separate PRs.

adrinjalali: That all sounds good to me. Supporting custom templates via dict also makes sense.

adrinjalali (commented):
I didn't check the tests in detail, but they seem pretty good.

To me this seems pretty close to being mergeable.

skops/card/_model_card.py
```diff
@@ -232,6 +358,17 @@ class Card:
...
+        Whether to add default sections or not. The template can be a predefined
+        template, which at the moment can only be the string ``"skops"``, which
+        is a template provided by ``skops`` that is geared towards typical
+        sklearn models. If you don't want any prefilled sections, just pass
+        ``None``. If you want custom prefilled sections, pass a ``dict``, where
+        keys are the sections and values are the contents of the sections. Note
+        that when you use no template or a custom template, some methods will
+        not work, e.g. :meth:`Card.add_metrics`, since it's not clear where to
+        put the metrics when there is no template or a custom template.
+
```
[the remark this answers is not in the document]
BenjaminBossan: You mean here in the docstring or in general, somewhere else in the docs?

skops/card/_model_card.py
```diff
@@ -288,31 +415,91 @@ class Card:
...
+        This should be called in case the model was changed. The model argument
+        should be an sklearn model instance, not a path.
+
+        """
+        model_file = self.metadata.to_dict().get("model_file")
+        if model_file:
+            self._add_get_started_code(model_file)
+
+        self._add_model_section(model)
+        self._add_hyperparams(model)
```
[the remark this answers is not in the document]
BenjaminBossan: Hmm, interesting problem actually. We could encounter the following possibilities:

1. `self.model` is a model instance and there _is no_ `metadata.model_file`
2. `self.model` is a model instance and there _is_ a `metadata.model_file`
3. `self.model` is a `str` or `Path` and there _is no_ `metadata.model_file`
4. `self.model` is a `str` or `Path` and there _is_ a `metadata.model_file` and they're _identical_
5. `self.model` is a `str` or `Path` and there _is_ a `metadata.model_file` and they're _different_

I think the _getting started_ code snippet should always refer to the `model_file` from `metadata` if present. For case 3, one could argue that the path from `self.model` should be used for the _getting started_ code snippet, but it's a bit surprising.

For the model diagram and hyperparams, I would argue we should use the data from `self.model`. If so, we will have the discrepancy you mentioned.

My impression is that the more general problem is that users can theoretically specify the model in two independent ways, once via the `model` parameter and once via the `metadata.model_file`. Those two could be completely different models. Ideally, we could agree on a "single source of truth" for the model, but that is kinda outside the scope of this PR.

skops/card/_model_card.py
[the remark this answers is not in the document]
adrinjalali: Probably a separate example.py file I had in mind, with a link to it here. It's major enough to deserve its own example. But I'm also okay if you open an issue for this and do it in a separate PR.

skops/card/_model_card.py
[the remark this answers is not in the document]
adrinjalali: Then let's move this discussion to a new separate issue, and leave this code as is for this PR.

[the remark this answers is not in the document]
BenjaminBossan: I created an issue: https://github.com/skops-dev/skops/issues/247

skops/card/_model_card.py
```diff
@@ -323,29 +510,277 @@ def get_model(self) -> Any:
...
+        taken from the keys and the contents are taken from the values.
+
+        To add to an existing section, use a ``"/"`` in the section name, e.g.:
+
+        ``card.add(**{"Existing section/New section": "content"})``.
+
+        If the parent section does not exist, it will be added automatically.
+
+        To add a section with ``"/"`` in its title (i.e. not inteded as a
+        subsection), escape the slash like so, ``"\\/"``, e.g.:
```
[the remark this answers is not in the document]
adrinjalali: Ok, we can keep it and see what happens then.

skops/card/_model_card.py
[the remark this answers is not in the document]
BenjaminBossan: I created the issue here so that we don't forget: https://github.com/skops-dev/skops/issues/248

BenjaminBossan:
@adrinjalali I think I replied to all your comments, addressed most open points or created an issue for future us.

Here are the things in particular that I changed:

- Remove Python 3.7 compatibility code (requires https://github.com/skops-dev/skops/pull/246)
- Removed empty section in docs
- Don't add empty metrics table by default just to keep parity
- Remove option to select with a list of str (e.g. `card.select(['Foo', 'Bar', 'Baz'])` no longer works)
- Add option to chain selections using `select` method (e.g. `card.select('Foo').select('Bar').select('Baz')` works)

Regarding the last point, it's not quite what you suggested, but I found that having the same method name, `select`, is more intuitive than having to use `select` the first time and `get` (or `__getitem__`) all subsequent times. I could also live with naming all of them `get` instead of `select`, but this way or that, I think the name should be the same.

skops/card/_model_card.py
```diff
@@ -165,6 +198,100 @@ def metadata_from_config(config_path: Union[str, Path]) -> ModelCardData:
...
+        "import json",
+        "import pandas as pd",
+    ]
+    if is_skops_format:
+        lines += [
+            "from skops.io import load",
+            f'model = load("{file_name}")',
+        ]
+    else:  # pickle
+        lines += [f'model = joblib.load("{file_name}")']
```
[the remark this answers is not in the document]
adrinjalali: the consensus from discord seems to be `import skops.io as sio`

skops/card/_model_card.py
```diff
@@ -417,212 +885,222 @@ def add_table(self, folded: bool = False, **kwargs: dict["str", list[Any]]) -> C
...
+
+    def _generate_metadata(self, metadata: ModelCardData) -> Iterator[str]:
+        """Yield metadata in yaml format"""
+        for key, val in metadata.to_dict().items() if metadata else {}:
+            yield aRepr.repr(f"metadata.{key}={val},").strip('"').strip("'")
+
+    def _generate_content(
+        self, data: dict[str, Section], depth: int = 1
+    ) -> Iterator[str]:
+        """Yield title and (formatted) contents"""
```
adrinjalali, line 958:
```suggestion
        """Yield title and (formatted) contents.
        
        It recursively goes through the data and yields the title
        with the appropriate number of `#`s and the associated
        content.
        """
```

BenjaminBossan: I added something similar, changed the wording a bit.

skops/card/_model_card.py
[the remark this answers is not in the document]
BenjaminBossan: Oh well, democracy has spoken. I changed it here, but would prefer to change the other occurrences (docs, examples) for another PR, since this one is already big as is.

BenjaminBossan (commented):
I addresses your comments.

Tests for Python 3.7 still failing until we drop it.

BenjaminBossan:
@adrinjalali CI is green.

skops/card/_model_card.py
```diff
@@ -288,64 +438,373 @@ class Card:
...
+        if leaf_node_name in section:
+            # entry exists, only overwrite content
+            section[leaf_node_name].content = val
+        else:
+            # entry does not exist, create a new one
+            section[leaf_node_name] = Section(title=leaf_node_name, content=val)
+
+        return section[leaf_node_name]
+
+    def _add_model_section(self, model) -> None:
```
adrinjalali, line 744: we talked about exposing these somehow to users with custom templates, how is that done now?

adrinjalali (commented):
I might be missing it, but I don't see this point resolved.

skops/card/_model_card.py
adrinjalali: The thing with these methods is tat they do more than a simple `add`. So what we could do in this PR, is to make them public and change the signature to:

```py
    def add_model_section(self, model, section=None) -> None:
```

Then we'd only accept `section=None` if template is the default template.

BenjaminBossan: I don't remember that we discussed it like that, but I can take a look :)

Presumably, we don't need the `model` argument and use `self.model` instead? Should users be allowed to add the model reprs multiple times into different sections?

adrinjalali: if we can use `self.model`, then this argument shouldn't be here in the first place.

I don't care for now if users end up adding it multiple times. If they do, their model card looks silly.

BenjaminBossan:
@adrinjalali I added the feature you requested. Copied the 
commit message for better visibility:

So far, if users had a custom template (including no template at all),
they would lose the possibility to add some default sections:

- metrics
- model plot
- hyperparamters
- getting started code

Now we expose methods to the users to add those sections with a simple
call (no need to manually format the content and use `card.add`). For this
to work, users have to indicate the desired section, since we would
otherwise not know where to put the content.

During the work on this, I also cleaned up the `Card` tests, which became
messier and messier over time. Related tests are now all contained in a
class, which makes it a little bit easier to see if for a certain
method, all test cases have been covered.

adrinjalali (approved):
Very happy with this now. I'll let @E-Aho have a look.

E-Aho:
In any case, happy for @adrinjalali to merge this in, the comment I made can easily be tweaked after merging.

skops/card/_model_card.py
```diff
@@ -165,6 +165,129 @@ def metadata_from_config(config_path: Union[str, Path]) -> ModelCardData:
...
+        The section name consisting potentially of multiple subsections. It has
+        to be ensured beforhand that this is not an empty string.
+
+    Returns
+    -------
+    parts : list of str
+        The individual (sub)sections.
+
+    """
+    placeholder = "$%!?"  # arbitrary sting that never appears naturally
```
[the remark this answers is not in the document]
BenjaminBossan: It's a valid concern. I had a regex version that should achieve the same:

```python
def split_subsection_names(key: str) -> list[str]:
    # Use a negative lookbehind assert to make sure that the "/" character is
    # not preceded by a backslash
    parts = re.split(r"/(?<!\\/)", key)
    res = [part.strip().replace("\\", "") for part in parts]
    return res
```

Discussed with @adrinjalali and we found it harder to understand and performance-wise, it was the same.

I haven't thought through the implication of using control codes instead. Maybe we can change it later? This shouldn't really affect users.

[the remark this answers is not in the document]
E-Aho: Happy to change later, I'll merge this in and maybe raise a small PR to do the above over the weekend

adrinjalali:
Woohoo, this is massive! 🥳

PR #204, opened 2022-10-26
FEAT Audit before loading a skops file

adrinjalali:
This adds auditing before load for the `skops` file format.

It creates a tree of nodes by traversing the `state` json stored in the `.skops` file, and loads the information in memory w/o loading any modules or constructing any instances.

Then we can check this tree for existing types/functions and report things which are not trusted. The user then has to pass this list to a `load`/`loads` function to successfully load the `.skops` file:

```py
    from sklearn.linear_model import LogisticRegression
    from skops.io import dump, load

    clf = LogisticRegression(random_state=0, solver="liblinear")
    clf.fit(X_train, y_train)
    from skops.io import dumps, loads
    serialized = dumps(clf)
	# either blindly trust the source
    loaded = loads(serialized, trusted=True)

	# or load by passing a trusted list
    from skops.io import get_untrusted_types
    untrusted_types = get_untrusted_types(file="my-logistic-regression.skops")
    print(untrusted_types)

    loaded = load("my-logistic-regression.skops", trusted=untrusted_types)
```


adrinjalali:
@skops-dev/maintainers working on some tests and doing a self review, which will add a bunch to the PR. But how much should we include in this PR? There is a ton of work which can be done later. This PR as is allows the tests to pass as long as the user trusts the input.

Otherwise there are issues which I'm fixing, and there is also work to actually fill a bunch of security holes. I'm not sure how much we want to go on, on the same PR.

BenjaminBossan:
> Otherwise there are issues which I'm fixing, and there is also work to actually fill a bunch of security holes. I'm not sure how much we want to go on, on the same PR.

I don't think we need to be too strict for the initial release, as long as it's clear to the user what they can and what they can't expect. Regarding this, could you please summarize what you want to have in this PR? E.g. the `load` docstring says:

```
    trusted: bool, or list of str, default=False
        If ``True``, the object will be loaded without any security checks. If
        ``False``, the object will be loaded only if there are only trusted
        objects in the dumped file. If a list of strings, the object will be
        loaded only if there are only trusted objects and objects of types
        listed in ``trusted`` are in the dumped file.
```

but at least in the tests, I don't see the list of string option being used. Does it work? How would I specify the types? And if `False`, how is it determined what the "trusted objects" are? Those would be some of the questions I would have as a user looking to adopt the feature.

adrinjalali:
@BenjaminBossan I completely agree with all you said, I was gonna add those to this PR anyway. So I think we're on the same page. I'll get this PR to a reviewable state and ping back.

adrinjalali:
@BenjaminBossan API question. Which one would you prefer?

```py
load("file.skops", trusted=["mymodule.myclass"])
# or
load("file.skops", trusted=["mymodule.myclass", "builtins.int", "builtins.dict", ...])
```

As in, should we always trust the types we usually trust, or should we require user to even pass those if they want to customize trusted types?

I personally lean towards the first option.

BenjaminBossan:
> As in, should we always trust the types we usually trust, or should we require user to even pass those if they want to customize trusted types?
> 
> I personally lean towards the first option.

I think this type of question pops up quite often in different disguises, like each time you have some allow list and disallow list. In this case, the second option would be far too much work and would be error prone, so option 1 looks better. I wonder, however, if we can make it so that if needed, a user could still choose not to trust all the defaults.

adrinjalali:
@E-Aho your thoughts on this one are also appreciated.

skops/io/_numpy.py
```diff
@@ -50,39 +49,57 @@ def ndarray_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
...
-        for i, v in enumerate(tmp):
-            val[i] = v
-    else:
-        val = np.array(tmp, dtype="O")
-    return val
+class NdArrayNode(Node):
+    def __init__(self, state, load_context: LoadContext, trusted=False):
+        super().__init__(state, load_context, trusted)
+        self.type = state["type"]
+        self.trusted = self._get_trusted(trusted, ["numpy.ndarray"])
```
E-Aho, line 56: I feel it might be neater to avoid having to hard-code the types as strings in `_get_trusted`.

My understanding is that this was done to allow users to pass in strings to the `load` that they want to trust, which makes sense to me.

Maybe we could write a helper function that takes in a type (e.g `np.ndarray`), and returns `f"{x.__module__}.{x.__name__}"` or something similar so we don't need to write in the string ourselves

E-Aho (commented):
I'll give this a more thorough look through when I'm free this evening/weekend, but overall I think this looks like a really valuable addition. One minor question for the time being :)

skops/io/_persist.py
```diff
@@ -133,14 +144,64 @@ def loads(data):
...
+        content = io.BytesIO(data)
+    else:
+        content = file
+
+    with ZipFile(content, "r") as zip_file:
+        schema = json.loads(zip_file.read("schema.json"))
+        tree = get_tree(schema, load_context=LoadContext(src=zip_file))
+        untrusted_types = tree.get_unsafe_set()
+
+    return list(untrusted_types)
```
BenjaminBossan, line 207: Similar argument as above, do we want to return `sorted(untrusted_types)` here?

skops/io/_sklearn.py
```diff
@@ -87,50 +87,65 @@ def reduce_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
     return res
 
 
-def reduce_get_instance(state, load_context, constructor):
-    reduce = state["__reduce__"]
-    args = get_instance(reduce["args"], load_context)
-    instance = constructor(*args)
+class ReduceNode(Node):
+    def __init__(
+        self, state, load_context: LoadContext, constructor=None, trusted=False
```
BenjaminBossan, line 92: Why have default `constructor=None`, is that ever a valid value?

skops/io/tests/test_audit.py
```diff
@@ -0,0 +1,111 @@
...
+    def __init__(self, value):
+        self.value = value
+
+
+@pytest.mark.parametrize(
+    "module_name, type_name, trusted, expected",
+    [
+        ("sklearn", "Pipeline", ["sklearn.Pipeline"], True),
+        ("sklearn", "Pipeline", ["sklearn.preprocessing.StandardScaler"], False),
+        ("sklearn", "Pipeline", True, True),
```
BenjaminBossan, line 26: For completeness, can we add a test for trusting a builtin type?

skops/io/tests/test_audit.py
```diff
@@ -0,0 +1,111 @@
...
+
+
+def test_audit_tree_untrusted():
+    var = {"a": CustomType(1), 2: CustomType(2)}
+    state = dict_get_state(var, SaveContext(None, 0, {}))
+    node = DictNode(state, LoadContext(None), trusted=False)
+    with pytest.raises(
+        TypeError, match="Untrusted types found in the file: {'test_audit.CustomType'}."
+    ):
+        audit_tree(node, trusted=False)
```
BenjaminBossan, line 41: Instead of testing `audit_tree`, which, AFAICT, would never be used by the user directly, could we test `loads` with the same logic?

skops/io/tests/test_audit.py
```diff
@@ -0,0 +1,111 @@
...
+def test_Node_get_trusted(trusted, defaults, expected):
+    assert Node._get_trusted(trusted, defaults) == expected
+
+
+@pytest.mark.parametrize(
+    "values, is_safe",
+    [
+        ([1, 2], True),
+        ([1, {1: 2}], True),
+        ([1, {1: CustomType(1)}], False),
```
BenjaminBossan, line 79: For completeness, let's also check a builtin type which is not a default trusted type, e.g. `eval` (and maybe a 3rd party but not custom type as well?)

skops/io/_dispatch.py
```diff
@@ -2,34 +2,233 @@
...
+            return self._is_safe
+
+        # if trusted is set to True, we don't do any safety checks.
+        if self.trusted is True:
+            self._is_safe = True
+            return True
+
+        is_safe = self.is_self_safe
+
+        for child, _type in self.children.items():
```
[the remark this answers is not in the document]
BenjaminBossan: I wonder if we should make `children` a list of tuples instead of a dict, where the first element is the attribute itself and not its name. The reason why is that if I'm not mistaken, we always do `getattr(self, child)` (where `child` is the name), so would it not be better to store the attribute itself instead of `getattr`ing it with its name? Besides performance, this also leaves a bit less room for error, e.g. that I have a typo in the name when setting `children`. Obviously, we have to use a list of tuples instead of a dict for this to work, but there is no strict reason why it needs to be a dict, right?

skops/io/_dispatch.py
```diff
@@ -2,34 +2,233 @@
...
+            if _type is list:
+                is_safe = is_safe and self._get_iterable_safety(getattr(self, child))
+            elif _type is dict:
+                is_safe = is_safe and self._get_iterable_safety(
+                    getattr(self, child).values()
+                )
+            elif _type is Node:
```
BenjaminBossan, line 125: Is there a specific reason why it should be `Node`? Or is it just an arbitrary guard value?

BenjaminBossan (changes requested):
This looks quite good, well done, I think this time around we have a solution that will stick.

This is quite a big PR, so there might be things I missed or misunderstood. Please take a look at the comments I left. Apart from those, I have a few high level remarks:

1. Some lines appear not to be covered, even though it looks like they should be. We should probably understand why that is before merging.
2. I wonder if we can somehow ensure that during a node's `__init__`, no object is constructed or module imported. Yes, of course we could just look hard at the code and make sure that nothing ever slips through, but I would like some automatic checking. I have no concrete idea honestly, but maybe we can somehow use the `LoadContext` and have separate modes on it, audit mode and construct mode, and during construction, we check that construct mode is set or else fail. Then, we only set construct mode temporarily inside of `construct`, so if some code accidentally constructs an object during the auditing part, it would fail.
3. Is there something we can do to ensure that the children of a node are set correctly? It seems to me that if we do a mistake there, it would leave a backdoor open. IIUC, we only have 3 possibilities for child nodes, a list, a dict or something else. Can we maybe infer the right way to check just from looking at `state`? I think that would be a huge simplification. Then we don't need custom `__init__` in the node subclasses, we just store `self.state = state` in `Node`, thus we don't need to think about what all the children are, which can be error prone. For this to work, we might have to ensure that our `state`s always have the same structure, and during loading verify that this structure is found, but it would be a worthwhile trade-off IMO (if it works).
4. API-wise, as a user, should I be able to say `load(..., trusted=["sklearn.*"])` or something like that? I.e. have a way to blanket trust a module as a whole. Maybe as a future addition?
5. I think the tests should be extended a bit to cover a handful of realistic use cases we could encounter in the wild. As an example, I would like to see a`Pipeline` containing a `FeatureUnion`, consisting of a couple of estimators in total. The test should check that if I only trust a subset of those estimators, the whole pipeline is not trusted. Or if I have a `FunctionTransformer` with a custom function or numpy function, it is not trusted unless that function is allowed.

skops/io/_dispatch.py
```diff
@@ -2,34 +2,233 @@
...
+        This property only checks the type of the node, not its children.
+        """
+        return check_type(self.module_name, self.class_name, self.trusted)
+
+    @property
+    def is_safe(self):
+        """Trie only if the node and all its children are safe."""
+        # the safety value is cached.
+        if self._is_safe is not None:
+            return self._is_safe
```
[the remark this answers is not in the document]
adrinjalali: removed

skops/io/_dispatch.py
```diff
@@ -2,34 +2,233 @@
...
+
+        # if trusted is set to True, we don't do any safety checks.
+        if self.trusted is True:
+            self._is_safe = True
+            return True
```
adrinjalali, line 114: for now we're not passing `trusted` to these classes, when we do, this like would be covered.

skops/io/_general.py
```diff
@@ -113,17 +160,37 @@ def function_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
...
+    def _get_function_name(self):
+        return self.content["module_path"] + "." + self.content["function"]
+
+    @property
+    def is_safe(self):
+        return self._get_function_name() in self.trusted
+
+    def get_unsafe_set(self):
+        if self.is_safe:
+            return set()
```
adrinjalali, line 183: same here, once we pass `trusted` down or when we start trusting a few functions from numpy/scipy, this would be covered.

skops/io/_numpy.py
```diff
@@ -50,39 +49,57 @@ def ndarray_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
...
+            )  # type: ignore
+            self.children = {}
+        elif self.type == "json":
+            self.shape = get_tree(state["shape"], load_context)
+            self.content = [
+                get_tree(o, load_context) for o in state["content"]  # type: ignore
+            ]
+            self.children = {"shape": Node, "content": list}
+        else:
+            raise ValueError(f"Unknown type {self.type}.")
```
adrinjalali, line 69: this only should happen if the given file is malformed, which we're not testing for now.

skops/io/_scipy.py
```diff
@@ -31,16 +32,23 @@ def sparse_matrix_get_state(obj: Any, save_context: SaveContext) -> dict[str, An
...
-        raise TypeError(
-            f"Cannot load object of type {state['__module__']}.{state['__class__']}"
-        )
+class SparseMatrixNode(Node):
+    def __init__(self, state, load_context: LoadContext, trusted=False):
+        super().__init__(state, load_context, trusted)
+        self.type = state["type"]
+        self.trusted = self._get_trusted(trusted, ["scipy.sparse.spmatrix"])
+        if self.type != "scipy":
+            raise TypeError(
```
adrinjalali, line 41: this should also only happen when input is malformed.

skops/io/_sklearn.py
```diff
@@ -87,50 +87,65 @@ def reduce_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
...
+    def _construct(self):
+        args = self.args.construct()
+        instance = self.constructor(*args)
+        attrs = self.attrs.construct()
+        if not attrs:
+            # nothing more to do
+            return instance
+
+        if isinstance(args, tuple) and not hasattr(instance, "__setstate__"):
+            raise UnsupportedTypeException(
```
adrinjalali, line 110: this would be the case on some odd types which we're not testing.

skops/io/_sklearn.py
```diff
@@ -87,50 +87,65 @@ def reduce_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
...
+                f"Objects of type {self.constructor} are not supported yet"
+            )
+
+        if hasattr(instance, "__setstate__"):
+            instance.__setstate__(attrs)
+        else:
+            instance.__dict__.update(attrs)
```
adrinjalali, line 117: same here about odd types.

adrinjalali (commented):
> 1. Some lines appear not to be covered, even though it looks like they should be. We should probably understand why that is before merging.

I've left comments on those lines now.

> 2. I wonder if we can somehow ensure that during a node's `__init__`, no object is constructed or module imported. Yes, of course we could just look hard at the code and make sure that nothing ever slips through, but I would like some automatic checking. I have no concrete idea honestly, but maybe we can somehow use the `LoadContext` and have separate modes on it, audit mode and construct mode, and during construction, we check that construct mode is set or else fail. Then, we only set construct mode temporarily inside of `construct`, so if some code accidentally constructs an object during the auditing part, it would fail.

We certainly should work on that, but I've been thinking of a series of following "_hardenning_" PRs which work on improving the security of this implementation. We could think of a few ways, for instance, patching `import_module` to prevent imports during `__init__` maybe.

> 3. Is there something we can do to ensure that the children of a node are set correctly? It seems to me that if we do a mistake there, it would leave a backdoor open. IIUC, we only have 3 possibilities for child nodes, a list, a dict or something else. Can we maybe infer the right way to check just from looking at `state`? I think that would be a huge simplification. Then we don't need custom `__init__` in the node subclasses, we just store `self.state = state` in `Node`, thus we don't need to think about what all the children are, which can be error prone. For this to work, we might have to ensure that our `state`s always have the same structure, and during loading verify that this structure is found, but it would be a worthwhile trade-off IMO (if it works).

Another thing we can do is to pass `allowed_types` to `get_tree` to limit the type of node which can be created. I don't think we can/should leave `state` as is, we should parse the data during `__init__` and load those values; e.g. we read the numpy data in memory from the zipfile in `__init__`, and construct the array later.

> 4. API-wise, as a user, should I be able to say `load(..., trusted=["sklearn.*"])` or something like that? I.e. have a way to blanket trust a module as a whole. Maybe as a future addition?

I don't think users should ever do `trusted=["sklearn.*]` because one can always find things like `os` through that path. But API wise, adding it later would be easy.

> 5. I think the tests should be extended a bit to cover a handful of realistic use cases we could encounter in the wild. As an example, I would like to see a`Pipeline` containing a `FeatureUnion`, consisting of a couple of estimators in total. The test should check that if I only trust a subset of those estimators, the whole pipeline is not trusted. Or if I have a `FunctionTransformer` with a custom function or numpy function, it is not trusted unless that function is allowed.

I agree, but I feel like doing that while working on adding `sklearn` safe types would be easier. Trying to limit this PR as much as we can. If you think we should add a specific test, we can do though. Also, I feel a bit exhausted working on this PR, if you add those tests here, I definitely wouldn't complain :P

BenjaminBossan (commented):
> We certainly should work on that, but I've been thinking of a series of following "hardenning" PRs which work on improving the security of this implementation.

Okay, we can do that as a follow up.

> I don't think we can/should leave state as is, we should parse the data during `__init__` and load those values; e.g. we read the numpy data in memory from the zipfile in `__init__`, and construct the array later.

Could you please elaborate on that, I don't understand it. Probably it also relates to this comment elsewhere:

> we need to do the traverse in `__init__`, and this traversal is node specific. We almost always call `get_tree` on children in `__init__`.

I checked each `self.children = ...` and none of them are calling `get_tree`.

I tried to modify the code as following to avoid having a bunch of different attributes for every node type and it worked (only tried for `DictNode`):

```python
class DictNode(Node):
    def __init__(self, state, load_context: LoadContext, trusted=False):
        super().__init__(state, load_context, trusted)
        self.trusted = self._get_trusted(trusted, ["builtins.dict"])
        # ideally, we could automatically infer the children from the state...
        self.children = {
            "key_types": get_tree(state["key_types"], load_context),
            "content": {
                key: get_tree(value, load_context)
                for key, value in state["content"].items()
            },
        }
        # no other custom attributes stored

    def _construct(self):
        key_types = self.children["key_types"]
        content_ = self.children["content"]
        content = gettype(self.module_name, self.class_name)()
        key_types = key_types.construct()
        for k_type, (key, val) in zip(key_types, content_.items()):
            if key == "categories_":
                pass
            content[k_type(key)] = val.construct()
        return content

# inside of get_unsafe_set, the loop is replaced with:

    def get_unsafe_set(self):
        ...
        for child in self.children.values():
            if child is None:
                continue

            # Get the safety set based on the type of the child. In most cases
            # other than ListNode and DictNode, children are all of type Node.
            if isinstance(child, list):
                for value in child:
                    res.update(value.get_unsafe_set())
            elif isinstance(child, dict):
                for value in child.values():
                    res.update(value.get_unsafe_set())
            elif isinstance(child, Node):
                res.update(child.get_unsafe_set())
            else:
                raise ValueError(f"Unknown type {type(child)}.")
        ...
```

It's not exactly my initial proposal with saving `state` but still addresses my concern. WDYT?

> I don't think users should ever do trusted=["sklearn.*] because one can always find things like os through that path. 

Even if a module in sklearn imports `os`, the module name would still not be `sklearn`, so there is no match, right?

> I agree, but I feel like doing that while working on adding sklearn safe types would be easier. 

About that, what's the plan? Enumerating each and every sklearn class and function that's safe? And keeping it up-to-date with each release, across multiple versions? Sounds infeasible.

skops/io/_utils.py
```diff
@@ -156,15 +155,19 @@ def get_state(value, save_context):
...
     # the original error alongside the json error.
+
+    # TODO: This should help with fixing recursive references.
+    # if id(value) in save_context.memo:
+    #     return {
+    #         "__module__": None,
+    #         "__class__": None,
+    #         "__id__": id(value),
+    #         "__loader__": "CachedNode",
+    #     }
```
E-Aho, line 166: I think this should work ok, but we would need to restructure how things get memoized.

Right now, things only get memoized once they've been fully constructued, so if an object has a recursive attribute, it never actually gets to be in the memo.

I thought about this originally, and started exploring if there's a nice way to hold a reference to an object that isn't initialized yet, so we could do something like:


```
    _obj = Object()  # pseudocode, not the right way to do this

    save_context.memoize[_obj, id]
    res = _get_state(value, save_context)

    _obj.assign(res)   

    return _obj
```

Still haven't fully thought this through, but I think we would need to do something like this to allow circular self-references to work the right way.

skops/io/_utils.py
```diff
@@ -58,9 +57,9 @@ def _import_obj(module, cls_or_func, package=None):
     return getattr(importlib.import_module(module, package=package), cls_or_func)
 
 
-def gettype(state):
-    if "__module__" in state and "__class__" in state:
-        return _import_obj(state["__module__"], state["__class__"])
+def gettype(module_name, cls_or_func):
+    if module_name and cls_or_func:
+        return _import_obj(module_name, cls_or_func)
```
E-Aho, line 62: I wanted to check something here:

As it stands, is there any way loading a .skops file could redefine something in the global namespace?

My understanding (correct me if I'm wrong), is that this could only happen if:
* Code is defined that does that before calling `loads`
* Something does this in a user's imported module 

In either of those cases, this isn't a vulnerability with skops itself, so I think it's ok, but I wanted to make sure I've not missed somewhere that global namespaces could be changed during `load`, as that could lead to a vulnerability.

In other words, there's not currently a way someone could structure a `.skops` file that redefines a type we deem "trusted", like `np.random.Generator`, right?

skops/io/_persist.py
E-Aho: It's a weird quirk that dicts have keys sorted by insert order since Python 3.7, but sets still don't.

Example:

```
foo = {1, 10, 5}
list(foo) # {1, 10, 5}

foo.add(3)
list(foo)  # {1, 10, 3, 5}
```

E-Aho (commented):
Had a few more thoughts :)

adrinjalali (commented):
> Even if a module in sklearn imports os, the module name would still not be sklearn, so there is no match, right?

when we persist yes, but somebody could curate a `.skops` file where they do `sklearn.datasets._base` as module and `os` as class name and we'd import it.

> About that, what's the plan? Enumerating each and every sklearn class and function that's safe? And keeping it up-to-date with each release, across multiple versions? Sounds infeasible.

It sounds feasible to me, but would require some automation to streamline it. By enumerating, I would think of getting all estimators through sklearn's API rather than adding manually though.

> It's not exactly my initial proposal with saving state but still addresses my concern. WDYT?

This doesn't look neat to me, but while doing it I found one bug, so it's a good pattern I'd say. Also, had to modify the `get_unsafe_set` quit a bit as a result.

skops/io/_dispatch.py
```diff
@@ -1,35 +1,256 @@
...
+        Returns
+        -------
+        unsafe_set : set
+            A set of unsafe types.
+        """
+        if hasattr(self, "_computing_unsafe_set"):
+            # this means we're already computing this node's unsafe set, so we
+            # return an empty set and let the computation of the parent node
+            # continue. This is to avoid infinite recursion.
+            return set()
```
adrinjalali, line 151: this is isn't tested since we haven't figured the whole recursive pointers thing out.

skops/io/_dispatch.py
```diff
@@ -1,35 +1,256 @@
...
+                elif isinstance(child, Node):
+                    # delegate to the child Node
+                    res.update(child.get_unsafe_set())
+                elif type(child) is type:
+                    # the if condition bellow is not merged with the previous
+                    # one because if the above condition is True, the following
+                    # conditions about BytesIO, etc should be ignored.
+                    if not check_type(get_module(child), child.__name__, self.trusted):
+                        # if the child is a type, we check its safety
+                        res.add(get_module(child) + "." + child.__name__)
```
adrinjalali, line 182: for all cases where the child is a type, we have it as trusted, This is only used in reduce.

skops/io/_dispatch.py
```diff
@@ -1,35 +1,256 @@
...
+
+
+class CachedNode(Node):
+    def __init__(self, state, load_context: LoadContext, trusted=False):
+        # we pass memoize as False because we don't want to memoize the cached
+        # node.
+        super().__init__(state, load_context, trusted, memoize=False)
+        self.trusted = True
+        self.cached = load_context.get_object(state.get("__id__"))
+        self.children = {}  # type: ignore
```
adrinjalali, line 206: cache node is used for recursive pointers as well.

adrinjalali (commented):
I think this is ready for another review @BenjaminBossan

BenjaminBossan (changes requested):
Pretty good work, I'm happy with the result. There are some small things left to do, please check my comments, but overall exceptional work.

> when we persist yes, but somebody could curate a .skops file where they do sklearn.datasets._base as module and os as class name and we'd import it.

Yes, good point. But I wonder how much of a problem that is. If sklearn imports `foo` somewhere, we cannot guarantee that `foo` will not be imported at some point, even if users pass fine grained types to allow. Therefore, we already need to assume that importing `foo` is safe.

Now, using `foo` could still be dangerous. However, when we create the instances, we could check if the path of `foo` corresponds to `sklearn.*` and if not, raise an error. That way, we could still prevent its usage, if I'm not missing something.

> It sounds feasible to me, but would require some automation to streamline it. By enumerating, I would think of getting all estimators through sklearn's API rather than adding manually though.

Okay, let's see how it'll work out in practice. Do you plan to include that before next release?

> This doesn't look neat to me, but while doing it I found one bug, so it's a good pattern I'd say. Also, had to modify the get_unsafe_set quit a bit as a result.

Not sure what part didn't look neat, but the way you refactored corresponds to my intent and is cleaner IMO, so I'm happy with the outcome.

skops/io/_sklearn.py
```diff
@@ -87,50 +87,67 @@ def reduce_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
...
 def tree_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
     state = reduce_get_state(obj, save_context)
-    state["__loader__"] = "tree_get_instance"
+    state["__loader__"] = "TreeNode"
     return state
 
 
-def tree_get_instance(state, load_context):
-    return reduce_get_instance(state, load_context, constructor=Tree)
+class TreeNode(ReduceNode):
```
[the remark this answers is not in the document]
adrinjalali: we could call it `SklearnTreeTreeNode` 😁

adrinjalali (commented):
> Yes, good point. But I wonder how much of a problem that is. If sklearn imports foo somewhere, we cannot guarantee that foo will not be imported at some point, even if users pass fine grained types to allow. Therefore, we already need to assume that importing foo is safe.
> 
> Now, using foo could still be dangerous. However, when we create the instances, we could check if the path of foo corresponds to sklearn.* and if not, raise an error. That way, we could still prevent its usage, if I'm not missing something.

Yes, that'd be interesting, but we should do the check before creating the instance. We can do that in a followup PR, and add extra checks for it.

>     It sounds feasible to me, but would require some automation to streamline it. By enumerating, I would think of getting all estimators through sklearn's API rather than adding manually though.
> 
> Okay, let's see how it'll work out in practice. Do you plan to include that before next release?

No, I think we can release a first version w/o trusting much from sklearn.

adrinjalali:
I think I'm happy with this now. It could be merged and we could release early and get feedback, and in the meantime work on the lot left to improve.

cc @skops-dev/maintainers

BenjaminBossan (approved):
Great work, I think this is now ready to be merged. Everything that's still open, we can work on later, e.g. I'd like to add some light typing (-:

@e-aho Do you want to give this another pass too?

E-Aho (approved):
I caught a tiny typo in a docstring, but overall LGTM!

adrinjalali:
Feel free to hit the "Squash and merge" button then @E-Aho :) (we almost always squash and merge)

BenjaminBossan:
> Feel free to hit the "Squash and merge" button then E-Aho :) (we almost always squash and merge)

We usually try to craft a nice commit message, not just using the GH suggestion.

E-Aho:
All set :) I

Side note, is there a list of the prefixes for commits? [FEAT, FIX, DOC, etc] that goes at the beginning of most of the main branch commits?

adrinjalali:
I don't have a list, but I use:

- DOC: documentation changes
- FEAT/FEA: new major features
- ENH: enhancements to existing features
- CI: continuous integration, sometimes overlaps with MNT
- MNT/MAINT: maintenance, technical debt, etc
- FIX: bug fixes
- TST: new tests, refactoring tests
- PERF: performance improvements

BenjaminBossan:
I was also searching for such a list. How about adding it to the contribution guide?

adrinjalali:
We should add it to the maintainers guide instead of contributing guide (we don't have the separation now). Maintainers can/should fix commit messages/titles before merging, and I don't think we should burden first time contributors with such details.

PR #205, opened 2022-10-28
load model from path/str to generate Card

p-mishra1:
#96 Added required changes and followed the pattern @BenjaminBossan advised.
Creating PR from another branch as the last one had too many commits and reverts.
@BenjaminBossan have a look.

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -165,12 +165,14 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
 
 
 def _load_model(model: Any) -> Any:
-    """Loads the mddel if provided a file path, if already a model instance return it unmodified.
+    """Loads the mddel if provided a file path, if already a model instance return it
+       unmodified.
 
     Parameters
     ----------
     model : pathlib.path, str, or sklearn estimator
-        Path/str or the actual model instance. if a Path or str, loads the model on first call.
+        Path/str or the actual model instance. if a Path or str, loads the model on
+        first call.
 
     Returns
     -------
@@ -184,7 +186,7 @@ def _load_model(model: Any) -> Any:
 
     model_path = Path(model)
     if not model_path.exists():
-        raise ValueError("Model file does not exist")
+        raise FileNotFoundError(f"File is not present:{model_path}")
 
     if model_path.suffix in (".pkl", ".pickle"):
         model = joblib.load(model_path)
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -167,11 +167,11 @@ def metadata_from_config(config_path: Union[str, Path]) -> ModelCardData:
 
 def _load_model(model: Any) -> Any:
     """Loads the mddel if provided a file path, if already a model instance return it
-       unmodified.
+    unmodified.
 
     Parameters
     ----------
-    model : pathlib.path, str, or sklearn estimator
+    model : pathlib.Path, str, or sklearn estimator
         Path/str or the actual model instance. if a Path or str, loads the model.
 
     Returns
@@ -212,9 +212,7 @@ class Card:
     ----------
     model: pathlib.path, str, or sklearn estimator object
         ``Path``/``str`` of the model or the actual model instance that will be
-        documented. If a ``Path`` or ``str`` is provided, model will be loaded
-        on first use.
-
+        documented. If a ``Path`` or ``str`` is provided, model will be loaded.
 
     model_diagram: bool, default=True
         Set to True if model diagram should be plotted in the card.
@@ -306,7 +304,7 @@ class Card:
 
     def get_model(self) -> Any:
         """Returns sklearn estimator object if ``Path``/``str``
-            is provided.
+        is provided.
 
         Returns
         -------
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -213,7 +213,8 @@ class Card:
     Parameters
     ----------
     model: pathlib.path, str, or sklearn estimator object
-        Model that will be documented.
+        Path/str of the model or the actual model instance that will be documented.
+        if a Path or str is provided, model will be loaded.
 
     model_diagram: bool, default=True
         Set to True if model diagram should be plotted in the card.
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -210,7 +210,7 @@ class Card:
 
     Parameters
     ----------
-    model: pathlib.path, str, or sklearn estimator object
+    model: pathlib.Path, str, or sklearn estimator object
         ``Path``/``str`` of the model or the actual model instance that will be
         documented. If a ``Path`` or ``str`` is provided, model will be loaded.
 
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -11,8 +11,6 @@ from reprlib import Repr
 from typing import Any, Optional, Union
 
 import joblib
-from yaml import load
-
 from huggingface_hub import CardData, ModelCard
 from sklearn.utils import estimator_html_repr
 from tabulate import tabulate  # type: ignore
@@ -165,17 +163,18 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
 
     return card_data
 
-def _load_model(model:Any)->Any:
+
+def _load_model(model: Any) -> Any:
     """Loads the mddel if provided a file path, if already a model instance return it unmodified.
 
     Parameters
     ----------
-    model : pathlib.path, str, or sklearn estimator 
+    model : pathlib.path, str, or sklearn estimator
         Path/str or the actual model instance. if a Path or str, loads the model on first call.
 
     Returns
     -------
-    model : object 
+    model : object
         Model instance.
 
     """
@@ -189,16 +188,17 @@ def _load_model(model:Any)->Any:
 
     if model_path.suffix in (".pkl", ".pickle"):
         model = joblib.load(model_path)
-    elif model_path.suffix ==".skops":
+    elif model_path.suffix == ".skops":
         model = load(model_path)
     else:
         msg = (
             f"Cannot interpret model suffix {model_path.suffix}, should be"
             "'.pkl', '.pickle', or '.skops'"
-         )
+        )
         raise ValueError(msg)
-    
+
     return model
+
 
 class Card:
     """Model card class that will be used to generate model card.
@@ -425,7 +425,6 @@ class Card:
         for metric, value in kwargs.items():
             self._eval_results[metric] = value
         return self
-
 
     def _generate_card(self) -> ModelCard:
         """Generate the ModelCard object
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -10,11 +10,15 @@ from pathlib import Path
 from reprlib import Repr
 from typing import Any, Optional, Union
 
+import joblib
+from yaml import load
+
 from huggingface_hub import CardData, ModelCard
 from sklearn.utils import estimator_html_repr
 from tabulate import tabulate  # type: ignore
 
 import skops
+from skops.io import load
 
 # Repr attributes can be used to control the behavior of repr
 aRepr = Repr()
@@ -161,6 +165,40 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
 
     return card_data
 
+def _load_model(model:Any)->Any:
+    """Loads the mddel if provided a file path, if already a model instance return it unmodified.
+
+    Parameters
+    ----------
+    model : pathlib.path, str, or sklearn estimator 
+        Path/str or the actual model instance. if a Path or str, loads the model on first call.
+
+    Returns
+    -------
+    model : object 
+        Model instance.
+
+    """
+
+    if not isinstance(model, (Path, str)):
+        return model
+
+    model_path = Path(model)
+    if not model_path.exists():
+        raise ValueError("Model file does not exist")
+
+    if model_path.suffix in (".pkl", ".pickle"):
+        model = joblib.load(model_path)
+    elif model_path.suffix ==".skops":
+        model = load(model_path)
+    else:
+        msg = (
+            f"Cannot interpret model suffix {model_path.suffix}, should be"
+            "'.pkl', '.pickle', or '.skops'"
+         )
+        raise ValueError(msg)
+    
+    return model
 
 class Card:
     """Model card class that will be used to generate model card.
@@ -172,7 +210,7 @@ class Card:
 
     Parameters
     ----------
-    model: estimator object
+    model: pathlib.path, str, or sklearn estimator object
         Model that will be documented.
 
     model_diagram: bool, default=True
@@ -254,12 +292,27 @@ class Card:
         model_diagram: bool = True,
         metadata: Optional[CardData] = None,
     ) -> None:
-        self.model = model
+        self._model = model
         self.model_diagram = model_diagram
         self._eval_results = {}  # type: ignore
         self._template_sections: dict[str, str] = {}
         self._extra_sections: list[tuple[str, Any]] = []
         self.metadata = metadata or CardData()
+
+    @property
+    def model(self):
+        model = _load_model(self._model)
+        if model is not self._model:
+            self._model = model
+        return model
+
+    @model.setter
+    def model(self, model):
+        self._model = model
+
+    @model.deleter
+    def model(self):
+        del self._model
 
     def add(self, **kwargs: str) -> "Card":
         """Takes values to fill model card template.
@@ -372,6 +425,7 @@ class Card:
         for metric, value in kwargs.items():
             self._eval_results[metric] = value
         return self
+
 
     def _generate_card(self) -> ModelCard:
         """Generate the ModelCard object
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -213,8 +213,10 @@ class Card:
     Parameters
     ----------
     model: pathlib.path, str, or sklearn estimator object
-        Path/str of the model or the actual model instance that will be documented.
-        if a Path or str is provided, model will be loaded.
+        ``Path``/``str`` of the model or the actual model instance that will be
+        documented. If a ``Path`` or ``str`` is provided, model will be loaded
+        on first use.
+
 
     model_diagram: bool, default=True
         Set to True if model diagram should be plotted in the card.
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -5,6 +5,7 @@ import json
 import re
 import shutil
 import tempfile
+import zipfile
 from dataclasses import dataclass
 from pathlib import Path
 from reprlib import Repr
@@ -171,8 +172,7 @@ def _load_model(model: Any) -> Any:
     Parameters
     ----------
     model : pathlib.path, str, or sklearn estimator
-        Path/str or the actual model instance. if a Path or str, loads the model on
-        first call.
+        Path/str or the actual model instance. if a Path or str, loads the model.
 
     Returns
     -------
@@ -186,18 +186,16 @@ def _load_model(model: Any) -> Any:
 
     model_path = Path(model)
     if not model_path.exists():
-        raise FileNotFoundError(f"File is not present:{model_path}")
+        raise FileNotFoundError(f"File is not present: {model_path}")
 
-    if model_path.suffix in (".pkl", ".pickle"):
-        model = joblib.load(model_path)
-    elif model_path.suffix == ".skops":
-        model = load(model_path)
-    else:
-        msg = (
-            f"Cannot interpret model suffix {model_path.suffix}, should be"
-            "'.pkl', '.pickle', or '.skops'"
-        )
-        raise ValueError(msg)
+    try:
+        if zipfile.is_zipfile(model_path):
+            model = load(model_path)
+        else:
+            model = joblib.load(model_path)
+    except Exception as ex:
+        msg = f'An "{type(ex).__name__}" occured during model loading.'
+        raise RuntimeError(msg) from ex
 
     return model
 
@@ -297,27 +295,24 @@ class Card:
         model_diagram: bool = True,
         metadata: Optional[CardData] = None,
     ) -> None:
-        self._model = model
+        self.model = model
         self.model_diagram = model_diagram
         self._eval_results = {}  # type: ignore
         self._template_sections: dict[str, str] = {}
         self._extra_sections: list[tuple[str, Any]] = []
         self.metadata = metadata or CardData()
 
-    @property
-    def model(self):
-        model = _load_model(self._model)
-        if model is not self._model:
-            self._model = model
+    def get_model(self) -> Any:
+        """Returns sklearn estimator object if ``Path``/``str``
+            is provided.
+
+        Returns
+        -------
+        model : Object
+            Model instance.
+        """
+        model = _load_model(self.model)
         return model
-
-    @model.setter
-    def model(self, model):
-        self._model = model
-
-    @model.deleter
-    def model(self):
-        del self._model
 
     def add(self, **kwargs: str) -> "Card":
         """Takes values to fill model card template.
@@ -469,7 +464,9 @@ class Card:
                         'clf.predict(pd.DataFrame.from_dict(config["sklearn"]["example_input"]))'
                     )
         if self.model_diagram is True:
-            model_plot_div = re.sub(r"\n\s+", "", str(estimator_html_repr(self.model)))
+            model_plot_div = re.sub(
+                r"\n\s+", "", str(estimator_html_repr(self.get_model()))
+            )
             if model_plot_div.count("sk-top-container") == 1:
                 model_plot_div = model_plot_div.replace(
                     "sk-top-container", 'sk-top-container" style="overflow: auto;'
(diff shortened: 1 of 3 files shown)


skops/card/_model_card.py
```diff
@@ -162,6 +164,42 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
     return card_data
 
 
+def _load_model(model: Any) -> Any:
+    """Loads the mddel if provided a file path, if already a model instance return it unmodified.
+
+    Parameters
+    ----------
+    model : pathlib.path, str, or sklearn estimator
+        Path/str or the actual model instance. if a Path or str, loads the model on first call.
```
BenjaminBossan, line 173: Could you please add line length limits to the docstring as well?

skops/card/_model_card.py
```diff
@@ -162,6 +164,42 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
...
+        Model instance.
+
+    """
+
+    if not isinstance(model, (Path, str)):
+        return model
+
+    model_path = Path(model)
+    if not model_path.exists():
+        raise ValueError("Model file does not exist")
```
BenjaminBossan, line 187: Could you please add a test case for this potential error? Also, `FileNotFoundError` is a more precise error.

skops/card/_model_card.py
```diff
@@ -162,6 +164,42 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
...
+    if model_path.suffix in (".pkl", ".pickle"):
+        model = joblib.load(model_path)
+    elif model_path.suffix == ".skops":
+        model = load(model_path)
+    else:
+        msg = (
+            f"Cannot interpret model suffix {model_path.suffix}, should be"
+            "'.pkl', '.pickle', or '.skops'"
+        )
+        raise ValueError(msg)
```
BenjaminBossan, line 198: Same here, could you please add a test?

skops/card/_model_card.py
```diff
@@ -172,7 +210,7 @@ class Card:
 
     Parameters
     ----------
-    model: estimator object
+    model: pathlib.path, str, or sklearn estimator object
         Model that will be documented.
```
BenjaminBossan, line 214: Could you please extend the docstring, now that we have more options here?

skops/card/tests/test_card.py
```diff
@@ -26,6 +26,29 @@ def fit_model():
...
+def test_load_model(suffix):
+    model0 = LinearRegression(n_jobs=123)
+    save_file = save_model_to_file(model0, suffix)
+    loaded_model_str = _load_model(save_file)
+    save_file_path = Path(save_file)
+    loaded_model_path = _load_model(save_file_path)
+
+    assert loaded_model_str.n_jobs is model0.n_jobs
+    assert loaded_model_path.n_jobs is model0.n_jobs
+    assert loaded_model_path.n_jobs is loaded_model_str.n_jobs
```
BenjaminBossan, line 49:
```suggestion
    assert loaded_model_str.n_jobs == model0.n_jobs
    assert loaded_model_path.n_jobs == model0.n_jobs
    assert loaded_model_path.n_jobs == loaded_model_str.n_jobs
```

Although `is` works in this case, let's use `==` to compare numbers and then we don't have to worry about the size of the number or the Python interpreter :)

skops/card/tests/test_card.py
```diff
@@ -405,6 +428,69 @@ def test_with_metadata(self, card: Card, meth):
...
+    def test_model_card_str(self, suffix):
+        model0 = LinearRegression(n_jobs=123)
+        save_file_str = save_model_to_file(model0, suffix)
+        card_from_str = Card(save_file_str)
+        card_from_model0 = Card(model0)
+
+        assert card_from_model0.model.n_jobs == card_from_str.model.n_jobs
+
+    @pytest.mark.parametrize("suffix", [".pkl", ".pickle", ".skops"])
+    def test_model_card_path(self, suffix):
```
BenjaminBossan, line 442: I think this test could be unified with the previous one, as there is only one line difference, by parametrizing it over whether the `save_file` is converted to `Path`  or not.

skops/card/tests/test_card.py
```diff
@@ -405,6 +428,69 @@ def test_with_metadata(self, card: Card, meth):
...
+        assert card.model is model1
+
+        # re-assigning back to original works
+        card.model = model0
+        assert card.model is model0
+
+    def test_model_is_str_pickle(self, destination_path):
+        model0 = LinearRegression(n_jobs=123)
+        f_name0 = destination_path / "lin_reg.pickle"
+        with open(f_name0, "wb") as f:
```
BenjaminBossan, line 471: Can we use `save_model_to_file` in this test?

skops/card/tests/test_card.py
```diff
@@ -26,6 +26,29 @@ def fit_model():
     return reg
 
 
+def save_model_to_file(model_instance, suffix):
```
BenjaminBossan, line 29: Nice refactor, a lot less _magic_ then using the implicit pytest parameters!

skops/card/tests/test_card.py
```diff
@@ -405,6 +428,69 @@ def test_with_metadata(self, card: Card, meth):
...
+        model0 = LinearRegression(n_jobs=123)
+        save_file = save_model_to_file(model0, suffix)
+        save_file_path = Path(save_file)
+        card_from_path = Card(save_file_path)
+        card_from_model0 = Card(model0)
+
+        assert card_from_model0.model.n_jobs == card_from_path.model.n_jobs
+
+
+class TestCardModelAttribute:
```
BenjaminBossan, line 452: I would like to see an additional test case here, namely that the model is only loaded once.

One way to test this would be to save the model to a file, pass the path to `Card`, then accessing `card.model` should return the model. After this, delete the file and access `card.model` again. This way, if the implementation accidentally loads on each access, the test would raise an error.

Maybe there is a better way to test this, but this seems more robust than, say, patching `_load_model`.

BenjaminBossan (changes requested):
@p-mishra1 Thanks a lot for listening to the feedback, I like the refactor you did. This PR is now close to being ready. I have a few minor comments, please take a look. Also, please add an entry to the `changes.rst`.

skops/card/_model_card.py
p-mishra1: i changed to this "pathlib.path, str, or sklearn estimator object"
what should we add more ?

p-mishra1:
> @p-mishra1 Thanks a lot for listening to the feedback, I like the refactor you did. This PR is now close to being ready. I have a few minor comments, please take a look. Also, please add an entry to the `changes.rst`.

Thanks to you @BenjaminBossan patiently guiding me and giving me feedback every time..

skops/card/_model_card.py
BenjaminBossan: How about something along the lines that if a str or Path is passed, the model will be loaded. This could be surprising behavior for some users and should be mentioned.

skops/card/tests/test_card.py
```diff
@@ -405,6 +428,93 @@ def test_with_metadata(self, card: Card, meth):
...
 
 
+class TestModelCardFromPath:
+    @pytest.mark.parametrize("file_type", [str, Path])
+    @pytest.mark.parametrize("suffix", [".pkl", ".pickle", ".skops"])
+    def test_model_card_str(self, suffix, file_type):
+        model0 = LinearRegression(n_jobs=123)
+        save_file = save_model_to_file(model0, suffix)
+        save_file = file_type(save_file)
+        print(type(save_file))
```
BenjaminBossan, line 438: Please remove :)

BenjaminBossan (commented):
Nice, we're almost done here.

From my point of view, there are only 3 minor issues left:

1. Adjusting the docstring as discussed
2. Remove the debug print (see comment)
3. Add an entry to `changes.rst`

skops/card/tests/test_card.py
p-mishra1: oh, sorry left it by mistake, 
one more thing @BenjaminBossan I'm unable to understand where should i add the changes in `changes.rst` it will go under `v0.3` right.

skops/card/_model_card.py
```diff
@@ -162,6 +164,44 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
...
 
+def _load_model(model: Any) -> Any:
+    """Loads the mddel if provided a file path, if already a model instance return it
+       unmodified.
+
+    Parameters
+    ----------
+    model : pathlib.path, str, or sklearn estimator
+        Path/str or the actual model instance. if a Path or str, loads the model on
+        first call.
```
BenjaminBossan, line 175:
```suggestion
        ``Path``/``str`` of the model or the actual model instance that will be
        documented. If a ``Path`` or ``str`` is provided, model will be loaded
        on first use.
```

p-mishra1: changed.

p-mishra1:
@BenjaminBossan the test were failing on windows so I changed the function 
`def save_model_to_file(model_instance, suffix):
    save_file_handle, save_file = tempfile.mkstemp(suffix=suffix, prefix="skops-test")
    if suffix in (".pkl", ".pickle"):
        with open(save_file, "wb") as f:
            pickle.dump(model_instance, f)
    elif suffix == ".skops":
        dump(model_instance, save_file)
    return save_file_handle, save_file`
    
   like this so that i can close the file before removing it somehow it is not a problem in linux/mac. 
   I'm thinking that is the issue here.

p-mishra1:
@BenjaminBossan do you have an idea why it is failing on windows, i could not figure out exactly why,

BenjaminBossan:
If you can't figure it out, I would be okay with skipping the test on Windows (with a comment explaining why). The error really comes from the test itself and is not something a Windows user would encounter in practice when using the model card.

p-mishra1:
@BenjaminBossan i tried somethings and also worked in the windows environment to mitigate the issue have changed the test a little bit hopefully it will work this time.
give it a look.

skops/card/_model_card.py
```diff
@@ -162,6 +164,44 @@ def metadata_from_config(config_path: Union[str, Path]) -> CardData:
...
+        return model
+
+    model_path = Path(model)
+    if not model_path.exists():
+        raise FileNotFoundError(f"File is not present:{model_path}")
+
+    if model_path.suffix in (".pkl", ".pickle"):
+        model = joblib.load(model_path)
+    elif model_path.suffix == ".skops":
+        model = load(model_path)
```
adrinjalali, line 194: we shouldn't rely on the extension here. We can check if the file is a zip file (`zipfile.is_zipfile(filename)`[¶](https://docs.python.org/3/library/zipfile.html#zipfile.is_zipfile) and load with `skops` if yes, otherwise try to load with `joblib`, and catch any exceptions which might be raised during each of them.

skops/card/_model_card.py
```diff
@@ -254,13 +297,28 @@ def __init__(
...
+            self._model = model
+        return model
+
+    @model.setter
+    def model(self, model):
+        self._model = model
+
+    @model.deleter
+    def model(self):
+        del self._model
```
adrinjalali, line 320: do we need this?

skops/card/_model_card.py
```diff
@@ -254,13 +297,28 @@ def __init__(
...
         self._template_sections: dict[str, str] = {}
         self._extra_sections: list[tuple[str, Any]] = []
         self.metadata = metadata or CardData()
 
+    @property
+    def model(self):
+        model = _load_model(self._model)
+        if model is not self._model:
+            self._model = model
+        return model
```
adrinjalali, line 312: I don't think we should change `model`. We should always keep `model` attribute as is given by the user, and only load when we need it in other methods.

adrinjalali (commented):
thanks @p-mishra1

skops/card/_model_card.py
merveenoyan: can't we just keep a key in config file to store which format the model was serialized in?

skops/card/_model_card.py
merveenoyan: agreed 👍🏻

merveenoyan (commented):
thanks a lot for working on this!

skops/card/_model_card.py
BenjaminBossan: I'm not sure I fully understand. You want the `model` argument passed by the user to remain unchanged, and when we access it in other methods, we check if it's a string/path and load it? So basically swap around the attribute names `model` and `_model`? Or load the model each time it's accessed (which would be wasteful, since it's used, for instance, in `__repr__` so could be accessed quite frequently).

skops/card/_model_card.py
BenjaminBossan: For consistency, the method should exist (assuming we leave the rest as is), no?

skops/card/_model_card.py
adrinjalali: yes, exactly @p-mishra1 .

skops/card/_model_card.py
p-mishra1: i agree depending on file extension is not a good way.
let's say the file loaded indeed was a zip then we will be conforming that it is sklearn estimator object. using this 
`[base.BaseEstimator](https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator)` ?

skops/card/_model_card.py
BenjaminBossan: -1 from me, why potentially reload the model multiple times?

p-mishra1: yes i agree with @BenjaminBossan as there can be multiple calls to `__repr__`.

BenjaminBossan: E.g. imagine I create a card in a repl/notebook and call `add` multiple times. Since the method always returns `self`, each call implicitly calls `repr` and thus loads the model.

BenjaminBossan: One thing I could see as a compromise would be to call `_load_model` each time but also to cache the model. The problem here is that we cannot use the path to the model as the cache key, we would need to use something like md5sum. And the caching decorators from `functools` don't support custom key functions like this (AFAIK), so we would need to write our own cache or add a dependency.

skops/card/_model_card.py
adrinjalali: > can't we just keep a key in config file to store which format the model was serialized in?

We could but I don't think this function is called late enough for that to work.

> let's say the file loaded indeed was a zip then we will be conforming that it is sklearn estimator object. using this
[base.BaseEstimator](https://scikit-learn.org/stable/modules/generated/sklearn.base.BaseEstimator.html#sklearn.base.BaseEstimator) ?

No, if it's a zip file, then it's a `skops` file, which means we can load using skops, otherwise we load with `joblib`, and we don't validate the output. We just catch any exception which might have been raised.

skops/card/_model_card.py
adrinjalali: I almost never see a deleter for properties.

skops/card/_model_card.py
adrinjalali: We can load the model everytime in this PR, and in a separate PR fix the caching issue.

BenjaminBossan: > We can load the model everytime in this PR, and in a separate PR fix the caching issue.

Okay, let's do it this way and add a caching decorator in a separate PR (or use a [library](https://toolz.readthedocs.io/en/latest/api.html?highlight=memoize#toolz.functoolz.memoize) for it). I guess it's okay to add it later (though I would have liked to have included here) because this functionality was not available beforehand, so it's still a net improvement.

@p-mishra1 Are you okay with that?

p-mishra1: Yes @BenjaminBossan sounds fine will work on this more, and
just curious do we have slack/forum/discord where we can talk about code design before commiting it, or is it the preferred way.

BenjaminBossan: Thanks for continuing the work.

I think ideally these discussions should come up in the PR or the corresponding issue. I thought there was consensus on the suggested approach here, I guess there was some confusion, sorry for that.

p-mishra1:
```python
    def __repr__(self) -> str:

        model = getattr(self, "model", None)
        if model:
            model = _load_model(model) # this line
            model_str = self._strip_blank(repr(model))
            model_repr = aRepr.repr(f"  model={model_str},").strip('"').strip("'")
        else:
            model_repr = None
```

```python
    def _extract_estimator_config(self) -> str:
        """Extracts estimator hyperparameters and renders them into a vertical table.

        Returns
        -------
        str:
            Markdown table of hyperparameters.
        """
        model = _load_model(self.model) # This line
        hyperparameter_dict = model.get_params(deep=True)
```

```python

    def _generate_card(self) -> ModelCard:
        .
        .
        .
        if self.model_diagram is True:
            model = _load_model(self.model) # this line
            model_plot_div = re.sub(r"\n\s+", "", str(estimator_html_repr(model)))

```
so these 3 functions will be changed like this(commented with "this line") while doing lazy loading in card class,
also the test for card __repr__ and __str__ either have to be changed to take card instance and path or the test class have to be duplicated like passing model file as param instead of estimator.
lmk @adrinjalali @BenjaminBossan which method should we prefer.

BenjaminBossan: I wonder if this is the right approach or if we should keep the `property` and just do the `_load_model` inside the property. The disadvantage of the suggested approach is that it's very error prone, as soon as new code is added that access the model, it would need to do the same thing. That's why I would prefer using `property`. @adrinjalali wdyt?

adrinjalali: how about `self.get_model()` which would return the model if `self.model` is a `str`? acceptable middle ground?

p-mishra1:
@BenjaminBossan @adrinjalali have a look at the changes, sorry for the long gap, was busy with some things.

adrinjalali (approved):
This LGTM now. but since the code has changed a bit since last review from @BenjaminBossan , I'll let him check and merge.

skops/card/_model_card.py
```diff
@@ -162,6 +165,41 @@ def metadata_from_config(config_path: Union[str, Path]) -> ModelCardData:
     return card_data
 
 
+def _load_model(model: Any) -> Any:
+    """Loads the mddel if provided a file path, if already a model instance return it
+       unmodified.
+
+    Parameters
+    ----------
+    model : pathlib.path, str, or sklearn estimator
```
BenjaminBossan, line 174:
```suggestion
    model : pathlib.Path, str, or sklearn estimator
```

skops/card/_model_card.py
```diff
@@ -263,6 +304,18 @@ def __init__(
         self._extra_sections: list[tuple[str, Any]] = []
         self.metadata = metadata or ModelCardData()
 
+    def get_model(self) -> Any:
+        """Returns sklearn estimator object if ``Path``/``str``
+            is provided.
```
BenjaminBossan, line 309:
```suggestion
        is provided.
```

skops/card/_model_card.py
```diff
@@ -162,6 +165,41 @@ def metadata_from_config(config_path: Union[str, Path]) -> ModelCardData:
     return card_data
 
 
+def _load_model(model: Any) -> Any:
+    """Loads the mddel if provided a file path, if already a model instance return it
+       unmodified.
```
BenjaminBossan, line 170:
```suggestion
    unmodified.
```

skops/card/_model_card.py
```diff
@@ -172,8 +210,11 @@ class Card:
 
     Parameters
     ----------
-    model: estimator object
-        Model that will be documented.
+    model: pathlib.path, str, or sklearn estimator object
+        ``Path``/``str`` of the model or the actual model instance that will be
+        documented. If a ``Path`` or ``str`` is provided, model will be loaded
+        on first use.
```
BenjaminBossan, line 216:
```suggestion
        documented. If a ``Path`` or ``str`` is provided, model will be loaded.
```

After the change, it will be loaded each time, so "on first use" is not correct anymore.

skops/card/_model_card.py
```diff
@@ -172,8 +210,11 @@ class Card:
...
+
```
BenjaminBossan, line 217:
```suggestion
```

skops/card/tests/test_card.py
```diff
@@ -26,6 +26,29 @@ def fit_model():
...
+    return save_file_handle, save_file
+
+
+@pytest.mark.parametrize("suffix", [".pkl", ".pickle", ".skops"])
+def test_load_model(suffix):
+    model0 = LinearRegression(n_jobs=123)
+    _, save_file = save_model_to_file(model0, suffix)
+    loaded_model_str = _load_model(save_file)
+    save_file_path = Path(save_file)
+    loaded_model_path = _load_model(save_file_path)
```
BenjaminBossan, line 45: Let's add one more case here, `_load_model(model0)`, for completeness.

skops/card/tests/test_card.py
```diff
@@ -26,6 +26,29 @@ def fit_model():
...
+
+    assert loaded_model_str.n_jobs == model0.n_jobs
+    assert loaded_model_path.n_jobs == model0.n_jobs
```
BenjaminBossan, line 48:
```suggestion
    assert loaded_model_str.n_jobs == 123
    assert loaded_model_path.n_jobs == 123
```

This test is a tiny bit more robust this way (in case that `model0` is being mutated by `_load_model` for some reason).

skops/card/tests/test_card.py
```diff
@@ -26,6 +26,29 @@ def fit_model():
...
+    assert loaded_model_path.n_jobs == loaded_model_str.n_jobs
```
BenjaminBossan, line 49: This test is not needed, right? If the two asserts above succeed, this cannot possibly fail.

skops/card/tests/test_card.py
```diff
@@ -405,6 +428,85 @@ def test_with_metadata(self, card: Card, meth):
...
+    def test_model_card_repr(self, card: Card, meth, suffix):
+        result_from_model = meth(card)
+
+        file_handle, file_name = save_model_to_file(card.model, suffix)
+
+        os.close(file_handle)
+
+        card_from_path = self.path_to_card(file_name)
+        result_from_path = meth(card_from_path)
+        print(result_from_path)
```
BenjaminBossan, line 471: Remove

skops/card/tests/test_card.py
```diff
@@ -405,6 +428,85 @@ def test_with_metadata(self, card: Card, meth):
...
+        card.add_plot(
+            roc_curve="ROC_curve.png",
+            confusion_matrix="confusion_matrix.jpg",
+        )
+        card.add_table(search_results={"split": [1, 2, 3], "score": [4, 5, 6]})
+        return card
+
+    @pytest.mark.parametrize("meth", [repr, str])
+    @pytest.mark.parametrize("suffix", [".pkl", ".skops"])
+    def test_model_card_repr(self, card: Card, meth, suffix):
```
BenjaminBossan, line 462: I'm not sure if this test is strictly necessary. Maybe it is sufficient to show that calling `repr` or `str` on a card that has a path/str for model works? The fact that it's the same output is implied by the other tests.

BenjaminBossan (changes requested):
I only have a few minor comments left, after that we should be good :+1:

skops/card/tests/test_card.py
p-mishra1: I changed it to only repr from card_from _path.

BenjaminBossan (approved):
This looks good. Thanks a lot @p-mishra1 for your work and patience.

p-mishra1:
> This looks good. Thanks a lot @p-mishra1 for your work and patience.

Thanks a lot to you guys @BenjaminBossan and @adrinjalali for guiding me and listening to my silly comments and doubts.
Also thanks to @merveenoyan who introduced me to the repo by posting about this issue on Twitter.

one request I have to @BenjaminBossan as most of the guidance and reviews were provided by you, can you provide me any feedback you have for me so that I can improve myself?

BenjaminBossan:
Hey @p-mishra1,

First of all, it's really hard to give feedback after one PR and without knowing you personally, so there is only so much I can say. But this is my personal impression:

I liked the initiative you showed. Your responses were quick and you were receptive to our feedback - even when it was conflicting sometimes ;-). As you know, this all took some time but still, you persisted, which is very good. You were willing to do some of the hard work and came up with solutions to new challenges that occurred on the way.

If there was one thing that would make the life of us maintainers easier, it is if right from the start, you would have delivered a more "complete" package, i.e. updating the docs everywhere, providing unit tests, making sure that code style and formatting are consistent with the rest of the code base. When we see that as maintainers right from the start, we're always excited.

Hopefully, this helps you a little bit. If you have feedback for us, also feel free to share.

p-mishra1:
> Hey @p-mishra1,
> 
> First of all, it's really hard to give feedback after one PR and without knowing you personally, so there is only so much I can say. But this is my personal impression:
> 
> I liked the initiative you showed. Your responses were quick and you were receptive to our feedback - even when it was conflicting sometimes ;-). As you know, this all took some time but still, you persisted, which is very good. You were willing to do some of the hard work and came up with solutions to new challenges that occurred on the way.
> 
> If there was one thing that would make the life of us maintainers easier, it is if right from the start, you would have delivered a more "complete" package, i.e. updating the docs everywhere, providing unit tests, making sure that code style and formatting are consistent with the rest of the code base. When we see that as maintainers right from the start, we're always excited.
> 
> Hopefully, this helps you a little bit. If you have feedback for us, also feel free to share.

Thank you @BenjaminBossan i appreciate you taking some time to give me feedback. I know it is quite tough for you to give feedback on the coding part .
I don't have any feedbacks from my side i really liked the way you guys handled each PR despite the noob mistakes that i did.
I would love contributing more and i will be on look out for the opportunities but if you guys feel there is some issues, feature i can contribute to i will be happy to take that up.

Issue #113, opened 2022-08-23
RFC make init create a minimal working README.md file


adrinjalali:
The inference and the widget would only work if the metadata section of the README file is present. Should we create an empty modelcard with the necessary sections for those backends to work by default when the user calls `hub_utils.init`?

BenjaminBossan:
I think since `init` already creates files, it would not hurt to create a minimal README.md. If users want to add a proper model card, it wouldn't prevent them from doing that.

jucamohedano:
Hi! I could try help with this issue. I am aware that `hub_utils.init` initializes a local repository with the model file and a configuration file that is used to generate the model card. Then we can use the `save()` method from the `Card` class to genereate a `README.md` file. Could you give me a few pointers on how to get started if you are still considering this issue? 
Thanks!

adrinjalali:
That sounds about right @jucamohedano . One simply needs to generate an empty README file. You can start a PR and we can help you there.

adrinjalali:
Closing as `hub_utils` is deprecated away.

PR #207, opened 2022-11-01
Generate minimal README.md file in repository initialization (issue #113)

jucamohedano:
This PR references issue #113

Implemented the function `_create_readme` which is called in `hub_utils.init`. It creates a minimal README.md file. The input to this function is the same as the input to `_create_config` function. 
At the moment, the function `_create_readme` only supports a `tabular` task for the widget. I haven't added the `text` task since other functions such as `metadata_from_config` in `_model_card.py` don't give support for the `text` task yet. Let me know if this can be added otherwise.

I have manually tested the initialization of a repository with the example script `plot_hf_hub.py` and it seemed to work as expected.


adrinjalali (commented):
Thanks for the PR @jucamohedano

skops/hub_utils/_hf_hub.py
```diff
@@ -213,6 +213,83 @@ def recursively_default_dict() -> MutableMapping:
     dump_json(Path(dst) / "config.json", config)
 
 
+def _create_readme(
```
adrinjalali, line 216: I don't think we need this function.

We can probably simply add such two lines to the `init`:

```py
model_card = card.Card(model, metadata=card.metadata_from_config(Path(local_repo)))
model_card.save(Path(local_repo) / "README.md")
```

skops/hub_utils/_hf_hub.py
```diff
@@ -320,6 +323,11 @@ def init(
             task=task,
             data=data,
         )
+
+        with open(model, "rb") as f:
+            model = load(f)
+            model_card = card.Card(model, metadata=card.metadata_from_config(dst))
+            model_card.save(dst / "README.md")
```
adrinjalali, line 330: thanks, could you please write a test and check if the file is created after `init`?

merveenoyan: Do you think it would be nice if there would be another test that makes sure card is overridden after user creates it manually? Or is it too much? 😅

adrinjalali: yeah that also makes sense.

jucamohedano: I added both tests. I am not experienced writing tests, so I am not sure if that's the way you would implement them. After testing each test individually, they both pass.

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -271,6 +271,23 @@ def test_init(classifier_pickle, config_json):
         )
 
 
+def test_init_modelcard_creation(classifier_pickle, config_json):
+    # create a temp directory and delete it, we just need a unique name.
+    dir_path = tempfile.mkdtemp()
+    shutil.rmtree(dir_path)
```
merveenoyan, line 277: Why do you make and remove a directory path?

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -271,6 +271,23 @@ def test_init(classifier_pickle, config_json):
...
+    version = metadata.version("scikit-learn")
+    init(
+        model=classifier_pickle,
+        requirements=[f'scikit-learn="{version}"'],
+        dst=dir_path,
+        task="tabular-classification",
+        data=iris.data,
+    )
+    _validate_folder(path=dir_path)
+    assert os.path.isfile(Path(dir_path) / "README.md")
```
merveenoyan, line 288: I wonder if we should simply add it under `test_init` to deduplicate (essentially everything before this step is same and this behavior will be a part of init anyway)? WDYT? @adrinjalali @jucamohedano

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -271,6 +271,49 @@ def test_init(classifier_pickle, config_json):
...
+    t0 = os.path.getmtime(Path(dir_path) / "README.md")
+
+    # override existent modelcard created by init
+    model = get_classifier()
+    model_card = card.Card(model, metadata=card.metadata_from_config(Path(dir_path)))
+    model_card.save(Path(dir_path) / "README.md")
+    t1 = os.path.getmtime(Path(dir_path) / "README.md")
+
+    # compare the times at which the files were last modified
+    assert t0 != t1
```
merveenoyan, line 314: I think instead of this we can render the content of model card and compare whether they're same or not, WDYT? @adrinjalali @jucamohedano

merveenoyan (commented):
Thanks a lot, I left a very general feedback 🙂

skops/hub_utils/tests/test_hf_hub.py
adrinjalali: I'm happy either way!

skops/hub_utils/tests/test_hf_hub.py
adrinjalali: because we need a directory name / path, but we don't want it to exist before init.

skops/hub_utils/tests/test_hf_hub.py
adrinjalali: That's true, this specific one can be in `test_init`.

skops/hub_utils/tests/test_hf_hub.py
merveenoyan: We want to make sure content is modified so it would be nice to have the former.

skops/hub_utils/tests/test_hf_hub.py
jucamohedano: I agree. Made the change.

skops/hub_utils/tests/test_hf_hub.py
jucamohedano: I made the change :)

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -395,7 +429,6 @@ def repo_path_for_inference():
 
 
 @pytest.mark.network
-@pytest.mark.inference
```
adrinjalali, line 398: this is probably removed by mistake

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -415,8 +448,7 @@ def test_inference(
     repo_path_for_inference,
     destination_path,
 ):
-    # test inference backend for classifier and regressor models. This test can
-    # take a lot of time and be flaky.
+    # test inference backend for classifier and regressor models.
```
adrinjalali, line 451: this change as well

adrinjalali (commented):
Thanks @jucamohedano , it's almost ready to be merged :)

skops/hub_utils/tests/test_hf_hub.py
jucamohedano: Thank you! I removed that somehow by mistake. I added it back.

skops/hub_utils/tests/test_hf_hub.py
jucamohedano: Thanks again! I added it back :)

adrinjalali (approved):

adrinjalali:
@BenjaminBossan any idea why the CI fails here? It's really odd to me.

BenjaminBossan:
> any idea why the CI fails here? It's really odd to me.

I didn't follow this PR along. Is the expectation here that the error should be caught by the `try`?

adrinjalali:
Oh I see where the issue comes from. The error is raised inside this test: `test_init_empty_model_file_warns`

So far, we have allowed people to pass an empty model file, we just raise a warning if they do so.

Now, this PR tries to actually load the model file and do things with it, which means it expects it to be a valid model file.

So, @BenjaminBossan , the question is: do we allow people to init a repo with an invalid model file? I'm leaning towards raising if the model file doesn't contain a valid model. WDYT?

BenjaminBossan:
> do we allow people to init a repo with an invalid model file? I'm leaning towards raising if the model file doesn't contain a valid model. WDYT?

I would be okay with that, don't really see a use case where you'd start out without a model.

adrinjalali:
https://github.com/skops-dev/skops/pull/214 should fix the issue.

adrinjalali:
@jucamohedano note that I've merge with latest `main`, you'd need to do a `git pull` before applying more changes to have this change applied locally on your machine.

jucamohedano:
I think that you have fixed the issue in your commit here 53119d75ec086513facd00fc73bc46219b523d61 ? @adrinjalali

adrinjalali:
@jucamohedano if you can't debug the issue, let us know. We can let another person continue this work :)

jucamohedano:
So the problem comes due to the fact that some tests are not using `pickle`, and instead use `skops.io` to dump and load models. So in `hub_utils.init` I am assuming that the model passed is a pickle file. Should I account for both formats or should I get rid of the pickle format and only accept the skops format in `hub_utils.init` ?

BenjaminBossan:
@jucamohedano Since #242 is merged, I think there is no blocker for this PR. Do you still intend on working on it?

jucamohedano:
@BenjaminBossan No, I was waiting for #242 to be merged. I have resolved some small conflicts in my las commit.

BenjaminBossan:
@jucamohedano Thanks for continuing to work on this. The tests are failing with:

> NameError: name 'load' is not defined

That's probably easy to fix, I assume after merging an import statement was changed.

jucamohedano:
I actually had to modified the functionality I added to `hub_utils.init` so that it checks for file format when loading the model file, previously I assumed it was just a pickle file, but now I also checked for the skops format.

adrinjalali:
Linter is failing here. You can consider enabling pre-commit hooks: https://github.com/skops-dev/skops/blob/main/CONTRIBUTING.rst#using-condamamba

jucamohedano:
Sorry! I did a clean installation, and forgot to install pre-commits hooks. Thank you for reminding.

skops/hub_utils/_hf_hub.py
```diff
@@ -427,6 +430,14 @@ def init(
             model_format=model_format,
             use_intelex=use_intelex,
         )
+
+        if model_format == "pickle":
+            with open(model, "rb") as f:
+                model = pikle_load(f)
```
adrinjalali, line 436: please write a test to cover these lines.

adrinjalali (commented):
Other than a missing test, looks good.

skops/hub_utils/_hf_hub.py
jucamohedano: Initally, the test was written for those lines (without the if-statement) in function `test_override_init_modelcard`. Now we are accounting for the possibility of the `skops` format. However, I believe that both formats are already tested since the pytest fixture for `config_json` in `test_hf_hub.py` provides both formats in its `params` argument. Please correct me if I'm wrong.

adrinjalali:
I wanted to submit an empty commit to double check the codecov, but you're using your `main` branch to submit the PR therefore I can't do that, you should always create a branch for each contribution on your repo.

Running the tests locally, I get the same result that those lines are not tested:

```
---------- coverage: platform linux, python 3.10.6-final-0 -----------
Name                              Stmts   Miss  Cover   Missing
---------------------------------------------------------------
skops/__init__.py                     8      1    88%   31
skops/card/__init__.py                2      0   100%
skops/card/_model_card.py           377      0   100%
skops/card/_templates.py              7      0   100%
skops/cli/__init__.py                 0      0   100%
skops/cli/_convert.py                39      0   100%
skops/cli/_utils.py                   8      2    75%   11, 13
skops/cli/entrypoint.py              12      0   100%
skops/hub_utils/__init__.py           2      0   100%
skops/hub_utils/_hf_hub.py          182     31    83%   435-436, 438, 586-597, 696-712, 752-775
skops/hub_utils/tests/common.py       1      0   100%
skops/io/__init__.py                  2      0   100%
skops/io/_audit.py                  110     12    89%   183, 220, 239, 270, 275-282, 300-305, 311
skops/io/_general.py                230      5    98%   36, 172, 216, 469, 472
skops/io/_numpy.py                  105      3    97%   43, 74, 101
skops/io/_persist.py                 66      1    98%   207
skops/io/_scipy.py                   29      1    97%   46
skops/io/_sklearn.py                 90      9    90%   10-11, 77, 79, 84, 118, 124-128
skops/io/_trusted_types.py            5      0   100%
skops/io/_utils.py                   75      3    96%   16, 50, 152
skops/io/exceptions.py                6      0   100%
skops/io/tests/_utils.py            111      4    96%   33, 37, 109, 134
skops/utils/__init__.py               0      0   100%
---------------------------------------------------------------
TOTAL                              1467     72    95%
```

Could you please investigate?

jucamohedano:
Sorry, I know I should made a new branch for each feature. I was excited to contribute to the project, and completely forgot about doing it.

I have fixed the tests. I ran them locally, and the lines stated above within the `init` function aren't missing anymore.

adrinjalali:
Doc build is raising this issue which is relevant: 


```
Unexpected failing examples:
/home/docs/checkouts/readthedocs.org/user_builds/skops/checkouts/207/examples/plot_tabular_regression.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/skops/checkouts/207/examples/plot_tabular_regression.py", line 73, in <module>
    hub_utils.init(
  File "/home/docs/checkouts/readthedocs.org/user_builds/skops/checkouts/207/skops/hub_utils/_hf_hub.py", line 437, in init
    model = pikle_load(f)
_pickle.UnpicklingError: A load persistent id instruction was encountered,
but no persistent_load function was specified.
```

jucamohedano:
Is there a way to check if the model file that's being loaded is in the skops persistent format? Because I could catch the error that you highlighted and in consequence load the model file using `skops.io.load`, but I want to know if there's a cleaner way of doing it than catching the error.

adrinjalali:
after https://github.com/skops-dev/skops/pull/272 , the `model_format` gives you that information

jucamohedano:
Great, thanks! I have updated the logic in the code to account for model format and the extension of the file if `model_format` is `auto`. However, I believe that the error that you highlighted will still occur because the example `plot_tabular_regression.py` creates a skops persistent model which is saved with the `.pkl` extension. This makes the `_hf_hub.init` function fail when resolving the type of file to read. This can be easily fix by setting the `model_format` argument to `skops` or correcting the suffix given.

.github/workflows/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,16 +1,16 @@
 <!--
 Thanks for contributing a pull request! Please ensure you have taken a look at
 the contribution guidelines: https://github.com/skops-dev/skops/blob/main/CONTRIBUTING.rst
-This guideline contains crucial information for your PR to be merged, such as setting up
```
adrinjalali, line 4: not sure what happened here, but we've moved this file, you shouldn't be having any changes here.

adrinjalali (commented):
You also have an error in the doc build, related to loading a skops file.

jucamohedano (commented):
Besides fixing the argument type in `examples/plot_model_card.py`, other files were included in this commit as a result from the merge with the upstream skops repository.

jucamohedano:
Hi, sorry that this PR is going for too long.
I was looking at the output from the doc build, in particular this one
```
Unexpected failing examples:
/home/docs/checkouts/readthedocs.org/user_builds/skops/checkouts/207/examples/plot_intelex.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/skops/checkouts/207/examples/plot_intelex.py", line 156, in <module>
    hub_utils.init(
  File "/home/docs/checkouts/readthedocs.org/user_builds/skops/checkouts/207/skops/hub_utils/_hf_hub.py", line 444, in init
    model = pikle_load(f)
EOFError: Ran out of input
```
This error happens because `examples/plot_intelex.py` creates and opens temporary files that are also being opened in `hub_utils.init` function through the use of the `model` parameter. Maybe we can create those temporary files without using the wrapper that takes care of closing and deleting them, and instead we delete them explicitly. There might be a better solution that my knowledge can't reach.

BenjaminBossan:
> This error happens because examples/plot_intelex.py creates and opens temporary files that are also being opened in hub_utils.init function through the use of the model parameter.

Hmm, if that's really the reason, doesn't it mean that this code is breaking backwards compatibility? I'd much prefer if it didn't, e.g. by making the creation of the README.md opt-in. WDYT @adrinjalali 

Also, from a quick glance at the implementation, I think it can be greatly simplified. It is possible to instantiate the `Card` class with a path to the model file (perhaps that was added after this PR started), no need to load the model inside of `hub_utils.init`.

adrinjalali:
I'm quite confused by this. This code is a minimal reproducible for the issue:

```py
from tempfile import NamedTemporaryFile
import pickle

with NamedTemporaryFile(mode="bw", prefix="stock-", suffix=".pkl") as fp:
    pickle.dump("test", file=fp)
    print(fp.name)
    with open(fp.name, "rb") as f:
        pickle.load(f)
```

Now if we add a `seek(0)` for `fp`, it works:

```py
from tempfile import NamedTemporaryFile
import pickle

with NamedTemporaryFile(mode="bw", prefix="stock-", suffix=".pkl") as fp:
    pickle.dump("test", file=fp)
    print(fp.name)
    fp.seek(0)
    with open(fp.name, "rb") as f:
        pickle.load(f)
```

To me, that makes 0 sense. Does it make sense to you @BenjaminBossan ?

BenjaminBossan:
> To me, that makes 0 sense. Does it make sense to you @BenjaminBossan ?

I can't say I would have predicted it, but it doesn't seem too surprising. They way I visualize this to myself (probably technically quite incorrect) is that when I open a file, there is a "cursor" at position 0 for the file object. After writing to the file, it is at byte X, where X is the file size. When I try to read from the same file object, since the "cursor" is at the end, there is nothing to read. Therefore, `seek(0)` (putting the cursor to position 0) solves the issue. Normally, a file object is only read from/written to once, so it doesn't become an issue, but here, it is used twice.

For this PR specifically, as mentioned earlier, I would prefer if the default behavior is not changed, i.e. creating a model card should be opt in. IIUC this would solve the issue. When it comes to the intelex example, if it would be changed to use temporary files the same way we do in other examples (i.e. no context manager), it should solve issue too, but context manager has the advantage of immediately cleaning up the temp file.

adrinjalali:
Reading the [python docs](https://docs.python.org/3/library/tempfile.html#tempfile.NamedTemporaryFile):

> This function operates exactly as [TemporaryFile()](https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile) does, except that the file is guaranteed to have a visible name in the file system (on Unix, the directory entry is not unlinked). That name can be retrieved from the name attribute of the returned file-like object. Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows). If delete is true (the default), the file is deleted as soon as it is closed. The returned object is always a file-like object whose file attribute is the underlying true file object. This file-like object can be used in a [with](https://docs.python.org/3/reference/compound_stmts.html#with) statement, just like a normal file.

When passing the file to `init`, we should expect it to be read, therefore it should be something which can be opened, and therefore I think we need to fix the intelex example accordingly.

BenjaminBossan:
> I think the issue we have here is a python bug

Not sure about this :) At the very least, it's not specific to `NamedTemporaryFile`:

```python
with open('foo', 'w') as f:
    f.write("hello")
    # f.seek(0) fixes it
    with open(f.name, 'r') as g:
        bla = g.read()
        print(bla)  # prints nothing
```

> therefore I think we need to fix the intelex example accordingly

Yes, we can certainly do that, but there could still be code out there that would break in the same fashion.

adrinjalali:
> but there could still be code out there that would break in the same fashion.

We can be specific in our docs that we do read the file, therefore it should be readable, and if they pass something we can't read, it's their issue.

BenjaminBossan:
> would it be okay here to document that `init` _reads_ the model file and therefore it should be readable?

In general, yes, if we document what we expect, I think it's fine. Here, specifically, is it really the problem that the file needs to be readable? Isn't the issue that the fd is stateful and has to be at position 0? Unless that's what is implied by saying it is "readable".

adrinjalali:
@BenjaminBossan I'm not sure if we should do a `seek(0)` on the file before reading it (if we do, we should do it everywhere we read a file I think).

BenjaminBossan:
> I'm not sure if we should do a `seek(0)` on the file before reading it (if we do, we should do it everywhere we read a file I think).

No, we shouldn't seek, I just meant that having the file be readable is not sufficient. Anyway, I think it practice this is not going to be a very big deal, the pattern used in the failing example is not very common, we shouldn't stall this PR because of that.

adrinjalali:
@jucamohedano would you mind updating the example to fix the issue? we can then proceed to merge.

jucamohedano:
One thing to add. In my last commit I have added a fix to the file reading logic in `init` so that the `plot_tabular_regression.py` example runs without errors,  otherwise it throws an untrusted type error for `numpy.float64`. However, I'm questioning whether the `trusted` arg should be added where I have added it in the last commit, or maybe it should be added if we know for sure that the `model_format=='skops'`. If we go with the latter, then we have to modify the example to specify that the model format is `skops`. Please let me know if my explanation is clear enough.

BenjaminBossan:
> However, I'm questioning whether the `trusted` arg should be added where I have added it in the last commit, or maybe it should be added if we know for sure that the `model_format=='skops'`. If we go with the latter, then we have to modify the example to specify that the model format is `skops`. Please let me know if my explanation is clear enough.

As a user, I would be surprised to find out that `model_format="skops"` does not auto-trust all types but `model_format="auto"` does. It should be one or the other. Presumably, we need to always trust everything, otherwise running `init` would not work for certain models if it uses the skops format. At the same time, trusting all types is something we want to discourage, after all, this is not much better than using pickle. We could add a `trusted` argument to `init`, but the function is already quite big as is and at first glance, it's not apparent why it would need that argument.

Overall, this has become much more complicated than I initially expected. @adrinjalali what would you suggest?

jucamohedano:
I agree with the suggestion from @adrinjalali, otherwise it looks like we would have to hard code the `model_format` arg. What do you think about his suggestion @BenjaminBossan ?

BenjaminBossan:
Yes, that might be the better solution. I'm not 100% sure if there might not be new issues with that approach, but it's worth trying.

jucamohedano:
Maybe we run into some issue, but we'll see. Shall we keep working on this PR or a new open one?

BenjaminBossan:
> Maybe we run into some issue, but we'll see. Shall we keep working on this PR or a new open one?

Do whatever you feel more comfortable with.

jucamohedano:
I have written an implementation to test in my local setup where I have introduced a new argument to the `_hf_hub_.init` function. the new argument is a model object. However, as expected many tests fail. Maybe we could use the model object if it is given, and if not we could load the model. If the latter raises an error, we just let the user know about this to fix incompatibilties with the saved model object the user's trying to load. WDYT?

I will remove unnecessary changes from the diff in future commits!

BenjaminBossan:
> I have written an implementation to test in my local setup where I have introduced a new argument to the `_hf_hub_.init` function. the new argument is a model object. However, as expected many tests fail. Maybe we could use the model object if it is given, and if not we could load the model. If the latter raises an error, we just let the user know about this to fix incompatibilties with the saved model object the user's trying to load. WDYT?

I would prefer for the change to not be backwards incompatible, i.e. all existing tests should still pass. I have a suggestion, not sure if there is something I missed that would invalidate it, let me know:

The `model` argument to `init` will, in the future, also allow an sklearn estimator to be passed. If `model` is a `str` or `Path`, nothing changes, everything stays the same. If `model` is an estimator, use it to create a README.md (if there isn't already one). _Also_, save the model in the destination directory using a generic name, like `"model.pkl"` or `"model.skops"` (depending on the `model_format` argument). WDYT about that?

adrinjalali:
> The model argument to init will, in the future, also allow an sklearn estimator to be passed. If model is a str or Path, nothing changes, everything stays the same. If model is an estimator, use it to create a README.md (if there isn't already one)

That doesn't sound too bad to me, at some point we might want to deprecate passing a path to the model maybe.

jucamohedano:
I totally agree with the implementation and that's why I decided to go ahead and implement it.
Docs build is currently failing. I will go over that later today and fix it!

jucamohedano:
Hi, I'm not sure what to do about the failing example in the docs build. Since `hub_utils.init` is creating a README.md file by default when creating the repository, `hub_utils.add_files` fails to execute because an existing README.md is already there. I would assume it's safe to overwrite it? 
This is the error from the docs build failed:
```
Unexpected failing examples:
/home/docs/checkouts/readthedocs.org/user_builds/skops/checkouts/207/examples/plot_california_housing.py failed leaving traceback:
Traceback (most recent call last):
  File "/home/docs/checkouts/readthedocs.org/user_builds/skops/checkouts/207/examples/plot_california_housing.py", line 1852, in <module>
    hub_utils.add_files(
  File "/home/docs/checkouts/readthedocs.org/user_builds/skops/checkouts/207/skops/hub_utils/_hf_hub.py", line 531, in add_files
    raise FileExistsError(msg)
FileExistsError: File 'README.md' already found at '/tmp/tmpwmea7ezv'.
```

BenjaminBossan:
Maybe I misunderstand your code, but it seems like it doesn't do what we discussed, resulting in the error? IIUC, `init` will always create a README.md. My suggestion was 

> If model is a str or Path, nothing changes, everything stays the same. If model is an estimator, use it to create a README.md (if there isn't already one).

jucamohedano:
Thank you for pointing that out @BenjaminBossan, I realised the code wasn't doing what was discussed. I reverted back to the behaviour of creating a README if model is a str/Path, otherwise handle it if model is an estimator object. I believe the code should be doing that now. However, docs still fail to build on the example `plot_california_housing.py`. I actually think that the example needs to be updated to account for the creation of a README at the initialization of the repository, and avoid adding the README file manually to the repository. WDT?

Mistakenly closed the PR while scrolling.

BenjaminBossan:
Thanks for the updates @jucamohedano.

This is not a full on review, as I think there are still some bigger changes needed. Most notably, I feel like `init` has become too complicated, it is hard to track what is going on. Let's try to simplify it. A good start is to take a look at [`_load_model`](https://github.com/skops-dev/skops/blob/b0908b3c2d736c6397bff1251bd794d318a491b6/skops/card/_model_card.py#L334), which I believe you can re-use for this purpose.

> I actually think that the example needs to be updated to account for the creation of a README at the initialization of the repository, and avoid adding the README file manually to the repository. WDT?

Maybe I misunderstand, but shouldn't it be the other way round? What I mean is that the README.md created by the example is the one that contains all the important information and should thus take precedence over the README.md that was automatically created.

jucamohedano:
Hey! Thanks for the suggestion about `_load_model` because it can simplify the code as long as we can trust the types within the model file to be loaded.

> Maybe I misunderstand, but shouldn't it be the other way round? What I mean is that the README.md created by the example is the one that contains all the important information and should thus take precedence over the README.md that was automatically created.

Thanks for making it clear. The call to `hub_utils.add_files` in the example is the one that throws the error about an already existing README.md in the repository. We could address this problem in `hub_utils.add_files` to account for the existing file created by the call to `hub_utils.init`.

skops/hub_utils/_hf_hub.py
```diff
@@ -415,19 +418,64 @@ def init(
...
+                data=data,
+                model_format=model_format,
+                use_intelex=use_intelex,
+            )
+
+            # load model from file
+            model = _load_model(model, trusted=True)
+            model_card = card.Card(model, metadata=card.metadata_from_config(dst))
+            model_card.save(dst / "README.md")
+        elif isinstance(model, BaseEstimator):
```
BenjaminBossan, line 443: I think we need no check for `BaseEstimator` here, so this can just be `else` and the error below can be removed. There could be valid models here that don't inherit from `BaseEstimator`. It is the user's responsibility to provide an sklearn-compatible model.

skops/hub_utils/_hf_hub.py
```diff
@@ -415,19 +418,64 @@ def init(
...
+            # if it is a model object and its format is set to auto, choose skops by default
+            if model_format == "auto":
+                model_format = "skops"
+            elif model_format in ["pkl", "pickle", "joblib"]:
+                model_format = "pickle"
```
BenjaminBossan, line 448: We need an `else` clause below because if the user passes `model_format="skosp"` or another typo, that would just be accepted as is, even though it is invalid. Then further below, the model would not be saved without any indication that something went wrong.

Essentially, the model format checking logic should be the same as inside `_create_config`. And since checking the model format is new performed in `init`, we should also be able to remove it from `_create_config` completely.

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -303,6 +320,56 @@ def test_init(classifier, config_json):
...
+
+
+def test_override_init_modelcard(classifiers, config_json):
+    # create a temp directory and delete it, we just need a unique name.
+    dir_path = tempfile.mkdtemp()
+    shutil.rmtree(dir_path)
+
+    version = metadata.version("scikit-learn")
+    _, model_format = config_json
+    # joblib type falls unders auto format, explicityly set to auto
```
BenjaminBossan, line 342:
```suggestion
    # joblib type falls under auto format, explicitly set to auto
```

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -303,6 +320,56 @@ def test_init(classifier, config_json):
...
+    # because we can't repeat key "auto" in CONFIG dict
```
BenjaminBossan, line 343: Sorry, I don't understand that, could you please explain further?

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -303,6 +320,56 @@ def test_init(classifier, config_json):
...
+    _validate_folder(path=dir_path)
+
+    # inital card does not have a license set
+    with pytest.raises(
+        AttributeError, match="'CardData' object has no attribute 'license'"
+    ):
+        model_card = RepoCard.load(Path(dir_path) / "README.md")
+        model_card.data.license
+
+    # override existent modelcard created by init with license attribute
```
BenjaminBossan, line 364: For my better understanding, is this testing some new behavior added by this PR or some general behavior?

BenjaminBossan (changes requested):
Thanks for providing updates to bring this PR closer to be merged. I have made a few comments and had some questions for clarification, please take a look.

> We could address this problem in hub_utils.add_files to account for the existing file created by the call to hub_utils.init.

`add_files` already has an option, `exist_ok`, which can be set to `True` to allow overriding.

skops/hub_utils/_hf_hub.py
jucamohedano: Okay, got it!

skops/hub_utils/_hf_hub.py
jucamohedano: If so, I can remove that logic from `_create_config` and have it just in `init`. I actually started this PR by looking at `_create_config`. Let me know if you would like to make this change now.

skops/hub_utils/tests/test_hf_hub.py
jucamohedano: Sorry, that comment isn't self-explainable. The dictionary `CONFIG` in `test_hf_hub` contains 3 types of models. I wanted to test for a model with name `model.joblib`, so I added that model to the dictionary. If I recall correctly, adding the key:value pair for the joblib model made other tests fail. But I could fix this by changing the `joblib` model type `auto` and still have the name `model.joblib` to test for.

skops/hub_utils/_hf_hub.py
BenjaminBossan: Yes, I think that change should be good.

skops/hub_utils/tests/test_hf_hub.py
BenjaminBossan: Hmm, I guess it depends on why the other tests fail. If this uncovers a bug with existing code, that would be good to know. Otherwise, I think it's okay to have a separate test for joblib and not changing `CONFIG`.

skops/hub_utils/tests/test_hf_hub.py
BenjaminBossan: > we can make this test simpler by checking that the README.md exists after the call to init and that's it?

Yes, that sounds like it is sufficient as a test.

adrinjalali:
We can close this since we've deprecated `hub_utils` away.

PR #209, opened 2022-11-04
ENH Implement LoadContext to handle multiple instances

### LINKED ISSUE #206: Single object referenced in multiple places doesn't load as single instance
Raising issue to track the issue noticed in #195.

We currently have [this xfail test to show the issue.](https://github.com/skops-dev/skops/blob/main/skops/io/tests/test_persist.py#L873-L885)

As noted by Adrin, currently, if we had code like:
```
a = A()
f = a.f
g = a.g
c = C(f, g)
```
And we tried to persist `C`, `f` and `g` would be bound to different instances of `A` after loading.

There has been some discussion in #195 already about ways to solve this, and it would likely require a way to track the `LoadState` during object load.

E-Aho:
Fixes #206.

PR to add ability to persist a single instance that is referenced in multiple places.

Notes:
1) Currently, this *does* save the state of the object multiple times, but only initialises once. This is done due to the issue discussed in #195 around not knowing which instance will be loaded first, particularly in complex cases with interwoven dependencies.

2) ~~At the moment, I've only added `__id__` in the places needed to solve the xfail test, but if others are happy with how this looks I will expand this to other `get_state_*` methods :)~~ This is now global, done at the high level `get_state` and `get_instance` functions so works for all types.

----

Regarding 1:  I looked into any ways to store DAGs in JSON and bumped into [a relatively new addition to JSON syntax, JSON-LD](https://www.w3.org/TR/json-ld/), which also happens to [have a Python implementation](https://github.com/digitalbazaar/pyld). I feel like if we want to only save these objects once, we would need some kind of DAG implementation to hold dependencies. This, however, would need a major rework and feels like it might be overkill for the time being, but it's worth holding in mind.

### DIFF
--- a/skops/io/_dispatch.py
+++ b/skops/io/_dispatch.py
@@ -13,6 +13,7 @@ def get_instance(state, load_state: LoadState):
         return json.loads(state["content"])
 
     saved_id = state.get("__id__")
+
     if saved_id and saved_id in load_state.memo:
         # an instance has already been loaded, just return the loaded instance
         return load_state.get_instance(saved_id)
--- a/skops/io/_dispatch.py
+++ b/skops/io/_dispatch.py
@@ -2,13 +2,20 @@ from __future__ import annotations
 
 import json
 
+from skops.io._utils import LoadState
+
 GET_INSTANCE_MAPPING = {}  # type: ignore
 
 
-def get_instance(state, src):
+def get_instance(state, load_state: LoadState):
     """Create instance based on the state, using json if possible"""
     if state.get("is_json"):
         return json.loads(state["content"])
+
+    saved_id = state.get("__id__")
+    if saved_id and saved_id in load_state.memo:
+        # an instance has already been loaded, just return the loaded instance
+        return load_state.get_instance(saved_id)
 
     try:
         get_instance_func = GET_INSTANCE_MAPPING[state["__loader__"]]
@@ -17,4 +24,11 @@ def get_instance(state, src):
         raise TypeError(
             f" Can't find loader {state['__loader__']} for type {type_name}."
         )
-    return get_instance_func(state, src)
+
+    loaded_obj = get_instance_func(state, load_state)
+
+    # hold reference to obj in case same instance encountered again in save state
+    if saved_id:
+        load_state.memoize(loaded_obj, saved_id)
+
+    return loaded_obj
--- a/skops/io/_dispatch.py
+++ b/skops/io/_dispatch.py
@@ -2,18 +2,18 @@ from __future__ import annotations
 
 import json
 
-from skops.io._utils import LoadState
+from skops.io._utils import LoadContext
 
 GET_INSTANCE_MAPPING = {}  # type: ignore
 
 
-def get_instance(state, load_state: LoadState):
+def get_instance(state, load_context: LoadContext):
     """Create instance based on the state, using json if possible"""
 
     saved_id = state.get("__id__")
-    if saved_id in load_state.memo:
+    if saved_id in load_context.memo:
         # an instance has already been loaded, just return the loaded instance
-        return load_state.get_instance(saved_id)
+        return load_context.get_instance(saved_id)
 
     if state.get("is_json"):
         loaded_obj = json.loads(state["content"])
@@ -26,10 +26,10 @@ def get_instance(state, load_state: LoadState):
                 f" Can't find loader {state['__loader__']} for type {type_name}."
             )
 
-        loaded_obj = get_instance_func(state, load_state)
+        loaded_obj = get_instance_func(state, load_context)
 
     # hold reference to obj in case same instance encountered again in save state
     if saved_id:
-        load_state.memoize(loaded_obj, saved_id)
+        load_context.memoize(loaded_obj, saved_id)
 
     return loaded_obj
--- a/skops/io/_dispatch.py
+++ b/skops/io/_dispatch.py
@@ -11,7 +11,7 @@ def get_instance(state, load_state: LoadState):
     """Create instance based on the state, using json if possible"""
 
     saved_id = state.get("__id__")
-    if saved_id and saved_id in load_state.memo:
+    if saved_id in load_state.memo:
         # an instance has already been loaded, just return the loaded instance
         return load_state.get_instance(saved_id)
 
--- a/skops/io/_dispatch.py
+++ b/skops/io/_dispatch.py
@@ -9,24 +9,24 @@ GET_INSTANCE_MAPPING = {}  # type: ignore
 
 def get_instance(state, load_state: LoadState):
     """Create instance based on the state, using json if possible"""
-    if state.get("is_json"):
-        return json.loads(state["content"])
 
     saved_id = state.get("__id__")
-
     if saved_id and saved_id in load_state.memo:
         # an instance has already been loaded, just return the loaded instance
         return load_state.get_instance(saved_id)
 
-    try:
-        get_instance_func = GET_INSTANCE_MAPPING[state["__loader__"]]
-    except KeyError:
-        type_name = f"{state['__module__']}.{state['__class__']}"
-        raise TypeError(
-            f" Can't find loader {state['__loader__']} for type {type_name}."
-        )
+    if state.get("is_json"):
+        loaded_obj = json.loads(state["content"])
+    else:
+        try:
+            get_instance_func = GET_INSTANCE_MAPPING[state["__loader__"]]
+        except KeyError:
+            type_name = f"{state['__module__']}.{state['__class__']}"
+            raise TypeError(
+                f" Can't find loader {state['__loader__']} for type {type_name}."
+            )
 
-    loaded_obj = get_instance_func(state, load_state)
+        loaded_obj = get_instance_func(state, load_state)
 
     # hold reference to obj in case same instance encountered again in save state
     if saved_id:
--- a/skops/io/_utils.py
+++ b/skops/io/_utils.py
@@ -94,7 +94,7 @@ def persist_id(func):
     Intended to be used as a decorator.
 
     NB: Not all get_state functions should include ids. Ephemeral objects
-    have their IDs reused, and so storing some objects (like dicts, lists, arrays
+    have their IDs reused, and so storing some objects (some dicts, lists, arrays
     etc.) can cause problems.
     """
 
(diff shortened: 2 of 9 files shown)


BenjaminBossan:
I think this looks pretty good, like the simplest way to solve the problem.

> if others are happy with how this looks I will expand this to other get_state_* methods

This would mean that we need to add something like this everywhere, right?

```python
def get_instance_foo(state, load_state):
    saved_id = state.get("__id__")
    if saved_id and saved_id in load_state.memo:
        # an instance has already been loaded, just return the loaded instance
        return load_state.get_instance(saved_id)
    ...
    if saved_id:
        load_state.memoize(loaded_obj, saved_id)
    return loaded_obj
```

So is it something we want to move into a decorator to avoid the boilerplate?

> 1. Currently, this _does_ save the state of the object multiple times

Let's leave it like this for now and bother about the problem later.

> I looked into any ways to store DAGs in JSON and bumped into [a relatively new addition to JSON syntax, JSON-LD](https://www.w3.org/TR/json-ld/),

I only took a glance but I hope we can manage to solve it without such a solution.

E-Aho:
> This would mean that we need to add something like this everywhere, right?

Thankfully no, because of the way I structured it, the memo check is at the top level `get_instance` function, before it gets sent to the specific `get_instance_foo` for whatever type that is, so it will be able to handle any memoized things with minimal boilerplate.

We could switch to a decorator pattern, but I felt it would be easier to just pass `__id__` as one of the state parameters for any types we want to persist as a singular instance. So, to implement elsewhere, you'd just need to add `res["__id__"] = id(obj)` in `get_state_foo`.

> I only took a glance but I hope we can manage to solve it without such a solution.

I hope there's a nice solution I've missed too, but it feels like once you start getting into edge cases with nested dependencies and such, you do need some kind of DAG implementation...

BenjaminBossan:
> Thankfully no, because of the way I structured it, the memo check is at the top level `get_instance` function

Ah yes, nice, I missed that.

> you'd just need to add `res["__id__"] = id(obj)` in `get_state_foo`.

Okay, this might be something we would want to move out of the `get_state_foo` functions, since it's easy to forget and not statically checked.

> I hope there's a nice solution I've missed too, but it feels like once you start getting into edge cases with nested dependencies and such, you do need some kind of DAG implementation...

Hmm, not sure, let's see. Maybe we can learn how pickle implements it and copy that approach.

E-Aho:
I've added in a decorator that can be used to wrap get_state functions that we want to be able to persist as singletons.

I initially tried having it in the high level `get_state` function, but ran into issues I think were being caused with IDs being garbage collected and reused during persist for things which were not equivalent.

BenjaminBossan:
> I initially tried having it in the high level `get_state` function, but ran into issues I think were being caused with IDs being garbage collected and reused during persist for things which were not equivalent.

Interesting, do you still have the code for this? It's not clear to me why moving to a decorator solves this.

Assuming that code looks similar to what you have now:

```python
        result = func(obj, save_state)
        result["__id__"] = id(obj)
```

did you try if this works instead?

```python
        __id__ = save_state.memoize(obj)
        result = func(obj, save_state)
        result["__id__"] = __id__
```

BenjaminBossan:
> My intuition was that CPython might be garbage collecting these, then reusing the memory address, giving the same **id**.

I think this is the most likely explanation, which is why I suggested to memoize the object first. This should prevent gc and make the `id` stable.

E-Aho:
I suppose either method would work, but memoizing every object would mean basically never garbage collecting, 

> > My intuition was that CPython might be garbage collecting these, then reusing the memory address, giving the same **id**.
> 
> I think this is the most likely explanation, which is why I suggested to memoize the object first. This should prevent gc and make the `id` stable.

That should work, although it would also mean holding a lot more things in memory at all times, which might give a minor performance hit. If we don't need to actually hold the object, just get a unique `id`, it might be easier to just use a UUID, but I'm happy to just memoize the objects as well.

BenjaminBossan:
> I suppose either method would work, but memoizing every object would mean basically never garbage collecting,

The memory is cleared once dumping is over, so gc should work after that. But it's true that during the dumping process, we could potentially see a memory spike, though I assume that most objects are held in memory anyway. If I'm not missing something, It should only really matter for attributes that are generated on the fly and require a lot of memory.

E-Aho:
> > I suppose either method would work, but memoizing every object would mean basically never garbage collecting,
> 
> The memory is cleared once dumping is over, so gc should work after that. But it's true that during the dumping process, we could potentially see a memory spike, though I assume that most objects are held in memory anyway. If I'm not missing something, It should only really matter for attributes that are generated on the fly and require a lot of memory.

No that's true, I'm overthinking it and these on-the-fly objects being held in memory should be a negligable memory hit.

I'll rework it this eve and check if it helps solve the flaky tests

E-Aho:
Memoizing seems to have solved the problem without needing any UUID rework, good call @BenjaminBossan!

I've added a test that checks it works for a few different object types, let me know if there's anything else you think would be worth testing!

E-Aho:
RE: CodeCov, as far as I can tell from the report, this PR is at 100% coverage, and just failing on an "indirect coverage change", so I think the above failing is just noise, but let me know if this isn't the case

skops/io/tests/test_persist.py
```diff
@@ -96,16 +96,16 @@ def wrapper(obj, save_state):
...
+        assert list(signature.parameters.keys()) == ["state", "load_state"]
 
         @wraps(func)
-        def wrapper(state, src):
+        def wrapper(state, load_state):
             assert "__class__" in state
             assert "__module__" in state
             assert "__loader__" in state
-            assert isinstance(src, ZipFile)
+            assert isinstance(load_state, LoadState)
```
BenjaminBossan, line 106: Can we adjust the `debug_get_state` wrapper to check that each result contains an `"__id__"` key?

skops/io/_dispatch.py
```diff
@@ -2,19 +2,34 @@
...
+
 GET_INSTANCE_MAPPING = {}  # type: ignore
 
 
-def get_instance(state, src):
+def get_instance(state, load_state: LoadState):
     """Create instance based on the state, using json if possible"""
+
+    saved_id = state.get("__id__")
+    if saved_id and saved_id in load_state.memo:
```
BenjaminBossan, line 14: I wonder if we even need to check if the `"__id__"` key exists? I see that its presence is not strictly necessary for loading to work. On the other hand, will it ever be the case that the key doesn't exist if the `state` was created with skops? (Okay, if someone already used `dump` with the current main branch, but I think that's negligible?)

Overall, I'm fine either way, just considering if we can slightly simplify the code.

BenjaminBossan (commented):
Looks very good to me, I just have some minor comments/questions.

Regarding the naming, it's a bit unfortunate that we now have `state` and `LoadState`. When adding `SaveState`, I didn't think about this issue. Maybe we can find a better name (for both for consistency)?

skops/io/_dispatch.py
E-Aho: I wanted to include that for backwards compatibility mostly, happy to remove these checks if we don't need backwards compatibility for models saved with 0.1 & 0.2 versions

E-Aho:
On the names, how do you feel about changing them to `SaveContext` and `LoadContext` @BenjaminBossan?

skops/io/_dispatch.py
BenjaminBossan: We added persistence after the 0.2 release, so it's not officially released at all. Therefore, I think we don't need to worry about it.

E-Aho: Oh, I see, that makes sense. I can definitely remove this check then 👍

skops/io/tests/test_persist.py
E-Aho: It looks like that wrapper wraps the `GET_STATE_DISPATCH_FUNCTIONS` and not the high level `get_state` function. 

Given __id__ doesn't get returned from the individual dispatch funcs, it's set in the high level one, I think that check isn't needed.

I've added a check in the`get_instance` wrapper though, so __id__ is being checked during persist

BenjaminBossan: Yes, makes sense, thanks

BenjaminBossan (approved):
Thanks a lot, fantastic work.

@adrinjalali would you please also give this a look?

E-Aho:
@BenjaminBossan after merging in your change from main, a few of the `_hub_util` tests are failing, but on my machine these tests are fine for this branch...

Could you rerun some of the CI to see if this is an actual bug or just a single test being flaky in an odd way? I've given it a quick look and didn't see any immediate changes that would've broken these specific `_hub_util` tests

E-Aho:
When I merged in main, I got a strange error in `_hub_utils` across all the tests that I think was just it being flaky (it was fine on my machine), I've bumped the CI to check it's fine

skops/io/tests/test_persist.py
```diff
@@ -785,11 +787,13 @@ def test_loads_from_str():
 
 
 def test_get_instance_unknown_type_error_msg():
-    state = get_state(("hi", [123]), None)
```
adrinjalali, line 788: why do we need to remove `None`? We should be able to persist `None`.

skops/io/_utils.py
```diff
@@ -122,22 +122,50 @@ def clear_memo(self) -> None:
...
+class LoadContext:
+    """Context required for loading an object
+
+    This context is passed to each ``get_instance_*`` function.
+
+    Parameters
+    ----------
+    src: zipfile.ZipFile
+        The zip file the target object is saved in
+
```
adrinjalali, line 135: empty line could be removed.

adrinjalali (commented):
Other than the minor comments, looks brilliant.

skops/io/tests/test_persist.py
E-Aho: Oh sorry, I've done this for clarity

We now do essentially `state = get_state(("hi", [123]), SaveContext(None))`

The reason I did this is that `None` in the original line is the `zipfile` param to memoize certain objects, not something that actually gets persisted. I thought maybe moving it to 3 lines was clearer what was going on.

adrinjalali: Oh my bad, I just misread the parentheses, lol

E-Aho:
Thanks for the PR reviews guys ❤️ I think it should be ready to merge in!

PR #213, opened 2022-11-14
MNT Fixes for hf_hub v0.11.0

BenjaminBossan:
@skops-dev/maintainers This PR started out as a test if upgrading to hf_hub 0.11 would work. Since it caused some mypy errors, I updated the code to fix it, so it should now be compatible once 0.11 is released. Please review.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -54,7 +54,6 @@ jobs:
 
     - name: Install dependencies
       run: |
-        pip install huggingface_hub=="0.11.0rc1"
         pip install .[docs,tests]
         pip install black=="22.6.0" isort=="5.10.1" mypy=="0.981"
         pip uninstall --yes scikit-learn
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -55,6 +55,8 @@ jobs:
     - name: Install dependencies
       run: |
         pip install .[docs,tests]
+        pip uninstall huggingface_hub
+        pip install huggingface_hub=="0.11.0rc0"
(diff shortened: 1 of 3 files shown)


.github/workflows/build-test.yml
```diff
@@ -55,6 +55,8 @@ jobs:
     - name: Install dependencies
       run: |
         pip install .[docs,tests]
+        pip uninstall --yes huggingface_hub
+        pip install huggingface_hub=="0.11.0rc0"
```
adrinjalali, line 59:
```suggestion
        pip install huggingface_hub=="0.11.0rc0"
        pip install .[docs,tests]
```

more like this one, it'd also make it faster. (I don't think it matters if we're not keeping this in the CI though, was thinking more in the direction of adding it permanently to the CI).

BenjaminBossan: Yeah, I just quickly added it, if we choose to create a task for it, your suggestion would be cleaner.

BenjaminBossan:
@skops-dev/maintainers I tried with the new `huggingface_hub==0.11.0rc1" and [CI passed](https://github.com/skops-dev/skops/actions/runs/3487323831/jobs/5834818917) so this should be good to go.

adrinjalali:
Does this also pass with our minimum dependency of `huggingface_hub`?

BenjaminBossan:
> Does this also pass with our minimum dependency of `huggingface_hub`?

Yes, that's 0.10.1, so the most up-to-date before the 0.11 release. I tested it and it works.

adrinjalali (approved):

PR #214, opened 2022-11-21
MNT error in init if given file is empty

adrinjalali:
related to the error raised in https://github.com/skops-dev/skops/pull/207

This makes us raise if a given model file is empty.


BenjaminBossan (approved):
Overall this LGTM. But the escaping error should be fixed before merging.

PR #215, opened 2022-11-27
MNT add helper functions to convert types to strings in .audit

E-Aho:
A few minor changes to the new audit functionality that allows us to pass in types as well as strings as defaults.

This PR adds in a few helper functions to convert from types to their namespace strings, and a helper function that can convert any non-string `defaults` in a passed in list to their namespace.

Also includes new tests that cover this added functionality


skops/io/_utils.py
```diff
@@ -171,3 +171,21 @@ def get_state(value, save_context):
 
     res["__id__"] = __id__
     return res
+
+
+def get_type_name(
+    t: Any,
+) -> str:
+    """Simple helper function to take in a type, and return its name as a string"""
+    return f"{t.__module__}.{t.__name__}"
```
adrinjalali, line 180: `t.__module__` is not always there, or even correct, that's why we have the `_utils.get_module` helper.

skops/io/_dispatch.py
```diff
@@ -110,15 +110,17 @@ def _get_trusted(trusted, default):
 
         This is a convenience method called by child classes.
         """
+        _defaults = cast_defaults_to_strings(default)
```
BenjaminBossan, line 113: Technically, this can be moved below the first return (`if trusted is True...`), resulting in a minuscule performance increase.

skops/io/_utils.py
```diff
@@ -4,7 +4,7 @@
 import sys
 from dataclasses import dataclass, field
 from functools import singledispatch
-from typing import Any
+from typing import Any, List
```
BenjaminBossan, line 7: Just a comment, doesn't need to be changed since we don't do it consistently in the code base: With `from __future__ import annotations`, you can use `list[str]` etc., no need to import `List`.

skops/io/_utils.py
```diff
@@ -171,3 +171,21 @@ def get_state(value, save_context):
...
+
+
+def get_type_name(
+    t: Any,
+) -> str:
+    """Simple helper function to take in a type, and return its name as a string"""
+    return f"{t.__module__}.{t.__name__}"
+
+
+def cast_defaults_to_strings(defaults: List[Any]) -> List[str]:
```
adrinjalali, line 183: maybe call this something like `get_type_paths`, and accept either type or strings, and either way, return strings. Then it's up to the user/us to pass types or strings.

skops/io/_utils.py
```diff
@@ -171,3 +171,21 @@ def get_state(value, save_context):
 
     res["__id__"] = __id__
     return res
+
+
+def get_type_name(
```
BenjaminBossan, line 176: Formatting of this function looks messed up, the whole `def` can fit on one line.

adrinjalali (commented):
Otherwise LGTM.

skops/io/_utils.py
```diff
@@ -171,3 +171,21 @@ def get_state(value, save_context):
...
+    t: Any,
+) -> str:
+    """Simple helper function to take in a type, and return its name as a string"""
```
BenjaminBossan, line 179:
```suggestion
    """Helper function to take in a type, and return its name as a string"""
```

I think "simple" doesn't add anything here. I'm guilty of using words like "simple" myself, but I learned recently that they should be avoided in documentation :)

skops/io/_utils.py
```diff
@@ -171,3 +171,21 @@ def get_state(value, save_context):
...
+    return f"{t.__module__}.{t.__name__}"
+
+
+def cast_defaults_to_strings(defaults: List[Any]) -> List[str]:
+    """Simple helper function that takes in a list of defaults,
```
BenjaminBossan, line 184: Same comment as above, "simple" doesn't help here.

skops/io/_utils.py
BenjaminBossan, line 183:
```suggestion
def cast_defaults_to_strings(defaults: List[Any]) -> List[str]:
```

I'm not so happy with the name here. First there is the "cast" part, I'm not sure if this is actually the right word here. Second, I get why you chose "defaults", as that's where this function is used, but it could be used elsewhere and outside of that context, "defaults" doesn't really describe what the function does.

I wonder if we can just name it `get_type_names`, so very similar to the function above. Also, I'm not sure if that function has to be a separate function at all.

skops/io/_utils.py
```diff
@@ -171,3 +171,21 @@ def get_state(value, save_context):
...
+    and converts any types found to strings"""
+    if not defaults:
+        return []
+    if not type(defaults) in [list, tuple]:
+        defaults = [defaults]
+
+    return [get_type_name(t) if type(t) is not str else t for t in defaults]
```
BenjaminBossan, line 191: Would it not be better to check `if not isinstance(t, str)` or am I missing something?

BenjaminBossan (changes requested):
Pretty good overall, thanks. I only have minor comments, please take a look.

PS: I didn't see Adrin's review when I wrote this one.

skops/io/_utils.py
```diff
@@ -171,3 +171,21 @@ def get_state(value, save_context):
...
+    """Simple helper function to take in a type, and return its name as a string"""
+    return f"{t.__module__}.{t.__name__}"
+
+
+def cast_defaults_to_strings(defaults: List[Any]) -> List[str]:
+    """Simple helper function that takes in a list of defaults,
+    and converts any types found to strings"""
+    if not defaults:
+        return []
+    if not type(defaults) in [list, tuple]:
```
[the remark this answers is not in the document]
E-Aho: Ah I see, I wasn't fully super familiar with mypy, that's handy! I'll rework this to accept `Any` instead of just `list`s

So in general, for internal functions, given we use mypy, are type checks like this best practice, or should we rely on mypy for that?

skops/io/_utils.py
E-Aho: That also works, only difference would be if something was a subclass of `str`, which is unlikely to ever be a problem for us (unless someone wanted to check the type safety of something that extends `str`).

Happy to rework this to be an `isinstance` check instead, I felt it would be safer to explicitly check that something is exactly `str` though

skops/io/_utils.py
E-Aho: I separated them so that the logic would be easier to read & test imo, but you're right that it could be a single function as it's currently only used together.

Happy to rework to be a single big function that combines the functionality if you'd prefer

skops/io/tests/test_utils.py
```diff
@@ -0,0 +1,57 @@
...
+    @pytest.mark.parametrize(
+        "input, output",
+        [
+            (None, []),
+            (int, ["builtins.int"]),
+            ((list,), ["builtins.list"]),
+        ],
+        ids=["None", "single int type", "list in tuple"],
+    )
+    def test_for_edge_cases_handles_as_expected(self, input, output):
```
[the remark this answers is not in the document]
E-Aho: In my head they were "edge cases" as they weren't happy path tests. 

Personal preference I guess, I normally like to seperate happy path tests from corner/error tests, but happy to conform a different test style

skops/io/_utils.py
BenjaminBossan: Hmm, maybe having a strict check here is more secure? But I can't really come up with a possible scenario where it would matter. FWIW, `UserString` would not pass the `isinstance` check. @adrinjalali any opinion?

skops/io/_utils.py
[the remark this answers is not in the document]
BenjaminBossan: Technically, mypy should warn if the wrong type is passed and thus it shouldn't happen. In practice, since not all our code is typed, or some functions just return `Any`, we could still end up calling this function with a non-list without mypy warning us. Therefore, I think the check has its merit and at the same time, we should keep the annotation, so scratch what I said earlier.

Usually, I would be fine not checking and failing when a wrong type is passed (i.e. rely on duck typing), the problem is that if a user were to pass a string, since it's iterable, it would work but the totally wrong thing would happen, so the check is needed.

Just note that this time, I think it's really better to check `isinstance(defaults, (list, tuple))` :)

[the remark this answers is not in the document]
adrinjalali: I wouldn't rely on mypy for runtime type checks, those are two different things. The functions should still check for validity of their input.

skops/io/_utils.py
BenjaminBossan: > I separated them so that the logic would be easier to read & test imo, but you're right that it could be a single function as it's currently only used together.

Up to you, I don't insist.

skops/io/_utils.py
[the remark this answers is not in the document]
BenjaminBossan: Well, the majority of our functions don't validate the input, right? :) So I think it's more of a matter of checking where there is a reasonable error case that would be missed otherwise.

As to mypy, if we had a strictly typed code base (which we chose not to enforce), all internal functions would only be called by the annotated types, otherwise mypy would complain (excluding mypy bugs).

skops/io/_utils.py
BenjaminBossan: > Here we only want to compare strings ... So I'd do isinstance here rather than strict type checking

Do I misunderstand or is this a contradiction? If we only want to compare strings and not any subclasses, should we not do the strict check, i.e.  `type(t)`?

So if a user creates a subclass from string and then says `trusted = [MyString("foo.bar")]`, would we trust `foo.bar` or `mytypes.MyString` (or whatever its module is)? (Note that the latter would actually fail right now with `'MyString' object has no attribute '__name__'`)

E-Aho: Yeah, that exact case was the reason I did this strict check in the first place, but I think that actually hits on the whole problem.

`MyString("foo.bar")` is no longer a type, it's an instance. So we should in that case treat it as `foo.bar` instead of `MyString`. 

I'll add a test that covers this and change it to support this.

Essentially, the functionality should (I think) be:

`UserDefinedString -> ["module.UserDefinedString"]`
`UserDefinedString("foo") ->["foo"]`

adrinjalali: I'm not sure what the issue here is, I might be missing something. This is how I think about it:

```python
def get_type_paths(trusted):
	def _get_str(value):
		if isinstance(value, str):
			return value
		if type(value) is type:
			return get_module(value)
		raise Exception()

	return [_get_str(value) for value in trusted]
```

The user either passes types, or instances, for instances we keep them if they're strings (or subclasses), for types, we return their address.

skops/io/_dispatch.py
```diff
@@ -110,15 +110,17 @@ def _get_trusted(trusted, default):
...
         """
+
         if trusted is True:
             # if trusted is True, we trust the node
             return True
 
         if trusted is False:
             # if trusted is False, we only trust the defaults
-            return default
+            _defaults = get_type_paths(default)
```
adrinjalali, line 120: I'm not sure where exactly to put this conversion. We should convert everything, including the `trusted` list here as well.

So maybe do `return get_type_paths(trusted)` as well?

Or should we do that inside `load` and `get_untrusted_set`?

Also, why not `return get_type_paths(defaults)`?

adrinjalali (commented):
Otherwise LGTM

skops/io/_utils.py
BenjaminBossan: > The user either passes types, or instances, for instances we keep them if they're strings (or subclasses), for types, we return their address.

In general, I agree (and that's also how I understand @E-Aho's suggestion), this would be a reasonable approach that's also easy to remember. I just have two concerns:

1. Is there any security issue with allowing custom types there? As I said, I can't come up with any.
2. `collections.UserString` is not an instance of str, if we want to allow user defined strings, should we allow that?

adrinjalali: I think of supporting that if at some point somebody opens an issue with this request. I don't think it's used enough for us to complicate things for it at this point.

BenjaminBossan: Okay, so to summarize, the agreement is to check `isinstance` of `str`, not `type` `str`, not to support `UserString`, and that we're not aware of any security issue caused by `isinstance` vs `type` check.

adrinjalali (approved):
We can have a separate PR to work on the public API allowing users to pass types.

BenjaminBossan (approved):
Looks, great, this PR can be merged, there are just two minor improvements for docstrings that I would suggest.

adrinjalali:
you might want to enable pre-commit hooks, which include `isort` ;)

E-Aho:
I've got that all set up on my personal computer, trying to wrestle through it on my work laptop atm to get this merged lol

E-Aho:
This is a good reminder to me how nice it is to have pre-commit hooks 🙃 Should be good to go now

adrinjalali:
Note: I change the commit title to MNT since this doesn't expose any changes to the user. I feel like ENH has user-facing implications.

PR #218, opened 2022-11-28
MNT set version for the release

adrinjalali:
@BenjaminBossan nothing left for the release I guess? 

I can create the tag on the branch and release after we merge this PR.


BenjaminBossan:
We should be good, the only one I would have liked to see included is #142 but it's not a must have.

adrinjalali:
Yeah I agree it would have been nice. But we can include it in the next one. So I'll merge and continue the release process.

PR #234, opened 2022-12-01
ENH: Make skops.io.dump and load work with TextIOWrapper

BenjaminBossan:
(note: `load` already worked, only `dump` needed change)

The idea here is to make it possible to use `skops.io.dump` and `load` with a file wrapper, i.e. using `with open(...)`. This makes it easier to search and replace `pickle.dump` and` load` by `skops.io.dump` and `load`.

## Implementation

To decide how to check the way to treat the file argument, I went with `hasattr(file, "write")`, as this is what `np.save` does. I assume this is more robust than checking the type. Unfortunately, mypy doesn't understand this, so I had to tell it to shut up.

I didn't update the docstring or any examples because I don't think we need to "advertise" this usage.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,7 +11,7 @@ skops Changelog
 
 v0.4
 ----
-- :func:`.io.dump` and :func:`.io.load` now work with :class:`io.TextIOWrapper`,
+- :func:`.io.dump` and :func:`.io.load` now work with file like objects,
   which means you can use them with the ``with open(...) as f: dump(obj, f)``
   pattern, like you'd do with ``pickle``. :pr:`234` by `Benjamin Bossan`_.
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,6 +11,9 @@ skops Changelog
 
 v0.4
 ----
+- :func:`.io.dump` and :func:`.io.load` now work with :class:`io.TextIOWrapper`,
+  which means you can use them with the ``with open(...) as f: dump(obj, f)``
+  pattern, like you'd do with ``pickle``. :pr:`234` by `Benjamin Bossan`_.
 
 v0.3
 ----
--- a/skops/io/_persist.py
+++ b/skops/io/_persist.py
@@ -66,11 +66,11 @@ def dump(obj: Any, file: str | Path | BinaryIO) -> None:
     """
     buffer = _save(obj)
 
-    if hasattr(file, "write"):
-        file.write(buffer.getbuffer())  # type: ignore
-    else:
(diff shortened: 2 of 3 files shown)


BenjaminBossan:

skops/io/_persist.py
```diff
@@ -39,7 +39,7 @@ def _save(obj: Any) -> io.BytesIO:
     return buffer
 
 
-def dump(obj: Any, file: str) -> None:
+def dump(obj: Any, file: str | io.TextIOWrapper) -> None:
```
adrinjalali, line 42: is it a textiowrapper or anything which has `write`?

adrinjalali (commented):
This sounds pretty good to me.

Shouldn't load follow the same pattern?

Also needs a changelog.

skops/io/_persist.py
BenjaminBossan: It's the latter, if you prefer I can create a protocol for objects with write attributes, just thought `TextIOWrapper` is the most common case by far.

BenjaminBossan:
> Shouldn't load follow the same pattern?

`load` already works because `ZipFile` takes care of correctly dealing with the `file` argument.

> Also needs a changelog.

Done, can be reviewed again.

skops/io/_persist.py
adrinjalali: Or not have it typed? 👻 

We also accept a `Path`.

What we really accept here, is a file-like object, or a path like object. And that's best explained in the docstring w/o any types.

image

BenjaminBossan: I dug a bit further, it seems that `typing.BinaryIO` would be the correct type:

https://docs.python.org/3/library/typing.html#typing.BinaryIO

This allows mypy to flag code like the one below which uses the wrong mode:

```python
with open(file, 'w') as f:  # <= should be 'wb' mode
    dump(f)
```

WDYT?

Regarding `Path`: Yeah, that's supported, I originally didn't add it because it's not related to this PR, but I can still add it here.

adrinjalali: Ok, I could live with `BinaryIO` then.

E-Aho (approved):
LGTM

docs/changes.rst
```diff
@@ -11,6 +11,9 @@ skops Changelog
 
 v0.4
 ----
+- :func:`.io.dump` and :func:`.io.load` now work with :class:`io.TextIOWrapper`,
```
```suggestion
- :func:`.io.dump` and :func:`.io.load` now work with file like objects,
```

adrinjalali (approved):

PR #235, opened 2022-12-02
CI fix clean_skops.py

### LINKED ISSUE #231: clean_skops.py job broken
The job is broken: 

```
Run echo "y" | python scripts/clean_skops.py
/opt/hostedtoolcache/Python/3.11.0/x64/lib/python3.11/site-packages/huggingface_hub/utils/_deprecation.py:233: FutureWarning: 'list_models' currently returns a list of objects but is planned to be a generator starting from version 0.14 in order to implement pagination. Please avoid to use `list_models(...).__len__` or explicitly convert the output to a list first with `list(iter(list_models)(...))`.
  warnings.warn(self._deprecation_msg.format(attr_name=attr_name), FutureWarning)
Are you sure you want to delete all repos under skops-ci older than 7 days? (y/[n])Found 1 models, checking their age...
Skipping model: skops-ci/test-bf49e533-8b1f-40ef-9084-fcad0e8bf7c5, age: 1
```

There's also a deprecation issue in the GH CI definition which needs to be fixed.

adrinjalali:
Fixes #231

This fixes the broken `clean_skops.py`.

- We can now pass the token to the client instance
- The output of `list_models` is a generator now
- Most models are private, so the token was required to list them, which were previously ignored

cc @skops-dev/maintainers 

The script works locally now.

### DIFF
--- a/scripts/clean_skops.py
+++ b/scripts/clean_skops.py
@@ -10,19 +10,19 @@ from huggingface_hub import HfApi
 
 # This is the token for the skops user. TODO remove eventually, see issue #47
 token = "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"
-client = HfApi()
-user = client.whoami(token=token)["name"]
+client = HfApi(token=token)
+user = client.whoami()["name"]
 answer = input(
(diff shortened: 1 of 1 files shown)


BenjaminBossan (approved):
Thanks for fixing this.

PR #236, opened 2022-12-02
MNT cleanup user links in the changelog

adrinjalali:
For users who are not in the `authors.rst` list, they should be referred to with the `:user:` directive.

The `_authors.rst` file is only there for convenience so that for recurring contributors we don't have to repeat the `:user:` directive.


BenjaminBossan (approved):
It works, great.

PR #237, opened 2022-12-04
ENH All sklearn estimators in trusted list

### LINKED ISSUE #223: Add all scikit-learn estimators to the default trusted list
We can read this list from `all_estimators` and trust them all. This doesn't include everything from sklearn, but is a good starting point and would substantially reduce the size of the `untrusted_types`, which would in turn make it easier for folks to notice malicious types in it.

We can probably read it dynamically from sklearn instead of hard coding it. Other installed/loaded packages might change the list, but if the user has some malicious packages installed, they're already compromised anyway.

EdAbati:
Fixes #223 

Hi all, I have added all `sklearn` estimators to the trusted list of `ObjectNode` and added test for it. 

A couple of things that I am not 100% sure about:
- the issue mentioned to add some 'sanity checks' when getting the list of estimators. For now I am checking that the module name starts with `'sklearn.'`, should I add any extra ones?
- is `ObjectNode` the only place where we should add the sklearn estimators in the trusted list? What about `TypeNode`?
- I have extracted `get_tested_estimators` in a common place (`testing_utils.py`) so I could reuse it. I could not use the name `common.py` because there is already a `common.py` in `hub_utils` and `pytest` complained. Should we maybe have a common utils file for all the tests?


skops/io/tests/test_audit.py
```diff
@@ -146,6 +148,15 @@ def __init__(self):
     assert not hasattr(temp, "b")
 
 
+@pytest.mark.parametrize(
+    "estimator", get_tested_estimators(), ids=_get_check_estimator_ids
+)
+def test_sklearn_trusted_types(estimator):
+    untrusted_types = get_untrusted_types(data=dumps(estimator))
+    sklearn_untrusted_types = [t for t in untrusted_types if t.startswith("skelarn.")]
+    assert len(sklearn_untrusted_types) == 0
```
adrinjalali, line 157: This is correct as what we need to test, but it's adding a round of dumping estimators to the whole test suite. I wonder if we should add it right after where we do `get_untrusted_types` in the existing tests, instead of adding a new round of tests.

WDYT @skops-dev/maintainers

adrinjalali (commented):
Thanks for the PR @EdAbati , this looks great.

BenjaminBossan:
@adrinjalali You are most qualified to review this, would you take it?

> * I have extracted `get_tested_estimators` in a common place (`testing_utils.py`) so I could reuse it. I could not use the name `common.py` because there is already a `common.py` in `hub_utils` and `pytest` complained. Should we maybe have a common utils file for all the tests?

I don't think we need a "common" `common.py`, since the different sub-modules deal with quite different objectives. Regarding the name, I feel like `testing_` is a bit redundant, since we're already in the `tests` directory, so how about just `_utils.py`?

skops/io/tests/test_audit.py
BenjaminBossan: I would be fine either way, having different tests for different things makes sense but performance also matters.

Btw., there is a typo, `"skelarn."`

EdAbati: (Oops, fixed the typo in the meantime)

EdAbati: I moved this `assert` into `test_can_persist_fitted`. Some of the `sklearn` types that are still untrusted were revealed by testing this with a fitted estimator. (see comment below)

skops/io/tests/test_persist.py
```diff
@@ -67,6 +67,47 @@
 # Python 3.8)
 ATOL = 1e-6 if sys.platform == "darwin" else 1e-7
 
+# TODO: remove when these are added to trusted types
+SKLEARN_EXCLUDED_TYPES = (
```
EdAbati, line 71: These are some of the `sklearn` types that are still untrusted. Therefore, I would like to exclude them when we look for `untrusted_types` in `test_can_persist_fitted`.

I don't think this is the _cleanest_ way to write something like this. Can you think of a better alternative?

skops/io/tests/test_persist.py
```diff
@@ -67,6 +67,47 @@
...
+    "sklearn.gaussian_process.kernels.ConstantKernel",
+    "sklearn.gaussian_process.kernels.Product",
+    "sklearn.gaussian_process.kernels.RBF",
+    "sklearn.impute._iterative._ImputerTriplet",
+    "sklearn.metrics._dist_metrics.EuclideanDistance",
+    "sklearn.metrics._scorer._passthrough_scorer",
+    "sklearn.model_selection._split.StratifiedKFold",
+    "sklearn.multiclass._ConstantPredictor",
+    "sklearn.neighbors._ball_tree.BallTree",
+    "sklearn.neighbors._kd_tree.KDTree",
```
EdAbati, line 106: Interestingly `sklearn.neighbors._ball_tree.BallTree` and `sklearn.neighbors._kd_tree.KDTree` are not present in the list returned by `all_estimators`

skops/io/tests/test_persist.py
adrinjalali: I think instead of doing this, we should check if any of the types reported by `get_untrusted_set` belong to `SKLEARN_ESTIMATOR_TYPE_NAMES`, and avoid having this list here.

adrinjalali (commented):
It's definitely worth having a way to get other types which are not trusted by this PR, but I don't think we should have that list in the codebase.

EdAbati:
Hi @adrinjalali , I tweaked the unittest it looks nicer now, thanks :) 

Regarding instead:
> It's definitely worth having a way to get other types which are not trusted 

do you have anything in mind?

skops/io/tests/test_persist.py
```diff
@@ -468,7 +468,7 @@ def get_input(estimator):
 @pytest.mark.parametrize(
     "estimator", _tested_estimators(), ids=_get_check_estimator_ids
 )
-def test_can_persist_fitted(estimator, request):
```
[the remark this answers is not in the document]
adrinjalali: I still don't understand pytest's `request`, @BenjaminBossan would know.

adrinjalali (commented):
This LGTM. Thanks @EdAbati . Could you please also add an entry in our changelog?

skops/io/tests/test_persist.py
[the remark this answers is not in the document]
BenjaminBossan: It's indeed unnecessary here.

EdAbati:
Thank you @adrinjalali for the help! :)
I have updated the changelog.

adrinjalali (approved):

PR #239, opened 2022-12-05
CI add dependabot config for GH actions

adrinjalali:
I realized we could also enable dependabot for GH actions instead of manually chasing after them :D 

cc @skops-dev/maintainers


BenjaminBossan (approved):
Hah, nice, but who keeps dependabot up-to-date? ;)

Issue #226, opened 2022-11-30
Make sure xgboost.XGBClassifier and xgboost.XGBRegressor can be persisted


adrinjalali:
To increase reach, we should make sure the sklearn compatible xgboost estimators can be persisted.

This issue is to:
- check if we can persist them out of the box
- maybe add a test and to the CI to make sure they can be persisted
- if needed, figure out what we'd need to implement on our side, or on the xgb side, to make this happen

BenjaminBossan:
Okay, I did a quick check with the 3 big ones:

1. XGBoost fails.
2. Catboost fails.
3. LightGBM: actually works for classifier and regressor!

Obviously, there are so many options with these estimators that a quick test can't cover all of them, so there might be cases where different results will occur.

Here is the code I used for testing LightGBM. Similar code was used for the others.

```python
# python -m pip install lightgbm
from lightgbm.sklearn import LGBMClassifier, LGBMRegressor
import numpy as np
from skops import io

X = np.array([[1, 1], [1, 2], [2, 2], [2, 3]])
y = np.dot(X, np.array([1, 2])) + 3

regr = LGBMRegressor().fit(X,y)
lr = io.loads(io.dumps(regr), trusted=[
    'collections.defaultdict', 'lightgbm.basic.Booster', 'lightgbm.sklearn.LGBMRegressor'])
lr.predict(X)

clf = LGBMClassifier().fit(X,y)
lc = io.loads(io.dumps(clf), trusted=[
    'collections.defaultdict', 'lightgbm.basic.Booster', 'lightgbm.sklearn.LGBMClassifier',
    'numpy.int64', 'sklearn.preprocessing._label.LabelEncoder'])
lc.predict_proba(X)
```

PR #244, opened 2022-12-08
Add support for LightGBM and XGBoost

BenjaminBossan:
Partially solves #226

## Description

LightGBM already worked out of the box, so for this, only tests were added. For XGBoost, extra code was added to support it.

## Implementation

Dependencies were extended to include xgboost and lightgbm for testing, so that they should be covered by CI. If those libraries are not installed, the tests should be gracefully skipped.

For xgboost, the xgboost-specific persistence format was chosen, which can be used through `estimator.save_model` and `estimator.load_model`. I could not verify (yet) that this format is 100% secure. However, since it's a binary json file (`.ubj`), it should as secure as json. More information here:

https://xgboost.readthedocs.io/en/stable/tutorials/saving_model.html

The actual implementation for xgboost mirrors the one for numpy arrays, which also stores its data in an external file. However, I did not apply any caching here, as it seems it wouldn't be useful.

I decided to create a new test file, `test_external.py`, to put the new tests, instead of cramming more into `test_persist.py`. Since some functions from `test_persist.py` were required here, I moved them a new `_utils.py` module.

Interestingly, I found some potential bugs where some values returned by `get_params()` may differ for xgboost models after loading. This is independent of skops. Those edge cases are documented but were excluded from tests, as it appears that they should be inconsequential. We may want to report them.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -87,6 +87,24 @@ means if you have custom functions (say, a custom function to be used with
 most ``numpy`` and ``scipy`` functions should work. Therefore, you can actually
 save built-in functions like ``numpy.sqrt``.
 
+Supported libraries
+-------------------
+
+Skops intends to support all of **scikit-learn**, that is, not only its
+estimators, but also other classes like cross validation splitters. Furthermore,
+most types from **numpy** and **scipy** should be supported, such as (sparse)
+arrays, dtypes, random generators, and ufuncs.
+
+Apart from this core, we plan to support machine learning libraries commonly
+used be the community. So far, those are:
+
+- `LightGBM <lightgbm.readthedocs.io/>`_ (scikit-learn API)
+
+If you run into a problem using any of the mentioned libraries, this could mean
+there is a bug in skops. Please open an issue on `our issue tracker
+<https://github.com/skops-dev/skops/issues>`_ (but please check first if a
+corresponding issue already exists).
+
 Roadmap
 -------
 
--- a/skops/io/_xgboost.py
+++ b/skops/io/_xgboost.py
@@ -1,0 +1,88 @@
+from __future__ import annotations
+
+import io
+import tempfile
+from typing import Any, Sequence
+from uuid import uuid4
+
+import sklearn.exceptions
+
+from ._audit import Node, get_tree
+from ._utils import LoadContext, SaveContext, get_module, get_state
+
+try:
+    import xgboost
+except ImportError:
+    xgboost = None  # type: ignore
+
+
+def xgboost_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
+    res = {
+        "__class__": obj.__class__.__name__,
+        "__module__": get_module(type(obj)),
+        "__loader__": "XGBoostNode",
+        "constructor": type(obj).__name__,
+    }
+
+    # fitted vs not fitted models require different approaches
+    try:
+        # xgboost doesn't allow to save to a memory buffer, so take roundtrip
+        # through temp file
+        tmp_file = f"{tempfile.mkdtemp()}.ubj"
+        obj.save_model(tmp_file)
+        with open(tmp_file, "rb") as f:
+            data_buffer = io.BytesIO(f.read())
+        f_name = f"xgboost_{uuid4()}.json"
+        save_context.zip_file.writestr(f_name, data_buffer.getbuffer())
+        res.update(type="xgboost-format", file=f_name)
+    except sklearn.exceptions.NotFittedError:
+        res.update(type="params", content=get_state(obj.get_params(), save_context))
+
+    return res
+
+
+class XGBoostNode(Node):
+    def __init__(
+        self,
+        state: dict[str, Any],
+        load_context: LoadContext,
+        trusted: bool | Sequence[str] = False,
+    ) -> None:
+        super().__init__(state, load_context, trusted)
+        self.type = state["type"]
+        self.trusted = self._get_trusted(trusted, [])
+
+        # list of constructors is hard-coded for higher security
+        constructors = {
+            "XGBClassifier": xgboost.XGBClassifier,
+            "XGBRegressor": xgboost.XGBRegressor,
+            "XGBRFClassifier": xgboost.XGBRFClassifier,
+            "XGBRFRegressor": xgboost.XGBRFRegressor,
+            "XGBRanker": xgboost.XGBRanker,
+        }
+        self.children = {"constructor": constructors[state["constructor"]]}
+        if self.type == "xgboost-format":  # fitted xgboost model
+            self.children["content"] = io.BytesIO(load_context.src.read(state["file"]))
+        else:
+            self.children["content"] = get_tree(state["content"], load_context)
+
+    def _construct(self):
+        cls = self.children["constructor"]
+        instance = cls()
+
+        if self.type == "xgboost-format":  # fitted
+            # load_model works with bytearray, so no temp file necessary here
+            instance.load_model(bytearray(self.children["content"].getvalue()))
+        else:
+            params = self.children["content"].construct()
+            instance.set_params(**params)
+
+        return instance
+
+
+GET_STATE_DISPATCH_FUNCTIONS = []  # type: ignore
+NODE_TYPE_MAPPING = {}
+
+if xgboost is not None:
+    GET_STATE_DISPATCH_FUNCTIONS.append((xgboost.sklearn.XGBModel, xgboost_get_state))
+    NODE_TYPE_MAPPING["XGBoostNode"] = XGBoostNode
--- a/skops/io/tests/_utils.py
+++ b/skops/io/tests/_utils.py
@@ -1,0 +1,145 @@
+import warnings
+
+import numpy as np
+from scipy import sparse
+from sklearn.base import BaseEstimator
+
+
+def _is_steps_like(obj):
+    # helper function to check if an object is something like Pipeline.steps,
+    # i.e. a list of tuples of names and estimators
+    if not isinstance(obj, list):  # must be a list
+        return False
+
+    if not obj:  # must not be empty
+        return False
+
+    if not isinstance(obj[0], tuple):  # must be list of tuples
+        return False
+
+    lens = set(map(len, obj))
+    if not lens == {2}:  # all elements must be length 2 tuples
+        return False
+
+    keys, vals = list(zip(*obj))
+
+    if len(keys) != len(set(keys)):  # keys must be unique
+        return False
+
+    if not all(map(lambda x: isinstance(x, (type(None), BaseEstimator)), vals)):
+        # values must be BaseEstimators or None
+        return False
+
+    return True
+
+
+def _assert_generic_objects_equal(val1, val2):
+    def _is_builtin(val):
+        # Check if value is a builtin type
+        return getattr(getattr(val, "__class__", {}), "__module__", None) == "builtins"
+
+    if isinstance(val1, (list, tuple, np.ndarray)):
+        assert len(val1) == len(val2)
+        for subval1, subval2 in zip(val1, val2):
+            _assert_generic_objects_equal(subval1, subval2)
+            return
+
+    assert type(val1) == type(val2)
+    if hasattr(val1, "__dict__"):
+        assert_params_equal(val1.__dict__, val2.__dict__)
+    elif _is_builtin(val1):
+        assert val1 == val2
+    else:
+        # not a normal Python class, could be e.g. a Cython class
+        assert val1.__reduce__() == val2.__reduce__()
+
+
+def _assert_tuples_equal(val1, val2):
+    assert len(val1) == len(val2)
+    for subval1, subval2 in zip(val1, val2):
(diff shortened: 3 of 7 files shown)


BenjaminBossan:
I removed the xgboost bug (?) docstring, since I couldn't find how to exclude it from running with doctest. Replicating it here so that it doesn't get lost. Ideally, someone with more experience in xgboost could comment on whether this really is a bug or not. If so, we could submit it to the xgboost issue tracker.

```python
>>> import xgboost
>>> print(xgboost.__version__)
1.7.1
>>> estimator = xgboost.XGBClassifier(tree_method=None)
>>> X, y = [[0, 1], [2, 3]], [0, 1]
>>> estimator.fit(X, y)
XGBClassifier(...)
>>> print(estimator.tree_method)
None
>>> print(estimator.get_params()["tree_method"])
exact
>>> # after save/load roundtrip, tree_method's get_params value changes from exact to auto
>>> import tempfile
>>> tmp_file = f"{tempfile.mkdtemp()}.ubj"
>>> estimator.save_model(tmp_file)
>>> estimator.load_model(tmp_file)
>>> print(estimator.tree_method)
None
>>> print(estimator.get_params()["tree_method"])
auto
>>> # a similar thing happens with gpu_id
>>> estimator = xgboost.XGBClassifier(tree_method='gpu_hist', booster='gbtree')
>>> estimator.fit(X, y)
XGBClassifier(...)
>>> print(estimator.gpu_id)
None
>>> print(estimator.get_params()["gpu_id"])
0
>>> # after save/load roundtrip, gpu_id's get_params value changes from 0 to -1
>>> estimator.save_model(tmp_file)
>>> # for gpu_id, the estimator needs to be re-initialized for the effect to occur
>>> estimator = xgboost.XGBClassifier()
>>> estimator.load_model(tmp_file)
>>> print(estimator.gpu_id)
None
>>> print(estimator.get_params()["gpu_id"])
-1
```

skops/io/tests/_utils.py
```diff
@@ -0,0 +1,145 @@
+import warnings
```
BenjaminBossan, line 1: Everything in here is unmodified, just moved around.

docs/persistence.rst
```diff
@@ -87,6 +87,24 @@ means if you have custom functions (say, a custom function to be used with
...
+
+Skops intends to support all of **scikit-learn**, that is, not only its
+estimators, but also other classes like cross validation splitters. Furthermore,
+most types from **numpy** and **scipy** should be supported, such as (sparse)
+arrays, dtypes, random generators, and ufuncs.
+
+Apart from this core, we plan to support machine learning libraries commonly
+used be the community. So far, those are:
+
+- `LightGBM <lightgbm.readthedocs.io/>`_ (scikit-learn API)
```
adrinjalali, line 101: shouldn't we add xgboost here as a result of this PR?

skops/io/_xgboost.py
```diff
@@ -0,0 +1,88 @@
...
+        self.trusted = self._get_trusted(trusted, [])
+
+        # list of constructors is hard-coded for higher security
+        constructors = {
+            "XGBClassifier": xgboost.XGBClassifier,
+            "XGBRegressor": xgboost.XGBRegressor,
+            "XGBRFClassifier": xgboost.XGBRFClassifier,
+            "XGBRFRegressor": xgboost.XGBRFRegressor,
+            "XGBRanker": xgboost.XGBRanker,
+        }
```
adrinjalali, line 62: why can't we use the type saved in `__class__` here?

skops/io/_xgboost.py
```diff
@@ -0,0 +1,88 @@
...
+        "__module__": get_module(type(obj)),
+        "__loader__": "XGBoostNode",
+        "constructor": type(obj).__name__,
+    }
+
+    # fitted vs not fitted models require different approaches
+    try:
+        # xgboost doesn't allow to save to a memory buffer, so take roundtrip
+        # through temp file
+        tmp_file = f"{tempfile.mkdtemp()}.ubj"
```
adrinjalali, line 31: can we use a context manager to make sure it's cleaned up after the fact?

skops/io/_xgboost.py
```diff
@@ -0,0 +1,88 @@
...
+        else:
+            self.children["content"] = get_tree(state["content"], load_context)
+
+    def _construct(self):
+        cls = self.children["constructor"]
+        instance = cls()
+
+        if self.type == "xgboost-format":  # fitted
+            # load_model works with bytearray, so no temp file necessary here
+            instance.load_model(bytearray(self.children["content"].getvalue()))
```
adrinjalali, line 75: I really don't like this. I can't easily figure out what this actually does, and if it's anywhere near being safe

adrinjalali (commented):
Could you maybe explain why going through the `__dict__` wouldn't work for xgboost? I suppose because the attributes are not C extension types, and rather pure C types?

I would like us to understand what save/load of xgboost does, before adding this. We should document that somewhere.

skops/io/_xgboost.py
BenjaminBossan: This is the equivalent of what we do with numpy arrays (in fact, the whole approach is analogous):

```python
class NdArrayNode(Node):
    def __init__(...):
        ...
        if self.type == "numpy":
            self.children = {
                "content": io.BytesIO(load_context.src.read(state["file"]))
            }

    def _construct(self):
        if self.type == "numpy":
            content = np.load(self.children["content"], allow_pickle=False)
        ...
```

The difference is that, unfortunately, `load_model` cannot work with `bytes`, instead the content first needs to be converted to a `bytearray` instead.

I could also load the model from a temp file, but I thought this approach would be better.

skops/io/_xgboost.py
BenjaminBossan: My thinking was that it's more secure, but we could also store the class itself and use that here.

BenjaminBossan (commented):
> Could you maybe explain why going through the `__dict__` wouldn't work for xgboost? I suppose because the attributes are not C extension types, and rather pure C types?

So without special treatment, xgboost goes through `ObjectNode`. Since it has `__getstate__` and `__setstate__`, those are being used. After we call the latter, there is some kind of validation going on:

https://github.com/dmlc/xgboost/blob/8824b40961c2f381fcd7a9a6af1697cbf20b608f/python-package/xgboost/core.py#L1674-L1684

This leads us to the C++ part of the code base, which I honestly didn't even want to inspect. 

My thinking is that trying to make that approach work is too error prone, we might get something to run by tinkering with the attributes, but we couldn't be sure that it would work with other parameters. However, the `save_model` way is officially recommended, is (presumably) well tested, _and_ it is stable across versions, so more robust for users than if we were directly storing attributes.

> I would like us to understand what save/load of xgboost does, before adding this. We should document that somewhere.

Regarding security, I couldn't find anything being mentioned explicitly, but the general advantages are explained in the [linked document](https://xgboost.readthedocs.io/en/stable/tutorials/saving_model.html). As I said, since it's json format, I don't see how that would not be safe.

adrinjalali:
So looking at this again, I don't understand why you think using `__setstate__` would be error prone. It can be the case that somebody creates a custom `.skops` file where they put odd attributes there, but if the user loads a file where the input to `__setstate__` is exactly what we got from `__getstate__`, then there shouldn't be any issues.

It seems the issue we have is that we don't support `bytearray`, I think. So maybe a better solution here would be to add support for that, which in turn would fix the issue for xgb w/o us explicitly having to do an xgb specific kinda thing.

BenjaminBossan:
@adrinjalali I tested your suggestion to add support for `byterray` and it is indeed sufficient for supporting xgboost. The more general point still stands to be decided.

adrinjalali:
> The more general point still stands to be decided.

I agree. But I think for now we can do with only supporting bytes/byte array, and figure out supporting custom formats of 3d party libraries later.

One thing I wonder: should we store bytes in a binary file instead? I feel like it'd make save/load much faster, and keep the schema json file cleaner.

BenjaminBossan:
> I agree. But I think for now we can do with only supporting bytes/byte array, and figure out supporting custom formats of 3d party libraries later.

Okay, then let's start with that.

> One thing I wonder: should we store bytes in a binary file instead? I feel like it'd make save/load much faster

It'll probably depend. In this case, maybe, but having the additional i/o is probably a big overhead if we have multiple small bytes.

adrinjalali:
> It'll probably depend. In this case, maybe, but having the additional i/o is probably a big overhead if we have multiple small bytes.

I wanted to suggest having a threshold for having them in the schema or in the file, but then I realized another thing to think of, is that if we allow binary data in the json schema, then people can create `.skops` files where there is a ton of binary data there, and we'd read them. That doesn't sound like a good idea to me. I'd be more comfortable reading those only from files. WDYT?

BenjaminBossan:
> then people can create `.skops` files where there is a ton of binary data there, and we'd read them. That doesn't sound like a good idea to me

I'm not against your proposal, I just want to understand what the concern is. Is it only about readability? Since you could argue that even if there is a big blob in the json, it's only one line so doesn't distract much (it could slow down some editors though).

adrinjalali:
Slowing down an editor is one thing, loading a massive json into memory is another thing. I'm not sure how people could exploit this, but I kinda feel like keeping them in separate files is cleaner. Imagine people having images as bytes, and we'd be keeping them in the json file all together.

Besides, I don't think there are many cases where we'd encounter small byte arrays. So my guess is that in most cases those byte arrays are large rather than small, and therefore having all of them in files would speed things up on average.

BenjaminBossan:
@adrinjalali I made the changes as you suggested, please review again. For good measure, I checked if adding the bytes support is sufficient for making catboost work and indeed it does, so I added tests for it as well.

skops/io/_general.py
```diff
@@ -475,12 +477,55 @@ def _construct(self):
...
+
+class BytesNode(Node):
+    def __init__(
+        self,
+        state: dict[str, Any],
+        load_context: LoadContext,
+        trusted: bool | Sequence[str] = False,
+    ) -> None:
+        super().__init__(state, load_context, trusted)
+        self.trusted = self._get_trusted(trusted, [])
```
adrinjalali, line 506: shouldn't we trust `bytes` and `bytearray`?

skops/io/tests/test_external.py
```diff
@@ -0,0 +1,325 @@
...
+    assert sum(group) == N_SAMPLES
+    return X, y, group
+
+
+def check_estimator(estimator, trusted):
+    loaded = loads(dumps(estimator), trusted=trusted)
+    assert_params_equal(estimator.get_params(), loaded.get_params())
+
+
+def check_method_outputs(estimator, X, trusted):
```
adrinjalali, line 52: isn't this common between these tests and scikit-learn ones? Would be nice to avoid duplication as much as we can here.

skops/io/tests/test_external.py
```diff
@@ -0,0 +1,325 @@
...
+
+    - When initialzing with tree_method=None, its value resolves to "exact", but
+      after loading, it resolves to "auto" when calling get_params().
+    - When initializing with tree_method='gpu_hist' and gpu_id=None, the
+      latter's value resolves to 0, but after loading, it resolves to -1, when
+      calling get_params()
+
+    These discrepancies occur regardless of skops, so they're a problem in
+    xgboost itself. We assume that this has no practical consequences and thus
+    avoid testing these cases.
```
adrinjalali, line 154: is there an upstream issue for this on the xgboost side? If not, it'd be nice to open one and link to it here.

adrinjalali (commented):
Pretty cool 🔥

skops/io/tests/test_external.py
BenjaminBossan: As [I commented](https://github.com/skops-dev/skops/pull/244/#issuecomment-1344280205), I could open on issue but ideally someone who knows xgboost well could comment on whether it's really a bug or not. If there is no one around, I can just open no matter what.

adrinjalali: yeah I think the devs on the xgboost repo can decide if it's a bug which they wanna fix or not.

BenjaminBossan:
@adrinjalali I addressed your comments, please review again.

Regarding the issue of forgetting to add bytes/bytearray to trusted: It's a bit of a sign that our test coverage could be improved, right? As is, those oversights were not caught with the standard set of tests we have.

adrinjalali:
tests failing @BenjaminBossan

skops/io/tests/test_external.py
BenjaminBossan: https://github.com/dmlc/xgboost/issues/8596

BenjaminBossan:
@adrinjalali sorry, I made a list minute change and rolled it back, introducing the bug. Tests are green now.

skops/io/tests/test_external.py
```diff
@@ -0,0 +1,314 @@
...
```
adrinjalali, line 135:
```suggestion
    avoid testing these cases. See https://github.com/dmlc/xgboost/issues/8596
```

adrinjalali (commented):
Otherwise happy to merge.

Issue #241, opened 2022-12-05
CLI for inspection, and conversion to skops format, of sklearn pickle files 


BenjaminBossan:
It would be useful for skops to provide a CLI that allows to find the untrusted types in an sklearn pickle file, and to convert it to the skops format.

Details have yet to be discussed, but it could look something like this:

```bash
$ ls
my-model.pkl my-other-model.pkl
$ python -m skops inspect my-model.pkl
untrusted types: ...
$ python -m skops convert my-model.pkl
error: untrusted types ... found, pass trusted=true if you trust them or pass a list of explicitly trusted types
$ python -m skops convert my-model.pkl truste=true  # works
$ ls
my-model.pkl my-model.skops my-other-model.pkl
$ python -m skops convert *.pkl  # convert multiple models
$ ls
my-model.pkl my-model.skops my-other-model.pkl my-other-model.skops
```

merveenoyan:
It will simply be a script with argument parser? Is there anything else on top of this?

BenjaminBossan:
Maybe we could add a utility function to `skops.io` so that the conversion can also be called from Python, but apart from that, it should be really straightforward. Probably for this purpose, `argparse` would be sufficient.

adrinjalali:
We would also need to define an _entry point_:

- https://packaging.python.org/en/latest/specifications/entry-points/
- https://amir.rachum.com/blog/2017/07/28/python-entry-points/

E-Aho:
Picking this up :)

E-Aho:
So I'm working on this, and I want to double check I'm not missing something:

To check a pkl file is "trusted" or not, or convert it, we need to call `pkl.load()` on a pickled object, then construct a tree for it, right? 

Double checking I've not missed a method or API we already have for this.

BenjaminBossan:
Yes, it involves loading the pickled object. That's a good point. So the help of those commands should emphasize that this will happen and what it implies.

adrinjalali:
Right now the tree can only be constructed from a skops file, not a pickle file/object, and there isn't much of a point to try and check security once the object is loaded anyway. If the pickle file is compromised, the act of loading it runs arbitrary code.

So as far as security goes, we only tell users that loading arbitrary pickle files can run arbitrary code, and then we convert the file. Once the file is converted, we can tell them which types they'd need to pass as trusted to load the file. This is also what I did in this space: https://huggingface.co/spaces/adrin/pickle-to-skops

[...]

PR #249, opened 2022-12-09
FEAT CLI conversion

E-Aho:
Addresses #241 

This PR adds in an entrypoint, `skops convert`, which allows users to automatically convert files from the command line.

I didn't end up adding `inspect`, as it felt odd to me to inspect a .pkl file which would require loading it into `skops` format based on how we currently have the safety checks implemented. Happy to implement an entrypoint to do it if other maintainers think it would be useful, it would essentially just need to not save the output `skops` in the current conversion method.


BenjaminBossan:
No detailed review from me yet, just some more general questions to discuss:

1. It would be nice to give some examples how to actually use the CLI :)
2. I feel like using `logging` is overkill here, do we really need it? I have been burnt by the `logging` module in the past, so I would avoid it if possible :D. If `print` and `print(..., file=sys.stderr)` are not enough, we could consider using [rich](https://github.com/Textualize/rich). If we want to support different "log levels", we could have `action='count'` for `verbose` to allow `-v`, `-vv` etc. (at least that's what I think it does)
3. Usage-wise, I think allowing to (optionally) specify an output file name makes more sense than an output folder. That's more in line with the majority of CLI tools and gives the user more control.
4. Regarding the testing, I think I'm fine with the general approach. Just throwing out there that we could also create real pickle files and shell out to call the CLI, which is a bit more end-to-end.

E-Aho:
Ready for review @skops-dev/maintainers :)

E-Aho:
Addressing @BenjaminBossan's comments:

1) I definitely will include some examples in documentation, but for example:
```skops convert abc.pkl``` will convert the abc.pkl file to `abc.skops` at the moment.

2) I used logging as I felt it was the default, 'pythonic; way to allow varied verbosity of print statements to stdout. I thought it would be nice to allow for users to have slightly better visibility if something was not working, and it felt extensible to me. I haven't used rich much, but if that's something you and the other devs are happier with I'm happy to change to that.

3) I think it would make sense to add support for that too. The reason I initially wasn't too keen to go in that direction was it might make the usage pattern for multiple files clunkier.
`skops convert a.pkl b.pkl c.pkl --output-dir my-skops-dir` felt tidier than having to specify a file name/path for each file specifically, but we could also support different usage patterns for single vs batch file conversions.

4) Happy to add in some integration tests on actual files for the main entrypoint to have visibility of the whole workflow.

**Comment by BenjaminBossan** [comment]
My impression is that some of the things we discuss depend on what the general expectations of a command line tool are.

I've never seen the use of `logging` for CLI, but maybe that's just me. Still, I'd argue we should at least evaluate how necessary it is and if we can reduce complexity overall by removing it, we should do that. For instance, we could just not print anything when there is no error (which is the standard) and if users want to see more, like unknown packages, they can do `-v`.

(To be clear, I don't think we need rich, it's just in case we want to make the output prettier.)

Regarding output dir, we could keep it (but I wonder if `-t --target-directory` would be more standard, but then we need something else for trusted types). But IMHO we could also do without it since it's easy to chain commands to achieve the same result:

`$ skops convert *.pkl && mv *.skops my-target-dir/`

On the other hand, having a way to indicate the output file name is something I'd expect from similar tools, that's why I would have added it (but we could do that later).

Regarding the `--trusted` argument, do we want to allow more fine grained control, i.e. letting users indicate the trusted types? If so, the warning should exactly indicate how to specify them.

File: skops/io/_cli.py [hunk 1]
```diff
@@ -0,0 +1,113 @@
+from __future__ import annotations
```
adrinjalali, line 1: I think it'd make sense to move everything related to CLI to a `skops._cli` module and have them all in the same place. WDYT?

skops/io/_cli.py
```diff
@@ -0,0 +1,113 @@
...
+import os
+import pathlib
+import pickle as pkl
+from typing import Optional
+
+from skops.io import dumps, get_untrusted_types
+
+
+def _convert(
+    input_file: os.PathLike, output_dir: pathlib.Path, is_trusted: bool = False
```
adrinjalali, line 14: I think the output should be a file name, not a folder name. If the user wants to convert a bunch of files, for now they can write a bash script to do that.

Please also add a docstring.

skops/io/_cli.py
```diff
@@ -0,0 +1,113 @@
...
+
+    untrusted_types = get_untrusted_types(data=skops_dump)
+
+    if not untrusted_types:
+        logging.debug(f"No unsafe types found in {model_name}.")
+    else:
+        untrusted_str = "\n".join(untrusted_types)
+
+        logging.warning(
+            "Unsafe Types Detected!\n"
```
adrinjalali, line 32:
```suggestion
            "Unknown Types Detected!\n"
```

I think @BenjaminBossan agrees here.

skops/io/_cli.py
```diff
@@ -0,0 +1,113 @@
...
+            "the following unsafe types were found: \n"
+            f"{untrusted_str}\n"
+        )
+
+        if not is_trusted:
+            logging.warning(
+                f"Model {model_name} will not be converted due to unsafe types.\n"
+                "To convert this anyway, add `-t` to this command."
+            )
+            return
```
adrinjalali, line 43: I think we should always convert. The unsafe part is loading a pickle file, and we've already done it at this point. So there's no point in not converting. 

Instead, we can print a `skops.io.load` command which users need to use (with `trusted` arg) to successfully load a converted file.

adrinjalali (commented):
I don't think we need an empty `__init__.py` file, do we?

Nothing major here, looks pretty good to me. I don't have a strong opinion on print vs logging, but if we're using `logging`, we should certainly get an instance for this library, and under that a CLI one, instead of using the global instance.

skops/io/_cli.py
E-Aho: I agree with that, that makes sense to me. I'll refactor to add this

skops/io/_cli.py
E-Aho: That makes sense to me. Would that mean removing the `--trusted` flag entirely, and just printing a statement that unknown types were found, and `--trusted` would need to be passed in during load?

adrinjalali: yes!

E-Aho:
Addressed your PR comments, feel free to have another look @adrinjalali  :)

As it stands, users can pass in

I still want to add in documentation to cover the additions, I could maybe use someone pointing me in the right direction to write up changes in the docs folder as I've not got much experience writing documentation :')

skops/cli/_convert.py
```diff
@@ -0,0 +1,114 @@
...
+
+    parser.add_argument("inputs", nargs="+", help="Input files to convert.")
+    parser.add_argument(
+        "-o",
+        "--output-files",
+        nargs="+",
+        help=(
+            "Specify output file names for the converted skops files."
+            "If not provided, will default to using the same name as the input file, "
+            "and saving to the current working directory."
```
BenjaminBossan, line 75: We should add here that the suffix is changed, otherwise it may be interpreted as the original file being overwritten (if in same directory).

skops/cli/_convert.py
```diff
@@ -0,0 +1,114 @@
...
+    parser.add_argument(
+        "-d",
+        "--debug",
+        help="Enable debug logging.",
+        action="store_const",
+        dest="loglevel",
+        const=logging.DEBUG,
+        default=logging.WARNING,
+    )
+    parser.add_argument(
```
BenjaminBossan, line 88: Where is `verbose` used? Do we need `verbose` and `debug`?

skops/cli/tests/test_cli.py
```diff
@@ -0,0 +1,142 @@
...
+    def write_unsafe_file(self, pkl_path, unsafe_obj):
+        with open(pkl_path, "wb") as f:
+            pkl.dump(unsafe_obj, f)
+
+    def test_base_case_works_as_expected(
+        self, pkl_path, tmp_path, skops_path, write_safe_file, safe_obj
+    ):
+        cli._convert_file(pkl_path, skops_path)
+        persisted_obj = load(skops_path)
+        assert np.array_equal(persisted_obj, safe_obj)
```
BenjaminBossan, line 53: Should we check here that no log is created?

skops/cli/tests/test_entrypoint.py
```diff
@@ -0,0 +1,31 @@
...
+    """
+
+    @pytest.fixture(autouse=True)
+    def clear_argv(self):
+        # Required to clear argv in case Pytest is called on this specific function.
+        # Otherwise, clogs parser.parse_known_args() in argparse
+        sys.argv = [""]
+
+    @mock.patch("skops.cli._convert._convert_file")
+    def test_convert_works_as_expected(self, mocked_convert: mock.MagicMock):
```
BenjaminBossan, line 23: As this is meant to be more of an "end-to-end" test, how about testing it without mocking? Otherwise, I don't think it adds a lot of value, given the other tests, and could be removed.

BenjaminBossan (changes requested):
Looks quite good, but there still some points to discuss. (Interesting how many design decisions still have to be made for such a _seemingly simple_ task.)

> but if we're using logging, we should certainly get an instance for this library,

I think this has not been addressed. IIUC, Adrin means that we should have a logger instance, i.e. `logger = logging.getLogger(...)` and call `logger.info(...)` etc.

Personally, I still find it strange to use logging here, but if you two agree, it's fine by me.

> Users can pass in multiple input files and output file names

Really not sure if we need this or whether it makes our code more complicated then needs be? If users want to convert multiple files, there are already ways to achieve this in the command line, e.g. through globbing or a small bash script.

> I still want to add in documentation to cover the additions, I could maybe use someone pointing me in the right direction to write up changes in the docs folder as I've not got much experience writing documentation :')

I'm sure you can do it, read the existing docs for persistence and then try to put yourself in the shoes of a user, what information would be most important to them? The outcome doesn't have to be perfect on first try, we can iterate in this review.

Could you please also add an entry to `changes.rst`?

One more minor thing, when I call `skops --help`, I get this error:

```
$ skops --help
usage: Skops {convert}
Skops: error: the following arguments are required: function
```

adrinjalali:
On holidays, so I'll either review next year or happy for this to be merged, but one think which I think is important is for us to have the `cli` module as _private_ so that we can change things easy at least for now. I'd rename `skops.cli` to `skops._cli`.

BenjaminBossan:
I'll be on holiday too until next year, but will take a look from time to time when I have the opportunity.

setup.py
```diff
@@ -34,6 +34,14 @@
 
 
 def setup_package():
+    package_data = dict(
+        entry_points={
+            "console_scripts": [
+                "skops = skops.cli.entrypoint:main_entrypoint",
```
adrinjalali, line 40: Should we simply call it `entrypoint` instead of `main_entrypoint`?

skops/cli/__init__.py
```diff
@@ -0,0 +1,4 @@
+from ._convert import _convert_file, main_convert
+from .entrypoint import main_entrypoint
+
+__all__ = ["_convert_file", "main_convert", "main_entrypoint"]
```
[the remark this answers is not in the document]
adrinjalali: I think if we make `cli` folder a private module (aka `_cli`), we don't need any init files.

skops/cli/_convert.py
```diff
@@ -0,0 +1,114 @@
...
+    command_line_args: Optional[list[str]] = None,
+    parent: Optional[argparse.ArgumentParser] = None,
+):
+    parents = [parent] if parent else []
+
+    parser = argparse.ArgumentParser(
+        description="Convert input Pickle files to .skops files", parents=parents
+    )
+
+    parser.add_argument("inputs", nargs="+", help="Input files to convert.")
```
adrinjalali, line 67: I think starting with a single file would make more sense and makes things much easier. We also don't have to deal with matching number of inputs/outputs etc. We can let a shell script handle multiple files.

adrinjalali (commented):
thanks @E-Aho 

A rather bigger question to me is, should we run our tests the same way we do in our test suite on the generated file here? It would be probably nice to tell users of potential issues if the persistence has gone wrong, but silently. WDYT? I've been thinking of making that part of the test suite public. But @BenjaminBossan would have a better idea of how stable that part is.

Also, empty init files shouldn't be needed.

BenjaminBossan:
> A rather bigger question to me is, should we run our tests the same way we do in our test suite on the generated file here? It would be probably nice to tell users of potential issues if the persistence has gone wrong, but silently. WDYT? I've been thinking of making that part of the test suite public. But @BenjaminBossan would have a better idea of how stable that part is.

Sorry, I don't understand what you're referring to here. Could you please explain?

adrinjalali:
Something like this:

```py
    loaded = loads(dumps(estimator), trusted=True)
    assert_params_equal(estimator.get_params(), loaded.get_params())
```

To have the above test included in the convert code so that we warn the users if something doesn't look right.

BenjaminBossan:
Gotcha, interesting idea. I'd say we can add this in a separate PR after merging this one, maybe put it behind an optional argument.

For this to be really useful, I think we would need to work a bit on `assert_params_equal`:

1. Since it would basically be exposed to the user, we might have to add tests for this test function.
2. Right now, it fails on the first attribute that differs, but it would be more useful if it could collect all failing attributes (similar to the auditing code). This is, however, not trivial, because it would require us to track the attributes through meta estimators etc.

adrinjalali:
Agree. Happy to have it in a separate PR then.

skops/cli/_convert.py
```diff
@@ -0,0 +1,114 @@
...
+
+    untrusted_types = get_untrusted_types(data=skops_dump)
+
+    if not untrusted_types:
+        logging.debug(f"No unsafe types found in {model_name}.")
+    else:
+        untrusted_str = "\n".join(untrusted_types)
+
+        logging.warning(
+            "Unknown Types Detected!\n"
```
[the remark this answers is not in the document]
E-Aho: Happy to remove, I personally lke line breaks in logs, feel like it makes it slightly more human-readible (especially if it's going to stdout), but happy to move it to single line.

skops/cli/_convert.py
```diff
@@ -0,0 +1,114 @@
...
+            f"While converting {model_name}, "
+            "the following unsafe types were found: \n"
+            f"{untrusted_str}\n\n"
+            f"When loading{output_file}, add ``--trusted`` to the skops.load call."
```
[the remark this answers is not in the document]
E-Aho: Can remove the line, although I feel something should be added to make sure the user knows how to load the model that gets created here

skops/cli/_convert.py
E-Aho: The normal pattern I've seen used before with varying log levels is to tie `--verbose` to `info` level logging, and `--debug` to debug level logging.

So in the current code, with the `-v` flag, the info, debug and above level statements will be written out, and with `-d`, only debug and above level logs are written out.

(The info level logging is really fine grained: it logs when a model starts, and finishes being converted. Adjusting the log level here also lets any logging used elsewhere in our codebase get handled correctly.)

skops/cli/tests/test_entrypoint.py
E-Aho: Sure, happy to do it that way

When I've written CLIs before in other projects, the standard pattern was to write unit tests for the functionality (here: `_convert_file`) and also unit tests for the interface itself (here: `entrypoint`, and checking it interfaces correctly with `_convert_file`)

But for a simple entrypoint, the happy path test is pretty simple, so happy to rework this one as an "end-to-end" test for the full functionality.

setup.py
E-Aho: We can, I named it that way so it didn't clash with the file name

`entrypoint.entrypoint` feels a little clunky to me, happy for other suggestions or to go with that if anyone feels strongly

skops/cli/_convert.py
E-Aho: I think that would be ok, though I think there should still be support for converting batches of files at a time. 

I can rework this PR to work for single files only, but I do feel we should consider ways to support it.

E-Aho:
I have been playing with this a little bit more, and I need to do a refactor of the `main_entrypoint` to make use of subparsers instead of child parsers. 

Because of the way it's currently written, it is very hard to allow both `skops --help` and `skops convert --help` to work as needed. The parent parser will see `--help` wherever it is, and parse it automatically. I think there are hacky ways around this, but it makes more sense to migrate to using the proper pythonic way to handle this.

skops/cli/_convert.py
BenjaminBossan: Ah I see how it's used now.

> The normal pattern I've seen used before with varying log levels is to tie --verbose to info level logging, and --debug to debug level logging.

Which libraries do that? Could you please give me a pointer?

> So in the current code, with the -v flag, the info, debug and above level statements will be written out, and with -d, only debug and above level logs are written out.

INFO level is 20, DEBUG level is 10, so when setting `-v`, debug is not shown, and when setting `-d`, debug and info are shown. 

Furthermore, setting both `-v` and `-d` sets the log level to INFO, which as a user, I would find confusing. That's why I would prefer having a single argument be in control of the granularity of the logging. That could be a `loglevel` argument, where users can pass `INFO` or `DEBUG` or an int. Or it could be an incremental `verbose` argument, where  users can pass `-v`, `-vv`, etc.

> Adjusting the log level here also lets any logging used elsewhere in our codebase get handled correctly

I don't understand this point.

skops/cli/tests/test_entrypoint.py
BenjaminBossan: I see the idea here, but logically, what bug could be prevented here that is not already covered by the previous tests? In contrast, when not mocking, there could be a case that this tests that everything is glued together correctly. But that's not very likely to be broken, which is why I would also be fine with removing this test, given the other ones above.

setup.py
BenjaminBossan: I don't mind either way.

skops/cli/_convert.py
BenjaminBossan: Let's start with single files. We can maybe document how to convert multiple files to make it easier for users to achieve that. Then, when we run into a case where batches would be really useful, we can add them later.

setup.py
adrinjalali: maybe `parent_cli` or `main_cli`?

BenjaminBossan (commented):
I commented on some of your comments. On top, regarding this issue:

> I think there are hacky ways around this, but it makes more sense to migrate to using the proper pythonic way to handle this.

If you find that `argparse` (or other stdlib cli packages) becomes too limited for the use case, we should seriously consider using a third party library like `click` instead. It's nice to have fewer dependencies, but not at the cost of a considerable increase in complexity or a buggier code base.

skops/cli/_convert.py
E-Aho: > INFO level is 20, DEBUG level is 10, so when setting -v, debug is not shown, and when setting -d, debug and info are shown.

I always get that mixed up, sorry. I can shift the info and debug level logs.

The usual pattern I've seen used is described [here in the Ubuntu community docs](https://discourse.ubuntu.com/t/cli-verbosity-levels/26973). Here you can see varied level outputs from quiet -> trace, but essentially:

* Verbose logs should include more detailed descriptions
* Debug logs should describe internal steps as they happen to help in debugging

And verbose typically uses `-v` or `--verbose` as flags, whereas debug uses `-d` or `--debug`.

However, this is for Linux CLIs, and from what I can tell, many python packages tend to prefer the incremental `verbose` pattern you described.

> Adjusting the log level here also lets any logging used elsewhere in our codebase get handled correctly

Basically I just meant we can add more logging in the future without having to manually add logic to handle it. 

With log levels, we can just add `logging.debug(...)` in the skops.io module and given the right log level, that will be output in the CLI without needing any new logic.

skops/cli/tests/test_entrypoint.py
E-Aho: I think it would be good to have a full integration test of some sort, that might be a good first contributor issue we can raise after this PR.

E-Aho:
Ok, I've done the following:

* Reworked ArgParse to work as we want, help now works for both `skops` and `skops convert`  :)
* Reworked to support single files only
* Removed uneeded tests
* Updated some of the CL text

Still to do:

* Add documentation and examples 
* Update Change.rst
* Consider changing from `-v` and `--debug` to `-v` and `-vv` for logging verbosity

BenjaminBossan:
> Still to do:

Let us know if you already want a review of the current state or if we should wait for the remaining changes.

skops/cli/_convert.py
BenjaminBossan: > The usual pattern I've seen used is described [here in the Ubuntu community docs](https://discourse.ubuntu.com/t/cli-verbosity-levels/26973). Here you can see varied level outputs from quiet -> trace, but essentially:

I found this to be unclear. In the table, it appears that the different options describe an increasing *scale* of verbosity. In that case, I would argue for a single option that controls verbosity, since we're talking about changing a single value.

On the other hand, in the text it says:

> In verbose mode, you want to talk to the audiences in a more descriptive and thorough description about what happens in transactions or a series of events. This could be a conversation log between a user/client and the machine. [...]
> In debug mode, the mental model is to think of 2 developers talking to one another trying to understand where things go wrong. [...]

This reads like different *types* of output, one designed for users, one for devs, not something on a scale. If this was the goal, I can see having two different options, `verbose` for users and `debug` for devs, which can be turned on/off independently and don't overlap. But this would *not* map onto `logging.INFO` and `logging.DEBUG`, since those are on a scale.

Can we maybe not simplify the CLI for now by only providing a default mode, which is to not print anything, except if an error occurs, and a `verbose` mode, which prints info level stuff (but not debug)?

> Basically I just meant we can add more logging in the future without having to manually add logic to handle it.

Got it.

E-Aho:
> Let us know if you already want a review of the current state or if we should wait for the remaining changes.

I think I should be able to make the above changes in the next day or so, so you can hold off on reviewing until then :) I'll ping you once it's ready for review

skops/cli/_convert.py
E-Aho: > Can we maybe not simplify the CLI for now by only providing a default mode, which is to not print anything, except if an error occurs, and a verbose mode, which prints info level stuff (but not debug)?

Happy to do this. 

> This reads like different types of output, one designed for users, one for devs, not something on a scale. If this was the goal, I can see having two different options, verbose for users and debug for devs, which can be turned on/off independently and don't overlap. But this would not map onto logging.INFO and logging.DEBUG, since those are on a scale.

I think *broadly* they do exist on a scale (at least in Ubuntu-land), where "verbose" means "log things up to INFO level" and "debug" means "log things up to DEBUG level", as well as the changes that can be seen in the table on that page of what get logged where. The way I read that description in the text was more about what logs at INFO and DEBUG level should look like.

I did more digging, however, and have seen that in [Android logging](https://developer.android.com/studio/debug/am-logcat), "verbose" means "write out ALL level logs", so am starting to think the "canonical" way to tell a CLI what log level just depends on who you ask.

Happy to go to a much simpler and just go with number of "v"s to adjust log level.

E-Aho:
Ok, phew, should now be ready to review again @skops-dev/maintainers  :)

A couple of notes:

* Current behaviour is to overwrite when writing a file if it already exists, is this the expected behaviour?
* I reintroduced the empty`__init__.py` in `cli` because Pytest was failing to initilaise because there is now an `_utils` file in `cli`
* I reintroduced the `test_entrypoint` test because of the new added complexity with parsers and subparsers
* I didn't create a full integration test for converting a file from the command line, might be a good test to add for a new contributor, or something I can add at a later date
* Let me know if I should add more/ reword the documentation at all, wasn't sure how much I should add there.

skops/cli/_utils.py
```diff
@@ -0,0 +1,15 @@
+import logging
+
+
+def get_log_level(level: int = 0):
```
BenjaminBossan, line 4:
```suggestion
def get_log_level(level: int = 0) -> int:
```

skops/cli/entrypoint.py
```diff
@@ -0,0 +1,44 @@
+import argparse
+
+import skops.cli._convert
+
+
+def main_cli(command_line_args=None):
+    """
+    Main command line interface entrypoint for all command line Skops methods.
```
BenjaminBossan, line 8:
```suggestion
    """Main command line interface entrypoint for all command line Skops methods.
```

skops/cli/entrypoint.py
```diff
@@ -0,0 +1,44 @@
...
+        },
+    }
+
+    for func_name, values in function_map.items():
+        subparser = subparsers.add_parser(func_name)
+        subparser.set_defaults(func=values["method"])
+        values["format_parser"](subparser)
+
+    args = entry_parser.parse_args(command_line_args)
+    args.func(args)
```
BenjaminBossan, line 44: Could you please add a comment on what this line does?

skops/cli/_utils.py
```diff
@@ -0,0 +1,15 @@
...
+    """Takes in verbosity from a CLI entrypoint (number of times -v specified),
+    and sets the logger to the required log level"""
+
+    all_levels = [logging.WARNING, logging.INFO, logging.DEBUG]
```
BenjaminBossan, line 8: IIUC, warning level logs would always be shown, even if I don't add `-v`. IMO, this should not be the case. So I would prefer that without `-v`, the log level should be `ERROR` or `CRITICAL`.

skops/cli/tests/test_convert.py
```diff
@@ -0,0 +1,138 @@
+import logging
+import pathlib
+import pickle as pkl
```
BenjaminBossan, line 3:
```suggestion
import pickle
```

for consistency with the rest of the code base.

skops/cli/tests/test_convert.py
```diff
@@ -0,0 +1,138 @@
...
+        with open(pkl_path, "wb") as f:
+            pkl.dump(unsafe_obj, f)
+
+    def test_base_case_works_as_expected(
+        self, pkl_path, tmp_path, skops_path, write_safe_file, safe_obj, caplog
+    ):
+        _convert._convert_file(pkl_path, skops_path)
+        persisted_obj = load(skops_path)
+        assert np.array_equal(persisted_obj, safe_obj)
+        assert MockUnsafeType.__name__ not in caplog.text
```
BenjaminBossan, line 53: I'm a bit worried about this specific test, because it can easily pass because of some change in the default log level. Not sure what the best way would be to avoid this, but one way would be to use dependency injection. That would mean that we pass a `logger` object to `_convert_file`, which by default is `logging.Logger()`, but in the test can be, e.g., a mock. Then we can run this check on the mock logger, instead of relying on `caplog`.

skops/cli/tests/test_convert.py
```diff
@@ -0,0 +1,138 @@
...
+        args = [input_path, "--output", output_path, verbosity.split()]
+
+        namespace, _ = _convert.format_parser().parse_known_args(args)
+
+        _convert.main(namespace)
+        self.assert_called_correctly(
+            mock_convert, path=input_path, output_file=output_path
+        )
+
+        assert caplog.at_level(expected_level)
```
BenjaminBossan, line 138: TIL

skops/cli/tests/test_convert.py
```diff
@@ -0,0 +1,138 @@
...
+        namespace, _ = _convert.format_parser().parse_known_args([path])
+
+        _convert.main(namespace)
+        self.assert_called_correctly(mock_convert, path)
+
+    @mock.patch("skops.cli._convert._convert_file")
+    @pytest.mark.parametrize(
+        "input_path, output_file, expected_path",
+        [
+            ("abc.123", "a/b/c", "a/b/c"),
```
BenjaminBossan, line 91: I don't understand, why is it `a/b/c`?

BenjaminBossan (changes requested):
This looks pretty good, I tested it locally and it works as expected. Thanks also for adding to the docs.

Btw. I checked and it seems we already have a transient dependency on `click`, so it would theoretically not cost us anything to build our CLI with it. I assume that after all this work, you're not interested in rewriting it with `click`, which is fine, I just wanted to mention it.

There still a few points left that I commented on, please take a look. Also, I wonder about this comment:

> > but if we're using logging, we should certainly get an instance for this library,

> I think this has not been addressed. IIUC, Adrin means that we should have a logger instance, i.e. logger = logging.getLogger(...) and call logger.info(...) etc.

docs/persistence.rst
```diff
@@ -110,6 +110,22 @@ you have custom functions (say, a custom function to be used with
...
 objects having references to functions such as ``numpy.sqrt``.
 
+Command Line Interface
+########
+Skops has a command line interface to convert SciKit-Learn models persisted with
+``Pickle`` to ``Skops`` files.
+
+To convert a file from the command line, use the ``skops convert`` entrypoint.
+
+Below is an example call to convert a file ``my_model.pkl`` to ``new_model.skops``:
```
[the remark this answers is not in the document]
E-Aho: Yep, I was going back and forth on whether this example should be as simple as possible or to include renaming, forgot to change the doc line

skops/cli/_convert.py
```diff
@@ -0,0 +1,109 @@
...
+def _convert_file(input_file: os.PathLike, output_file: os.PathLike):
+    """
+    Function that is called by ``skops convert`` entrypoint.
+
+    Loads a pickle model from the input path, converts to skops format, and saves to
+    output file.
+
+    Parameters
+    ----------
+    input_file : os.PathLike
```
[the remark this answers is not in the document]
E-Aho: I honestly think `os.PathLike` is pretty self-documenting, but I could also see a user getting confused by not seeing the type they're expecting in the docstring. 

I'm happy with either way

skops/cli/_convert.py
```diff
@@ -0,0 +1,109 @@
...
+
+    untrusted_types = get_untrusted_types(data=skops_dump)
+
+    if not untrusted_types:
+        logging.info(f"No unknown types found in {model_name}.")
+    else:
+        untrusted_str = ", ".join(untrusted_types)
+
+        logging.warning(
+            "Unknown Types Detected! "
```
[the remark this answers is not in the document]
E-Aho: Probably not, though I thought it might be useful to see an immediate descriptor at the start of the log, particularly if it's a long log. Will remove though.

skops/cli/_convert.py
```diff
@@ -0,0 +1,109 @@
...
+            f"While converting {model_name}, "
+            "the following unknown types were found: "
+            f"{untrusted_str}. "
+            f"When loading {output_file}, add 'trusted=True' to the skops.load call. "
```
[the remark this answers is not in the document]
E-Aho: I think from a users' perspective, it might be nicer to give them the easiest fix possible. 

When they're converting this file, and want to not deal, the log line above would've told them all of the untrusted types anyway, so if they wanted to go that route, they could just add those.

We could add a line along the lines of:
`..., or set each of the unknown types listed previously to "trusted".`

skops/cli/_utils.py
E-Aho: From the python docs:

```The default level is WARNING, which means that only events of this level and above will be tracked, unless the logging package is configured to do otherwise.```

I followed this because it's the default for python packages. 

We can adjust the default level for our CLI, or we could also add a `-q, --quiet` option to reduce the log level

File: skops/cli/tests/test_convert.py (hunk 21 above)
Comment (E-Aho) [reply skops/cli/tests/test_convert.py:91]: Ah, the test is a little hard to read, I can adjust that

`a/b/c` is the name of the user defined "output_file" that is supplied to the CLI, I'll rework that so it looks less like the input file name.

File: skops/cli/tests/test_convert.py (hunk 19 above)
Comment (E-Aho) [reply skops/cli/tests/test_convert.py:53]: Happy to rework to use dependency injection, I think it makes sense to keep things more testable.

**Comment by E-Aho** [comment]
Thanks for the review @BenjaminBossan, I've made adjustments along the lines of most of your comments, let me know if there's anything I've missed.

WRT logging, I've moved to using a ``logger`` instance, but still set a basic config at the top level for our selected entrypoint, as I felt that the config should be set at the top level for whatever entrypoint we're using.

That way, if we do implement any different logging elsewhere in our codebase, it automatically keeps the config specified for whatever command we're calling. 

Let me know if you think there's a better way to achieve this; 

Also, the log format currently looks like:

```INFO    : No unknown types found in abc.```

How does this look to all? We can add things like time, function call, etc at a later date if we need it.

Also, when we implement logging somewhere else, might make sense to move this config out to a common module.

File: skops/cli/_convert.py (hunk 23 above)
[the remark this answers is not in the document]
Comment (BenjaminBossan) [reply skops/cli/_convert.py:26]: Hard to say if a user who doesn't know the type would guess that `str` is a path-like. Anyway, I won't object to leaving it as is, especially since it means that the type annotation and docstring are identical, I just wanted to bring up the possible friction here.

File: skops/cli/_utils.py (hunk 17 above)
Comment (BenjaminBossan) [reply skops/cli/_utils.py:8]: I agree this is true for logging in general, but for CLIs, there is the convention to not print anything if there is no issue. Ideally, I would like to avoid having a `--quiet` option, because it increases complexity and because it's unclear how it interacts with `-v`.

File: skops/cli/tests/test_convert.py (hunk 19 above)
Comment (BenjaminBossan) [reply skops/cli/tests/test_convert.py]: Talked to a friend about using DI in this case. Although I think it's not a bad use of DI, it might be a bit unorthodox when it comes to logging, so if you feel this makes the code harder to maintain, feel free to ignore the suggestion.

**Review by BenjaminBossan — COMMENTED** [review]
Thanks for implementing the changes, this looks great to me now. Let's wait for one more review before merging.

> WRT logging, I've moved to using a logger instance, but still set a basic config at the top level for our selected entrypoint, as I felt that the config should be set at the top level for whatever entrypoint we're using.

> Also, when we implement logging somewhere else, might make sense to move this formatting out to a common module.

I think it's good for the time being.

> Also, the log format currently looks like:
> INFO    : No unknown types found in abc.
> How does this look to all? We can add things like time, function call, etc at a later date if we need it.

Yes, let's keep it simple for now.

File: docs/persistence.rst [hunk 26]
```diff
@@ -110,6 +110,31 @@ you have custom functions (say, a custom function to be used with
...
+.. code:: console
+
+    skops convert my_model.pkl
+
+To convert multiple files, you can use bash commands to iterate the above call.
+For example, to convert all ``.pkl`` flies in the current directory:
+
+.. code:: console
+
+    for FILE in *.pkl; do skops convert FILE; done
```
BenjaminBossan, line 132: Just noticed, wouldn't `skops convert *.pkl` do the same?

E-Aho: I played around with that syntax, and I think it essentially calls `skops convert file_1.pkl file_2.pkl`, not `skops convert file_1.pkl & skops convert file_2.pkl`

The functionality to do `skops convert file_1.pkl file_2.pkl` was removed following the earlier comments on this PR, but I think we could consider the right way to handle multiple input files at a single time

skops/cli/_convert.py
```diff
@@ -0,0 +1,111 @@
...
+from typing import Optional
+
+from skops.cli._utils import get_log_level
+from skops.io import dumps, get_untrusted_types
+
+
+def _convert_file(
+    input_file: os.PathLike,
+    output_file: os.PathLike,
+    logger: logging.Logger = logging.getLogger(),
```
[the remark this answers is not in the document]
E-Aho: Hmm, we could, although that would mean it gets initialised even if we import functions from the file to use elsewhere.

I don't think it would really impact much, I just personally dislike writing anything other than definitions at the top level

skops/cli/_utils.py
E-Aho: Hmm, good point.

I would raise that it *could* be considered an issue if untrusted types were present.

If a user had a file with untrusted types in it, and we converted it silently, they wouldn't know they needed to specify anything as trusted when using it.

docs/persistence.rst
BenjaminBossan: Ah yes, you're right, I misremembered how globbing is resolved (also not sure if Windows works the same as unix with globbing).

It's unfortunate that this doesn't work. I tried to find the minimal change required to make it work and came to this:

```python
# _convert.py
# l 69
- parser_subgroup.add_argument("input", help="Path to an input file to convert. ")
+ parser_subgroup.add_argument("input", help="Path to an input file to convert. ", nargs='+')

# l 95, replace main with
def main(
    parsed_args: argparse.Namespace,
) -> None:
    input_files = parsed_args.input
    output_file = parsed_args.output_file
    if len(input_files) > 1 and output_file:
        raise ValueError("TODO")

    for input_file in input_files:
        logging.basicConfig(format="%(message)s", level=get_log_level(parsed_args.loglevel))

        if not output_file:
            # No filename provided, defaulting to base file path
            file_name = pathlib.Path(input_file).stem
            output_file = pathlib.Path.cwd() / f"{file_name}.skops"

        _convert_file(
            input_file=input_file,
            output_file=output_file,
        )
        output_file = None
```

So basically the same as the old main function, but putting it inside of a loop.

I know this is a bit of a rollback of what we said earlier of not supporting multiple input files, but it is a simplified implementation of the earlier version of supporting multiple files. That said, I would also be totally fine with keeping it as is and using a loop in bash.

skops/cli/_convert.py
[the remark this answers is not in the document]
BenjaminBossan: > that would mean it gets initialised even if we import functions from the file to use elsewhere.

That's true. On the other hand it's now being initialized each time the function is called. Given that CLI usage consists of creating a new process anyway, it makes indeed no difference either way. If we ever get to a point where we use the logger elsewhere, we might have to rethink this.

skops/cli/_convert.py
```diff
@@ -0,0 +1,111 @@
...
+
+    untrusted_types = get_untrusted_types(data=skops_dump)
+
+    if not untrusted_types:
+        logger.info(f"No unknown types found in {model_name}.")
+    else:
+        untrusted_str = ", ".join(untrusted_types)
+
+        logger.warning(
+            f"While converting {model_name}, "
```
BenjaminBossan, line 49:
```suggestion
            f"While converting {input_file}, "
```

Arguably this is a bit more precise?

skops/cli/_utils.py
BenjaminBossan: IMO, this is more of an issue for the person loading the file, and they will know (unless they use `trusted=True`). Personally, I would not report it by default, but I can live with the current way.

docs/persistence.rst
E-Aho: Yep, similar to what I had before, but I think the complexity comes from that "ValueError" line

The UX for the user is the complex bit there, my previous implementation would basically allow the user to pass in multiple output files too, and resort to the `if not output_file:` logic for the last files if there was a mismatch between the number of input and output files.

Happy to work this back in, but I think this PR should probably stay to a single file implementation (Adrin seemed to want to stick to this for this PR) and discuss this further later :)

BenjaminBossan: Yeah, let's keep this in mind for a possible future PR.

> my previous implementation would basically allow the user to pass in multiple output files too

I think that most of the complexity came from exactly that, since it required dealing with num_inputs != num_outputs, while here, we just don't allow specifying outputs for more than 1 input, which is less user friendly but reduces complexity.

E-Aho:
@BenjaminBossan do we still want one more review before merging this in?

BenjaminBossan (approved):
Thanks so much for the hard work. This LGTM and can be merged if @adrinjalali is also fine with it.

As mentioned in some of the comments, there are some things that I think can be improved, but they are not blocking this PR, as they can be added later without BC breakage.

adrinjalali (approved):
Thanks @E-Aho . I think we might need improve docs when we get back to this, but happy to merge this and have it out.

PR #257, opened 2022-12-16
FEAT Parse model cards with pandoc

BenjaminBossan:
As discussed here:

https://github.com/skops-dev/skops/issues/72#issuecomment-1327360660

_For maintainers: No need to review anything, I just updated this branch after merging the alternative model card implementation and pointed the PR to the main repo instead of my fork._

## Description

This feature adds a new function, `skops.card.parse_modelcard`. When passing it the path to a dumped model card, it parses it using pandoc and returns a `Card` object, which can be further modified by the user.

Example:

```python
numpy as np
sklearn.linear_model import LinearRegression
skops.card import Card

X = np.array([[1, 1], [1, 2], [2, 2], [2, 3]])
y = np.dot(X, np.array([1, 2])) + 3
regr = LinearRegression().fit(X, y)
card = Card(regr)
card.save("README.md")

# later, load the card again
skops.card import parse_card

parsed_card = parse_modelcard("README.md")

# continue editing the card
parsed_card.add(**{"My new section": "My new content"})

# overwrite old card with new one
parsed_card.save("README.md")
```

There is a test that shows that cards with tables and plots are already possible to be parsed, reproducing the original content 1:1, with the exception of line breaks. The latter should generally be conserved, but there are a few cases where there is one too many, but IMO it doesn't matter.

### Comment

In the end, this turned out easier than I initially thought it would. The main difficulty are the data structures returned by the pandoc parser, for which I couldn't find any documentation. I guess Haskell code is just so self-documenting it's not needed.

For this reason (and out of laziness), there are probably quite a few edge cases that I haven't covered yet. Just as an example, when parsing tables, pandoc tells us how the columns are aligned. This information is currently completely discarded (we let tabulate choose the alignment). If we want to preserve the table alignment, we would need to make some changes.

I think, however, that it's fine not to support everything out of the box. We can incrementally add support for the rest as we collect more experience.

## Implementation

This feature requires the alternative card implementation from #203

pandoc is used for the following reasons:

- widely used and thus battle tested
- can read many other formats, not just markdown, so in theory, we should be able to read, e.g., rst model cards without modifying any code

The disadvantage is that pandoc is not a Python package, so users need to install it separately. But it is available on all common platforms.

For calling pandoc, I chose to shell out using `subprocess`. I think this should be fine but LMK if there is a better way.

There is a Python package that binds pandoc (https://github.com/boisgera/pandoc) but I don't think it's worth it for us to add it, just to avoid shelling out. The package seems to have low adoption and contains a bunch of stuff we don't need.

I chose to implement this feature such that the parser that generates the `Card` object should not have to know anything about markdown. Everything related to markdown is moved to a separate class in `_markup.py` which could be extended in the future to support other markup languages.

In an ideal world, we would not have to know anything about markdown at all. Instead the `Card` object should have methods (similar to what we already have for `add_table` etc.) that handles all special cases in a markup-language agnostic way. But in practice, this is far from being true. E.g. if a user wants to add bold text, there is no special method for it, so they would need to add raw markdown. The `Card` class is thus a leaky abstraction, which requires the parsing to directly deal with markdown.

## TODOs

This PR is not finished. Remaining TODOs that come to mind:

1. ~Documentation has to be updated in several places~
2. Possibly add more unit tests -- right now, it is only checked that parsed and original card markdown are (almost) identical
3. ~CI needs to install pandoc so that the tests are actually run~ pandoc is installed in CI for the ubuntu built, not windows or mac -- IMO that's enough.
4. ~There are some specifics here that won't work with all Python versions, like the use of `TypedDict`~
5. ~Clean up a couple of TODO comments, docstrings, and asserts~
6. Do we want a `card.load` method that uses pandoc parsing?


BenjaminBossan:
@skops-dev/maintainers I think this is now ready for review. For the full description, please look at the original comment.

More as a meta comment, I know this will probably not be an easy review. However, I think since this is more of a "side feature" and doesn't touch the skops "core", it is okay if it isn't perfect. There should be no problem making changes later on without breaking BC, e.g. by supporting a bigger subset of markdown features.

The testing is done in a kinda simple fashion, where I take a model card markdown, parse it, dump it again into markdown, then compare the diff to the original card. If the diff corresponds to the expected diff, the test passes. The test module should hopefully contain enough comments to explain these steps.

.github/workflows/build-test.yml
```diff
@@ -61,6 +61,9 @@ jobs:
           then pip install --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple scikit-learn;
           else pip install "scikit-learn~=${{ matrix.sklearn_version }}";
         fi
+        if [ ${{ matrix.os }} == "ubuntu-latest" ];
+          then wget -q https://github.com/jgm/pandoc/releases/download/2.19.2/pandoc-2.19.2-1-amd64.deb && sudo dpkg -i pandoc-2.19.2-1-amd64.deb;
```
adrinjalali, line 65: oh oh, I would have been happier if we could simply install it from default repos.

docs/model_card.rst
```diff
@@ -103,3 +109,43 @@ using :meth:`Card.select`, and you can delete sections using
...
+`parse_modelcard docs
+<https://skops.readthedocs.io/en/stable/modules/classes.html#skops.card.parse_modelcard>`_
+
+For the parsing part, we rely on `pandoc <https://pandoc.org/>`_. If you haven't
+installed it, please follow `these instructions
+<https://pandoc.org/installing.html>`_. The advantage of using pandoc is that
+it's a very mature library and that it supports many different document formats.
+Therefore, it should be possible to parse model cards even if they use a format
+that's not markdown, for instance reStructuredText, org, or asciidoc. For
+saving, we only support markdown, though.
```
adrinjalali, line 151:
```suggestion
saving, we only support markdown for now.
```

skops/card/_markup.py
```diff
@@ -0,0 +1,324 @@
...
+    markdown figure, or ``_raw_block``, to transform a pandoc ``RawBlock``.
+
+    From the caller side, only the ``__call__`` method should be used, the rest
+    should be considered internals.
+
+    """
+
+    def __init__(self):
+        # markdown syntax dispatch table
+        self.mapping = {
```
adrinjalali, line 39: this is bloody comprehensive!

skops/card/_parser.py
```diff
@@ -0,0 +1,353 @@
...
+        return cur_section
+
+    def _add_content(self, content: str, section: Section | None) -> None:
+        # Add content to the current section
+        if section is None:
+            # This may occur if the model card starts without a section. This is
+            # not illegal in markdown, but we don't handle it yet.
+            raise ValueError(
+                "Trying to add content but there is no current section, "
+                "this is probably a bug, please open an issue on GitHub"
```
adrinjalali, line 78:
```suggestion
                "this is probably a bug, please open an issue on GitHub."
```

not sure why you like to miss the dots at the end of exception messages :D

adrinjalali (commented):
Just a few nits, otherwise I think this is good to go! Awesome work!

.github/workflows/build-test.yml
BenjaminBossan: I followed the official install instruction: https://pandoc.org/installing.html#linux. If you have another suggestion, LMK.

docs/model_card.rst
BenjaminBossan: done

skops/card/_parser.py
BenjaminBossan: Not sure, probably because Python error messages also omit the period when it's only one sentence?

skops/card/_parser.py
```diff
@@ -0,0 +1,353 @@
...
+
+    _, _, actual_version = version_info.partition(" ")
+    pandoc_version = [int(v) for v in actual_version.split(".")]
+    return pandoc_version
+
+
+def _check_version_greater_equal(
+    version: Sequence[int], min_version: Sequence[int]
+) -> None:
+    """Very bad version comparison function to ensure that the first version is
```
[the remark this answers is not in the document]
BenjaminBossan: We had a similar discussion in skorch: https://github.com/skorch-dev/skorch/issues/887

We decided not to use `packaging` because we didn't want to have a dependency on it. It seems that in skops, it is available (probably a transient dependency), but it's not an explicit requirement. If you are okay with explicitly depending on `packaging`, I can make the change.

BenjaminBossan (commented):
@adrinjalali I addressed your comments, pls review again.

skops/card/_parser.py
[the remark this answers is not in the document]
adrinjalali: setuptools depends on it, I don't think `packaging` is a heavy dependency, I'm happy to include it as an explicit dep.

BenjaminBossan:
@adrinjalali I went with your `apt install` suggestion, but that has a very old version of pandoc (2 years), which creates problems with the parse results. I could try to make it work, but honestly I don't think it's worth the effort. There might be ppa's with newer pandoc versions, but at that point we could as well revert back to install from github releases, as is recommended, no?

(btw, yesterday pandoc 3.0 was released, I hope there is no breakage D:)

BenjaminBossan:
> So what I'm concerned about, is that when users are recommended to install a `.deb` file, then the package doesn't get upgraded. That's why installing from a `ppa` would be much preferred.

Yes, I totally agree. However, I feel this is more of an issue with pandoc distribution in general and not so much for us. For CI, I think we can live with the deb solution.

We could consider recommending a different way of installing pandoc in our docs, but I'm not sure if we should deviate from the officially recommended method.

adrinjalali (approved):
WDYT of putting `# pragma: no cover` for lines which are not supposed to be covered?

Otherwise LGTM.

skops/card/_markup.py
```diff
@@ -0,0 +1,338 @@
...
+
+        if not typef.startswith("fig:"):
+            # not sure if this can be reached, just to be safe
+            raise ValueError(f"Cannot deal with figure of type '{typef}'")
+
+        caption = "".join(self.__call__(i) for i in caption)
+        content = f"![{caption}]({dest})"
+        return content
+
+    def _figure(self, value) -> str:
```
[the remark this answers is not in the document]
BenjaminBossan: Yep, `Figure` is a new type that was added in pandoc 3.0. I didn't want to switch to 3.0 for CI because it is freshly released, so probably almost no one is using it. And having pandoc 2.x + 3.x in the CI matrix seemed like overkill to me.

BenjaminBossan:
> WDYT of putting `# pragma: no cover` for lines which are not supposed to be covered?

In general, that SGTM.

Just for me understanding: The idea would be that if we know that a line is not covered, we add this so that codecov et al don't complain about this line, and in order for us to signal that we are aware that this line is not tested? Would that be for all of skops or for this feature specifically?

I think there is a danger of using the pragma: IIUC, if we do end up covering the line (even if unintentionally) by adding a test, we would _not_ get an error that the line is covered now even though we "claim" it's not covered. Then, if someone else comes and, say, modifies the test so that the line no longer is covered, we would not get a warning about decreased coverage, right? So I feel like using the pramga would weaken the reason of tracking code coverage in the first place.

(This is on contrast, e.g., to using pytest with `xfail`, were one can indicate that the `xfail` is strict and thus it's hard to unintentionally make a test pass without noticing.)

adrinjalali:
If we write a test that covers those lines, we'd know, right? They're mostly raising exceptions which we'd be catching in the tests, and therefore we'd be removing the pragma.

To me this is very different than `xfail`. This is about us not necessarily caring about a line being covered, or that we know it's not covered.

The point of codecov is not to make sure every single line is covered. It's about us knowing if things which we should be testing are tested or not, and it's doing that for us.

BenjaminBossan:
@adrinjalali 

I added pragmas where it makes sense and tests for all the remaining lines that were not tested.

> If we write a test that covers those lines, we'd know, right? They're mostly raising exceptions which we'd be catching in the tests, and therefore we'd be removing the pragma.

Yes, for these specific lines, that's true.

> To me this is very different than `xfail`. This is about us not necessarily caring about a line being covered, or that we know it's not covered.

I only brought that up in the sense that it has the `strict` option, which would be nice to have for `pragma: no cover` as well.

PR #258, opened 2022-12-17
MNT: Change placeholder character to control character

E-Aho:
Small change that swaps from using an arbitrary string of real world characters to a unicode control character.

This should reduce the risk of bugs (e.g, the original code would cause bugs if a user decided to use '$%!?' in their model card subsection or table), and the `unit seperator` feels a good fit for this use case.

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -41,7 +41,7 @@ def _clean_table(table: str) -> str:
     # replace line breaks "\n" with html tag <br />, however, leave end-of-line
     # line breaks (eol_lb) intact
     eol_lb = "|\n"
-    placeholder = "$%!?"  # arbitrary string that never appears naturally
+    placeholder = "\x1f"  # unit separator control character
     table = (
         table.replace(eol_lb, placeholder)
         .replace("\n", "<br />")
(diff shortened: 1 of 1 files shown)


skops/card/_model_card.py
```diff
@@ -41,7 +41,7 @@ def _clean_table(table: str) -> str:
     # replace line breaks "\n" with html tag <br />, however, leave end-of-line
     # line breaks (eol_lb) intact
     eol_lb = "|\n"
-    placeholder = "$%!?"  # arbitrary string that never appears naturally
+    placeholder = "\x1f"  # unit separator control character
```
adrinjalali, line 44: I can't find this character, can you add something which would helps us find it in the tables, or add a description of what it is?

adrinjalali (commented):
Otherwise LGTM.

skops/card/_model_card.py
E-Aho: Sure, in the Python ASCII control character library it appears [here](https://github.com/python/cpython/blob/3.11/Lib/curses/ascii.py#L36), I could link to this part of the source code, or even use the `curses.ascii` character in place if that would help readibility?

BenjaminBossan (approved):
Correct me if I'm wrong, this change wouldn't make a bug completely impossible, just (even more) unlikely. If so, I'm happy with the change.

skops/card/_model_card.py
E-Aho: Incredibly annoyingly, curses, the cpython module that supports ascii characters, [is not supported on windows](https://docs.python.org/3/howto/curses.html).

I've just added a small comment that refers to it in the curses library, and left it as a coded string instead of trying to use `ascii.US`

PR #259, opened 2022-12-19
DOC make the persistence warning the same as the user guide

adrinjalali:
This is a hotfix to make the warning in the API docs the same as the user guide.

Since https://github.com/scikit-learn/scikit-learn/pull/25197 this merged and might be backported to the stable release soon, we should fix this warning, and backport it to the 0.4.X branch. It won't require a new release since it's only a documentation change.

I'll merge tomorrow if nobody else does. I don't like urgent things, but this might hurt us if we don't fix it soon.

cc @skops-dev/maintainers

### DIFF
--- a/skops/io/_persist.py
+++ b/skops/io/_persist.py
@@ -48,10 +48,9 @@ def dump(obj: Any, file: str | Path | BinaryIO) -> None:
 
     .. warning::
 
-       This feature is very early in development, which means the API is
-       unstable and it is **not secure** at the moment. Therefore, use the same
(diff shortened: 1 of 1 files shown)


BenjaminBossan (approved):
I'm fine with the change in wording.

Before merging, for my understanding, you mean this needs to be updated quick because sklearn will soon link to skops? Or how is it related?

adrinjalali:
Yes, the link to skops is already live on sklearn's dev branch, but will go live on the stable branch soon too.

https://scikit-learn.org/dev/model_persistence.html

PR #260, opened 2023-01-02
FIX: Persistence tests failing when using sklearn nightly

BenjaminBossan:
This PR is urgent because CI for our PRs will fail until there is a fix.

After https://github.com/scikit-learn/scikit-learn/pull/22094, sklearn estimators will contain an additional key in their `__dict__` after loading, namely `"__sklearn_pickle_version__"`. This causes our tests to fail, since they compare objects before and after loading.

The quick solution is to pop the item in our tests if it exists and only compare the remaining items.

Should the sklearn change be amended, we should remove the fix from this PR. Progress is tracked here:

https://github.com/scikit-learn/scikit-learn/issues/25273

### DIFF
--- a/skops/io/tests/_utils.py
+++ b/skops/io/tests/_utils.py
@@ -126,6 +126,13 @@ def _assert_vals_equal(val1, val2):
 
 
 def assert_params_equal(params1, params2):
+    # due to https://github.com/scikit-learn/scikit-learn/pull/22094, after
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
@skops-dev/maintainers this is ready for review.

[Here](https://github.com/skops-dev/skops/actions/runs/3822194388/jobs/6502083486) is an example test run that currently fails because of this issue.

PR #261, opened 2023-01-03
DOC Add section on security of supported libraries

BenjaminBossan:
This is to make it more transparent for users what they can expect when using skops with supported external libraries.


BenjaminBossan:

adrinjalali (approved):
I'll let @E-Aho to also have a look here.

E-Aho (approved):
LGTM :)

PR #264, opened 2023-01-08
Test Merve's importutils on GH CI

BenjaminBossan:
*Just a test, please ignore*

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -76,7 +76,7 @@ jobs:
       env:
         SUPER_SECRET: ${{ secrets.HF_HUB_TOKEN }}
       run: |
-        python -m pytest -s -v --cov-report=xml -m "not inference" skops/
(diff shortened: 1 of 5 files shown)


BenjaminBossan:
@merveenoyan I was looking for you but couldn't find you, so I just started working on the matplotlib testing issue on my own ;)

I could reproduce your error (test succeeding locally and failing on CI). After some digging, it turned out that there were a couple of issued stacked on top of each other (probably best to explain in person).

 But now, I think I found a viable solution. Please take a look.

adrinjalali:
@merveenoyan note that for a final solution, you'd need to add the deleted modules from cache back to the cache after the test. Happy to have a huddle when I'm back from holidays to work on this :)

merveenoyan:
I merged main and put these into my PR, will make it work and let you know! @adrinjalali @BenjaminBossan

Issue #251, opened 2022-12-13
Allow users to use scikit-learn-intelex on the API inference backend


adrinjalali:
This requirement comes from the contract between Hugging Face and Intel.

We should have a way to allow users to utilize `scikit-learn-intelex` at inference time. This requires a few steps:

- Check if models trained with `scikit-learn` can be served with `scikit-learn-intelex` and vice versa.
- Expose a method in `skops.hub_utils` to add a configuration flag to use `scikit-learn-intelex`
	- Should probably also add a tag, so that we can have better visibility on models using this option.
- On the https://github.com/huggingface/api-inference-community/tree/main/docker_images/sklearn side, if the flag exists, install the `scikit-learn-intelex` package in the environment and run the script with this command: `python -m sklearnex my_application.py`
	- We should check if packages on `conda-forge` are up-to-date: https://anaconda.org/conda-forge/scikit-learn-intelex

Project homepage: https://intel.github.io/scikit-learn-intelex/

BenjaminBossan:
> * Check if models trained with `scikit-learn` can be served with `scikit-learn-intelex` and vice versa.

I started with this point and found it's possible to load sklearn (w/o intelex) models with intelex and vice versa. Here are some quick test results:

I created two scripts, `make-sklearn.py` and `make-intelex.py`, that will fit and save either a pure sklearn model, or an sklearn+intelex model, respectively. The scripts `load-sklearn.py` and `load-intelex.py` load a model (either fitted with or w/o intelex) and call `predict_proba` on it, with the latter script patching sklearn with intelex. All directions seem to work (ignore the `200000`, which is just sample size):

```
$ python make-sklearn.py 200000
Fit time:	14.487415

$ python make-intelex.py 200000
Intel(R) Extension for Scikit-learn* enabled (https://github.com/intel/scikit-learn-intelex)
Fit time:	14.715771


$ python load-sklearn.py 200000 sklearn.pickle
Predict time:	2.506398

$ python load-sklearn.py 200000 intelex.pickle
Predict time:	2.360727


$ python load-intelex.py 200000 sklearn.pickle
Intel(R) Extension for Scikit-learn* enabled (https://github.com/intel/scikit-learn-intelex)
Predict time:	2.501361

$ python load-intelex.py 200000 intelex.pickle
Intel(R) Extension for Scikit-learn* enabled (https://github.com/intel/scikit-learn-intelex)
Predict time:	2.408000
```

[...]

PR #267, opened 2023-01-17
ENH Work on supporting scikit-learn intelex

BenjaminBossan:
Partially solves #251

This is one part of the work required to solve the mentioned issue. The other part will have to be added on the API inference side, once this PR is finalized.

## Description

Added the option `use_intelex` to `hub_utils.init`. It adds a new entry to `config.json` which can later be used by the inference API to decide whether to run it with intelex or not.

On top of that, if `use_intelex=True`, scikit-learn-intelex will be added as a dependency to the requirements if not already there. Moreover, if metadata for a model card is loaded, a `"scikit-learn-intelex"` tag will be added.

### DIFF
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -257,11 +257,12 @@ def _create_config(
         - ``"skops"`` if the extension is ``".skops"``
 
     use_intelex: bool (default=False)
-        Whether to enable scikit-learn-intelex. This can accelerate some sklearn
-        models by a large factor with the right hardware. Enabling this option
-        should not break any code, even if the model was not initially trained
-        with scikit-learn intelex and even if the hardware does not support it.
-        For more info, see https://intel.github.io/scikit-learn-intelex/.
+        Whether to enable ``scikit-learn-intelex``. This can accelerate some
+        sklearn models by a large factor with the right hardware. In most cases,
+        enabling this option should not break any code, even if the model was
+        not initially trained with scikit-learn intelex and even if the hardware
+        does not support it. For more info, see
+        https://intel.github.io/scikit-learn-intelex/.
     """
     # so that we don't have to explicitly add keys and they're added as a
     # dictionary if they are not found
@@ -387,11 +388,12 @@ def init(
         or ``"pickle"``. Defaults to ``"auto"`` that relies on file extension.
 
     use_intelex: bool (default=False)
-        Whether to enable scikit-learn-intelex. This can accelerate some sklearn
-        models by a large factor with the right hardware. Enabling this option
-        should not break any code, even if the model was not initially trained
-        with scikit-learn intelex and even if the hardware does not support it.
-        For more info, see https://intel.github.io/scikit-learn-intelex/.
+        Whether to enable ``scikit-learn-intelex``. This can accelerate some
+        sklearn models by a large factor with the right hardware. In most cases,
+        enabling this option should not break any code, even if the model was
+        not initially trained with scikit-learn intelex and even if the hardware
+        does not support it. For more info, see
+        https://intel.github.io/scikit-learn-intelex/.
     """
     dst = Path(dst)
     if dst.exists() and bool(next(dst.iterdir(), None)):
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -154,6 +154,7 @@ def _create_config(
         "pickle",
         "auto",
     ] = "auto",
+    use_intelex: bool = False,
 ) -> None:
     """Write the configuration into a ``config.json`` file.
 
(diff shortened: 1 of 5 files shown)


BenjaminBossan:

skops/hub_utils/_hf_hub.py
```diff
@@ -188,16 +189,19 @@ def _create_config(
...
         or ``"pickle"``. Defaults to ``"auto"``, which would mean:
 
         - ``"pickle"`` if the extension is one of ``{".pickle", ".pkl", ".joblib"}``
         - ``"skops"`` if the extension is ``".skops"``
 
-    Returns
-    -------
-    None
+    use_intelex: bool (default=False)
+        Whether to enable scikit-learn-intelex. This can accelerate some sklearn
```
adrinjalali, line 200:
```suggestion
        Whether to enable ``scikit-learn-intelex``. This can accelerate some sklearn
```

skops/hub_utils/_hf_hub.py
```diff
@@ -188,16 +189,19 @@ def _create_config(
...
+        models by a large factor with the right hardware. Enabling this option
```
adrinjalali, line 201:
```suggestion
        models by a large factor with the right hardware. In most cases, enabling this option
```

skops/hub_utils/_hf_hub.py
```diff
@@ -339,6 +349,12 @@ def init(
 
     dst.mkdir(parents=True, exist_ok=True)
 
+    # add intelex requirement, if it's used and not already in requirements
+    if use_intelex and not any(
+        r.startswith("scikit-learn-intelex") for r in requirements
+    ):
+        requirements.append("scikit-learn-intelex")
```
adrinjalali, line 356: should we add the current version as well? I think we should

skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -658,3 +658,51 @@ def test_adding_existing_file_raises(self, init_path, some_file_0):
...
+        config = get_config(dir_path)
+        return config
+
+    def test_no_intelex(self, classifier):
+        # by default, intelex is not being used
+        config = self.make_config(model=classifier, requirements=["foobar"])
+        environement = config["sklearn"]["environment"]
+
+        assert config["sklearn"]["use_intelex"] is False
+        assert not any(r.startswith("scikit-learn-intelex") for r in environement)
```
adrinjalali, line 686: then we should probably check for the case where the user says `use_intelex=False` but explicitly passes it as a requirement and raise as inconsistent

adrinjalali (commented):
Thanks @BenjaminBossan

skops/hub_utils/tests/test_hf_hub.py
BenjaminBossan: Is that really a good idea? I think if the user wants intelex installed without using it _for the purpose here_, we should allow it.

adrinjalali: fair point, I guess we don't care if it's installed.

skops/hub_utils/_hf_hub.py
BenjaminBossan: Hmm, maybe. That would require sklearnex to be installed though, and technically, users should be allowed to enable the intelex even if they don't have sklearnex installed.

The second problem is that I can't find how to determine the sklearnex version. What I found in their code is this:

https://github.com/intel/scikit-learn-intelex/blob/6b31cc8499d12a1ccbe8200b0e90993f6731c44e/setup_sklearnex.py#L24

So they use some kind of calendar versioning but I haven't found the version as an attribute on the package.

There's also this function:

https://github.com/intel/scikit-learn-intelex/blob/6b31cc8499d12a1ccbe8200b0e90993f6731c44e/sklearnex/_utils.py#L74

But it doesn't seem to return the version (AFAICT, it requires you to pass the version and does some checks on it).

If users want to have a specific version, they can indicate it, it's just a small extra step.

BenjaminBossan:
@adrinjalali I addressed your comments, pls review again.

skops/hub_utils/_hf_hub.py
adrinjalali: what would `importlib.metadata.version` give you?

BenjaminBossan: Hmm

```
$ python -c "from importlib.metadata import version; print(version('scikit-learn-intelex'))"
2023.20221220.91616
$ python -m pip install "scikit-learn-intelex==2023.20221220.91616"
ERROR: Could not find a version that satisfies the requirement scikit-learn-intelex==2023.20221220.91616 (from versions: 2021.2.1, 2021.2.2, 2021.2.3, 2021.3.0, 2021.4.0, 2021.5.1, 2021.5.3, 2021.6.0, 2021.6.2, 2021.6.3, 2023.0.0, 2023.0.1)
ERROR: No matching distribution found for scikit-learn-intelex==2023.20221220.91616
```

adrinjalali:
I just saw this:

https://github.com/intel/scikit-learn-intelex/blob/d71f63c12cba7ee99191e3013182b0650ac9ec71/sklearnex/dispatcher.py#L104-L107

```py
def patch_sklearn(name=None, verbose=True, global_patch=False):
    if not sklearn_check_version('0.22'):
        raise NotImplementedError("Intel(R) Extension for Scikit-learn* patches apply "
                                  "for scikit-learn >= 0.22 only ...")
```

Which patching wouldn't be an option for us. That's a VERY OLD sklearn version. I checked their codebase, and they have their own estimators which can be imported and used using explicit imports from their library.

If the user explicitly imports from the `scikit-learn-intelex` library, then everything works as expected. They can have it as an extra dependency, and the backend would install it and load the model successfully.

Which means this task would be reduced to a documentation and an example script on our side, to showcase how one can use `scikit-learn-intelex` on HF hub. WDYT @BenjaminBossan @echarlaix

BenjaminBossan:
> Which patching wouldn't be an option for us. That's a VERY OLD sklearn version.

I think this is the minimum sklearn version, so we should be good. When I tested locally, patching worked (or, at least, didn't raise an error).

adrinjalali:
Oh sorry, it's checking if the sklearn version is >0.22, that's a very bad way of writing code lol, but sure, it works.

adrinjalali (approved):

BenjaminBossan:
> Oh sorry, it's checking if the sklearn version is >0.22, that's a very bad way of writing code lol, but sure, it works.

Yes, naming is hard :)

PR #272, opened 2023-01-19
Fix for metadata having model format 

### LINKED ISSUE #271: metadata is missing the model format
We added the `model_format` to the `config.json` but when loading the metadata with `metadata_from_config`, we actually don't load that attribute. This also has some consequences for the model card, because that information is used to show the getting started code.

merveenoyan:
Fixes #271


BenjaminBossan (approved):
Thanks for fixing this!

PR #284, opened 2023-01-25
ENH pandoc parser minor improvements and support older pandoc

BenjaminBossan:
### Add support for `LineBreak` item

This one was a bit obscure, it requires the model card to contain a line break with trailing whitespace to be triggered.

### Add support for older pandoc versions

This one came up when working on a HF space. There, non-Python dependencies can be installed, but only as debian packages via `apt`. However, the `apt` repo only has an old pandoc version, which doesn't work with the current implementation.

The solutions are either to support older pandoc versions or to use a Docker space, which seems like overkill. Fortunately, supporting older pandoc versions wasn't as hard as expected. The main change was introduced with (afaict) pandoc 2.5, which changed handling of tables. Now there are two table implementations, old and new, which differ slightly.

Tested and works with the following pandoc versions:

- 2.0
- 2.2
- 2.5
- 2.19
- 3.0

Minimum pandoc version has been decreased to 2.0. CI has been adopted to install pandoc via apt, resulting in version 2.9.2.1.

### DIFF
--- a/skops/card/_markup.py
+++ b/skops/card/_markup.py
@@ -189,16 +189,14 @@ class Markdown:
         _, txt = item
         return f"`{txt}`"
 
-    def _table_cols_old(self, items) -> list[str]:  # pragma: no cover
-        # pandoc < 2.5
+    def _table_cols_old(self, items) -> list[str]:
         columns = []
         for (content,) in items:
             column = self.__call__(content)
             columns.append(column)
         return columns
 
-    def _table_cols_new(self, items) -> list[str]:
-        # pandoc >= 2.5
+    def _table_cols_new(self, items) -> list[str]:  # pragma: no cover
         columns = []
         fn = self.__call__
         for item in items:
@@ -207,7 +205,7 @@ class Markdown:
             columns.append(column)
         return columns
 
-    def _table_body_old(self, items) -> list[list[str]]:  # pragma: no cover
+    def _table_body_old(self, items) -> list[list[str]]:
         body = []
         for row_items in items:
             row = []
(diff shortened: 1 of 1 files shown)


BenjaminBossan:

skops/card/_markup.py
```diff
@@ -185,7 +189,16 @@ def _code(item: tuple[Any, str]) -> str:
         _, txt = item
         return f"`{txt}`"
 
-    def _table_cols(self, items) -> list[str]:
+    def _table_cols_old(self, items) -> list[str]:  # pragma: no cover
+        # pandoc < 2.5
```
adrinjalali, line 193: any idea when we can drop support for this?

skops/card/_markup.py
```diff
@@ -205,20 +231,34 @@ def _table_body(self, items) -> list[list[str]]:
...
         # attr capt specs thead tbody tfoot
         _, _, _, thead, tbody, _ = item
-
         # header
         (_, thead_bodies) = thead
         (attr, thead_body) = thead_bodies[0]  # multiple headers?
-
-        columns = self._table_cols(thead_body)
-
+        columns = self._table_cols_new(thead_body)
```
adrinjalali, line 248: I would have expected these lines to be covered.

adrinjalali (commented):
Nice!

skops/card/_markup.py
BenjaminBossan: So the table change actually happens after 2.9, not 2.5 as I first suspected. Therefore, given that we now install pandoc via apt on CI, which uses 2.9, the old table implementation is covered, not the new one. I thus moved around the pragmas to the new implementation.

skops/card/_markup.py
BenjaminBossan: I guess as soon as apt updates to a more recent pandoc version. I have no idea when that will be.

adrinjalali (approved):
Well, we basically tested the new stuff, and now also the old stuff. This is good to go. Thanks @BenjaminBossan

PR #286, opened 2023-01-30
DOC fix author links

adrinjalali:
The `_authors.rst` file is only for users who contribute often enough that it makes sense to have a short hand link for them. For others we should be using the `:user:` directive.

cc @BenjaminBossan


BenjaminBossan (approved):
Nice

PR #287, opened 2023-01-30
FIX: Support operators persistence

### LINKED ISSUE #283: `TypeError` when loading model using methodcaller
Came across this error for only this model generated using the test case generation script in `api-inference-community`: 
`*** TypeError: methodcaller needs at least one argument, the method name`

To reproduce:
```python
from skops import hub_utils
hub_utils.download(repo_id = "skops-tests/textclassification-sklearn-latest-hist_gradient_boosting-with-config-skops", dst="model_dir")

import skops.io as sio
sio.load("/content/model_dir/skops-v588xnq6.skops", trusted=True)
```

At first I thought something might've been wrong with how I serialize it but I think it's weird how every test passes and this model fails for inference.

BenjaminBossan:
Fixes #283

## Description

Some of the operator functions can be quite useful sklearn users, especially in conjunction with `FunctionTransformer`. This is because it often allows to avoid writing small functions or using `lambda`, which can't be persisted. However, since some of these functions are actually classes defined in CPython that need special treatment, our code didn't work with them.

This fix consists of two parts.

On the one hand, functions like `methodcaller` now have a special dispatch type that knows how to correctly initialize them.

On the other hand, we had a problem with some types like `operator.add`. Since they take 2 arguments, it would be useful for them to be passed without initialization in conjunction with `partial`. However, their type is `builtin_function_or_method`. By default, they would be handled by `object_get_state`, but that doesn't work correctly with uninitialized objects. This has now been fixed by dispatching to `type_get_state`.

## Open questions

To get the type of `operator.add` et al., we need `builtin_function_or_method`. Right now, I just do `builtin_function_or_method = type(len)`. Is there a better way?

Regarding operators like `attrgetter` and `methodcaller`, could there be a security implication by allowing them? Sure, users can decide not to trust that type, but whether it is trustworthy or not depends on its arguments. Right now, we just tell them that they need to trust `_operator.attrgetter`, which could seem harmless. The same argument can of course be made for `eval` & co, but there, users will be much more likely to recognize their danger.

### DIFF
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import io
 import json
+import operator
 import uuid
 from functools import partial
 from types import FunctionType, MethodType
@@ -368,6 +369,18 @@ def object_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
     return res
 
 
+def builtin_function_or_method_state(
+    obj: Any, save_context: SaveContext
+) -> dict[str, Any]:
+    # Some of the functions from the stdlib, like operator.add.
+    res = {
+        "__class__": obj.__name__,
+        "__module__": get_module(obj),
+        "__loader__": "TypeNode",
+    }
+    return res
+
+
 class ObjectNode(Node):
     def __init__(
         self,
@@ -528,6 +541,37 @@ class BytearrayNode(BytesNode):
         return content_bytearray
 
 
+def operator_func_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
+    _, attrs = obj.__reduce__()
+    res = {
+        "__class__": obj.__class__.__name__,
+        "__module__": "operator",
+        "__loader__": "OperatorFuncNode",
+        "attrs": attrs,
+    }
+    return res
+
+
+class OperatorFuncNode(Node):
+    def __init__(
+        self,
+        state: dict[str, Any],
+        load_context: LoadContext,
+        trusted: bool | Sequence[str] = False,
+    ) -> None:
+        super().__init__(state, load_context, trusted)
+        self.trusted = self._get_trusted(trusted, [bytes])
+        self.children["attrs"] = state["attrs"]
+
+    def _construct(self):
+        op = getattr(operator, self.class_name)
+        return op(*self.children["attrs"])
+
+
+# <class 'builtin_function_or_method'>
+builtin_function_or_method = type(len)
+
+
 # tuples of type and function that gets the state of that type
 GET_STATE_DISPATCH_FUNCTIONS = [
     (dict, dict_get_state),
@@ -541,6 +585,10 @@ GET_STATE_DISPATCH_FUNCTIONS = [
     (MethodType, method_get_state),
     (partial, partial_get_state),
     (type, type_get_state),
+    (builtin_function_or_method, builtin_function_or_method_state),
+    (operator.attrgetter, operator_func_get_state),
+    (operator.itemgetter, operator_func_get_state),
+    (operator.methodcaller, operator_func_get_state),
     (object, object_get_state),
 ]
 
@@ -558,4 +606,5 @@ NODE_TYPE_MAPPING = {
     "TypeNode": TypeNode,
     "ObjectNode": ObjectNode,
     "JsonNode": JsonNode,
+    "OperatorFuncNode": OperatorFuncNode,
 }
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -369,18 +369,6 @@ def object_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
     return res
 
 
-def builtin_function_or_method_state(
-    obj: Any, save_context: SaveContext
-) -> dict[str, Any]:
-    # Some of the functions from the stdlib, like operator.add.
-    res = {
-        "__class__": obj.__name__,
-        "__module__": get_module(obj),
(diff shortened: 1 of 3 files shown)


BenjaminBossan:

skops/io/_general.py
```diff
@@ -528,6 +529,37 @@ def _construct(self):
         return content_bytearray
 
 
+def operator_func_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
+    _, attrs = obj.__reduce__()
```
adrinjalali, line 533: instead of this, did you test what happens if we use the reduce loader we already have?

adrinjalali (commented):
Another alternative is to call `__new__` with the param we need.

BenjaminBossan:
> Another alternative is to call __new__ with the param we need.

Do you mean changing the line 

`instance = cls.__new__(cls)`

in `ObjectNode`. This was my first attempt (which also worked) but it required adding a bunch of `if ... else` into the generic object path, which made the resulting code harder to understand. Using a special dispatch function seemed to be the better way.

> instead of this, did you test what happens if we use the reduce loader we already have?

I tried it and it can be made to work. Here are a list of changes:

1. move the `reduce_get_state` and `ReduceNode` to `_general.py`
2. write custom `operator_reduce_get_state` and `OperatoreReduceNode` (since the generic versions are not intended to be used directly), which are about the same length of the current `operator_func_get_state` and `OperatorFuncNode`
3. for `operator.add` et al, we still need to use the type approach

Overall, it's the same amount of code but with more indirection, and more to rewrite should we ever get rid of the `reduce_get_state` and `ReduceNode`. Therefore, I would slightly prefer the existing suggestion. I have stashed the other solution though, If there are better arguments to make the switch.

skops/io/_general.py
```diff
@@ -532,6 +533,37 @@ def _construct(self):
...
+        self.trusted = self._get_trusted(trusted, [])
+        self.children["attrs"] = state["attrs"]
+
+    def _construct(self):
+        op = getattr(operator, self.class_name)
+        return op(*self.children["attrs"])
+
+
+# <class 'builtin_function_or_method'>
+builtin_function_or_method = type(len)
```
adrinjalali, line 564: what made you add this? Is this how we support `neg` for example? Can't be.

adrinjalali (commented):
One issue I noticed, is that `operator` module includes too much. It sounds like an innocent module, but then it has things like `call`, `setitem`, `deltem`, and inplace manipulators. I wanna make sure those are not supported out of the box, and certainly not as _trusted_.

skops/io/_general.py
BenjaminBossan: Some operators can reasonably be used without instantiating when they don't take an argument, e.g. `operator.neg`. Therefore, we have to match their type, which is `<class 'builtin_function_or_method'>`. I couldn't find a nice way to reference that type, which is why I used `type(len)`, but it doesn't really have to be `len`, could be something else. If you know a better method, please let me know.

BenjaminBossan (commented):
> One issue I noticed, is that `operator` module includes too much. It sounds like an innocent module, but then it has things like `call`, `setitem`, `delitem`, and inplace manipulators. I wanna make sure those are not supported out of the box, and certainly not as trusted.

As is, users need to trust those, we haven't added them to the list of automatically trusted modules. Do you want me to write a test that checks that loading fails if they are not set as trusted by the user?

adrinjalali:
A test to make sure those are not trusted would be nice. Generally, making sure things we know shouldn't be trusted are not trusted is a nice thing. I'm happy to do it here or in a separate PR, as you wish.

BenjaminBossan:
@adrinjalali I added tests as suggested. I also found a somewhat embarrassing bug, not sure if we could have added some checks to prevent this type of bug in the future.

skops/io/tests/test_persist.py
```diff
@@ -880,3 +881,63 @@ def test_estimator_with_bytes_files_created(tmp_path):
...
+    ("le", partial(operator.le, 0)),
+    ("lt", partial(operator.lt, 0)),
+    ("eq", partial(operator.eq, 0)),
+    ("neg", operator.neg),
+    ("attrgetter", operator.attrgetter("real")),
+    ("attrgetter", operator.attrgetter("real", "real")),
+    ("itemgetter", operator.itemgetter(None)),
+    ("itemgetter", operator.itemgetter(None, None)),
+    ("methodcaller", operator.methodcaller("round")),
+    ("methodcaller", operator.methodcaller("round", 2)),
```
adrinjalali, line 904: could we add `delitem`, `setitem`, `call`, and an inplace one like `iadd`?

adrinjalali (commented):
Otherwise LGTM.

As for the bug you found and the test for it, it's a good question.

skops/io/tests/test_persist.py
BenjaminBossan: The problem with `delitem` and `setitem` is that they can't be really used in this context. They need to be called with the collection as first argument, so when I set them as `partial(operator.delitem, 0)`, for instance, it won't work when this is called with `X` (it tries to delete `X` from 0 instead of the other way round). And since these operators take no keyword arguments, it's not possible to set up `partial` to call the arguments in the right order.

I did, however, add all the `i*` operators for the functions already tested, and for Python >= 3.11, I added `operator.call`.

BenjaminBossan:
> Hope this doesn't blow up

What do you mean? Security-wise? Maybe we can better document that users shouldn't blindly `trust` every type and what builtin types to look out for.

PR #288, opened 2023-01-31
ENH: Add possibility to turn section invisible

BenjaminBossan:
Sections of model cards have an additional flag called `visible`. It is by default `True` but if set to `False`, the corresponding section, and its subsections, are not rendered.

In contrast to deleting those section, turning them invisible will not remove them from the underlying data dict. This is useful because it allows us to restore those sections to exactly their previous place. In contrast, if we delete and add them again, they may end up in a different position, because the order of the underlying dict can change.

At the moment, the feature is not used anywhere in skops. I plan, however, to use it for a model card creation space, where it would be quite useful to have.

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -243,16 +243,20 @@ class Section:
     empty string) or a ``Formattable``, which is simply an object with a
     ``format`` method that returns a string.
 
-    Finally, the section can contain subsections, which again are dicts of
+    The section can contain subsections, which again are dicts of
     string keys and section values (the dict can be empty). Therefore, the model
     card representation forms a tree structure, making use of the fact that dict
     order is preserved.
+
+    The section may also contain a ``visible`` flag, which determined if the
+    section will be shown when the card is rendered.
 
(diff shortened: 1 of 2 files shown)


merveenoyan (approved):
I think we can merge, this looks like a small and straightforward change. Would be nice to document later on after the Space is out.

adrinjalali:
@BenjaminBossan Does this mean users need to first add a section, then `select` it and then set the flag? It'd be nicer maybe to allow that from the start maybe? Users could even be adding a `Section` object?

BenjaminBossan:
@adrinjalali Yes, basically there is no user API for making a section invisible yet. We can think about what would be the best way to do it. For now, I'm not sure if it needs to be exposed to users, let's perhaps wait and see if there is a need?

At the moment, this flag is used only for the model card space, because it allows to _undo_ the deletion of a section. If we deleted a section be deleting it from `_data` and then try to undo that by adding it back to `_data`, the order is changed, which is what I wanted to prevent.

PR #290, opened 2023-02-03
DOC Document compatibility across sklearn versions

### LINKED ISSUE #270: Forward compatibility of sklearn models saved with skops.io
Are sklearn models saved with skops.io forward compatible? For example, if I have a model trained and saved with sklearn 1.0, would it still work in sklearn 1.1?

BenjaminBossan:
Resolves #270

Add a section to the persistence docs that cautions about using skops to load sklearn models saved with one version while using another version. Refers to sklearn docs on that issue.

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -164,12 +164,13 @@ Compatibility across sklearn versions
 -------------------------------------
 
 Using skops to load a model saved in one sklearn version and loading it with
-another sklearn version is not recommended, because the behavior of the model may change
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
Generated documentation can be found [here](https://skops--290.org.readthedocs.build/en/290/persistence.html#compatibility-across-sklearn-versions). Ready for review @skops-dev/maintainers

BenjaminBossan:
@adrinjalali I merged your suggestions.

PR #254, opened 2022-12-15
Add tabular regression example

lazarust:

PR #291, opened 2023-02-03
DOC Add example of using a custom template with model cards

### LINKED ISSUE #248: Add example of using custom templates for model cards
Once #203 is merged, it would be good to give an example of how to use custom templates for model cards.

This should include how to add the sections that come by default with the default template:

- metrics (when calling `card.add_metrics`)
- getting started code (automatic)
- model hyperparams
- model plot (automatic if `model_diagram=True`)

Those are not added by default when using a custom template but it's still possible to add them the normal way (i.e. using `card.add`), although it's not obvious how. The examples should explain that.

This might also require exposing some convenience functions to the user, e.g. a way to create the sklearn model diagram with a single function call.

BenjaminBossan:
Resolves #248

This example showcases how to use custom templates with the skops model card class.

The new document is not yet explicitly linked in the documentation yet. Ideally, this can be done once #254 has been finalized.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,6 +11,8 @@ skops Changelog
 
 v0.6
 ----
+- Add example of using model card without the skops template. :pr:`291` by
+  `Benjamin Bossan`_.
 
 v0.5
 ----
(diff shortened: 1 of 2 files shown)


BenjaminBossan:
The resulting page can be found [here](https://skops--291.org.readthedocs.build/en/291/auto_examples/plot_custom_model_card.html#sphx-glr-auto-examples-plot-custom-model-card-py). Ready for review @skops-dev/maintainers

PR #295, opened 2023-02-06
`ENH` trusting scipy.ufuncs

### LINKED ISSUE #224: Add {numpy, scipy}.ufuncs to the trusted list
I think most (if not all) ufuncs are actually safe, so we can add them to the default trusted list and reduce the size of `untrusted_types`.

omar-araboghli:
(Partially) closes #224 

* Q1: Is it sufficient to only trust ufuncs under `scipy.special` ? As for the current `scipy` version, public ufuncs are only visible in `scipy.special`. Nevertheless, for future `scipy` versions (and also `numpy`), we can either recursively traverse `scipy` looking for public ufuncs or limit our search to the first submodules level. E.g.:
```python
# pseudo-code: recursively check all submodules. Warning: this needs proper handling to avoid RecursionError
for submodule in get_submodules(scipy, max_level=-1):
    look_for_type(submodule, np.func)

# pseudo-code: first submodule level
for submodule in get_submodules(scipy, max_level=1):
    look_for_type(submodule, np.func)
```
* Q2: Does `test_can_trust_ufuncs()` make sense in addition to extending `test_can_persist_fitted()` ?
* Q3: Why wouldn't we include  **current** public `numpy` ufuncs the same way we do for `scipy`, even when `numpy` 2.0 not there yet ?

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -12,8 +12,9 @@ skops Changelog
 v0.6
 ----
 - Added tabular regression example. :pr: `254` by `Thomas Lazarus`
-- All public ``scipy.special`` ufuncs (Universal Functions) are trusted by default.
-  :pr:`295` by :user:`Omar Arab Oghli <omar-araboghli>`.
+- All public ``scipy.special`` ufuncs (Universal Functions) are trusted by default
+  by :func:`.io.load`. :pr:`295` by :user:`Omar Arab Oghli <omar-araboghli>`.
+
 v0.5
 ----
 - Added CLI entrypoint support (:func:`.cli.entrypoint.main_cli`)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,6 +11,7 @@ skops Changelog
 
 v0.6
 ----
+- Added tabular regression example. :pr: `254` by `Thomas Lazarus`
 
 v0.5
 ----
@@ -102,4 +103,4 @@ Contributors
 :user:`Adrin Jalali <adrinjalali>`, :user:`Merve Noyan <merveenoyan>`,
 :user:`Benjamin Bossan <BenjaminBossan>`, :user:`Ayyuce Demirbas
 <ayyucedemirbas>`, :user:`Prajjwal Mishra <p-mishra1>`, :user:`Francesco Cariaggi <anferico>`,
-:user:`Erin Aho <E-Aho>`
+:user:`Erin Aho <E-Aho>`, :user:`Thomas Lazarus <lazarust>`
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,4 +1,4 @@
-.. include:: _authors.rst
+ds.. include:: _authors.rst
 
 .. _changelog:
 
@@ -12,7 +12,8 @@ skops Changelog
 v0.6
 ----
 - Added tabular regression example. :pr: `254` by `Thomas Lazarus`
-
+- All public ``scipy.special`` ufuncs (Universal Functions) are trusted by default.
+  :pr:`295` by :user:`Omar Arab Oghli <omar-araboghli>`.
 v0.5
 ----
 - Added CLI entrypoint support (:func:`.cli.entrypoint.main_cli`)
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -1,4 +1,4 @@
-ds.. include:: _authors.rst
+.. include:: _authors.rst
 
 .. _changelog:
 
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -216,7 +216,7 @@ class FunctionNode(Node):
         )
 
     def get_unsafe_set(self) -> set[str]:
-        if self.trusted is True or self._get_function_name() in self.trusted:
+        if (self.trusted is True) or (self._get_function_name() in self.trusted):
             return set()
 
         return {self._get_function_name()}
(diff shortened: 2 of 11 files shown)


omar-araboghli:
@BenjaminBossan kindly pinging you 👆🏻

skops/io/tests/test_persist.py
```diff
@@ -345,9 +354,18 @@ def test_can_persist_fitted(estimator):
...
     assert not any(type_ in SKLEARN_ESTIMATOR_TYPE_NAMES for type_ in untrusted_types)
+    assert not any(type_ in SCIPY_UFUNC_TYPE_NAMES for type_ in untrusted_types)
     assert_method_outputs_equal(estimator, loaded, X)
 
 
+@pytest.mark.parametrize("ufunc", _tested_ufuncs(), ids=SCIPY_UFUNC_TYPE_NAMES)
+def test_can_trust_ufuncs(ufunc):
+    dumped = dumps(ufunc)
+    untrusted_types = get_untrusted_types(data=dumped)
+    assert not any(type_ in SCIPY_UFUNC_TYPE_NAMES for type_ in untrusted_types)
```
BenjaminBossan, line 365: Can we not be sure here that `untrusted_types` is always the empty list?

docs/changes.rst
```diff
@@ -1,4 +1,4 @@
-.. include:: _authors.rst
+ds.. include:: _authors.rst
```
BenjaminBossan, line 1: Is this addition on purpose?

skops/io/_general.py
```diff
@@ -212,7 +216,7 @@ def _get_function_name(self) -> str:
         )
 
     def get_unsafe_set(self) -> set[str]:
-        if self.trusted is True:
+        if self.trusted is True or self._get_function_name() in self.trusted:
```
BenjaminBossan, line 219: Just my personal preference, but operator precedence can sometimes be tricky.

```suggestion
        if (self.trusted is True) or (self._get_function_name() in self.trusted):
```

skops/io/tests/test_persist.py
```diff
@@ -221,6 +221,15 @@ def _tested_estimators(type_filter=None):
     )
 
 
+def _tested_ufuncs():
+    for ufunc_name in SCIPY_UFUNC_TYPE_NAMES:
+        parts = ufunc_name.split(".")
+        module_name = ".".join(parts[:-1])
+        ufunc_name = parts[-1]
+
+        yield gettype(module_name=module_name, cls_or_func=ufunc_name)
```
BenjaminBossan, line 230:
```suggestion
    for full_name in SCIPY_UFUNC_TYPE_NAMES:
        module_name, _, ufunc_name = full_name.rpartition(".")
        yield gettype(module_name=module_name, cls_or_func=ufunc_name)
```

Just a suggestion to make the code more concise ;)

BenjaminBossan (changes requested):
Thanks a lot for your addition. Overall, this already looks quite good, but I have a few comments, so please take a look. As to your questions:

> Q1: Is it sufficient to only trust ufuncs under scipy.special ?

At the very least, this is a good start. If we miss something, we can add it later. Did you find any ufuncs that are not included with this approach?

I think in general, we could ask if we only want to add ufuncs or if we might not trust other scipy functions by default, like some of the `scipy.stats` functions. But let's leave that for later.

> Q2: Does test_can_trust_ufuncs() make sense in addition to extending test_can_persist_fitted() ?

I can see how it might come across as a bit of redundant, but I still think it's better to have if than not to have it.

If you look at the other tests, they almost always test some kind of sklearn estimator that contains the object in question. So here, we could use a `FunctionTransformer` to wrap the ufunc (similar to how `test_can_persist_fitted` does it). However, it's probably not possible to fit that transformer generically, so the `FunctionTransformer` would need to stay unfitted, which makes this extra step add little value.

> Q3: Why wouldn't we include current public numpy ufuncs the same way we do for scipy, even when numpy 2.0 not there yet?

We can certainly explore this and check if it's feasible and if we're missing a lot. I would do that in a separate PR though.

skops/io/tests/test_persist.py
omar-araboghli: That's true. I changed it accordingly.

docs/changes.rst
omar-araboghli: Of course not 😮  I even couldn't catch it on the github changes view... Thanks!

skops/io/tests/test_persist.py
omar-araboghli: Adapted!

docs/changes.rst
```diff
@@ -12,7 +12,8 @@ skops Changelog
 v0.6
 ----
 - Added tabular regression example. :pr: `254` by `Thomas Lazarus`
-
+- All public ``scipy.special`` ufuncs (Universal Functions) are trusted by default.
+  :pr:`295` by :user:`Omar Arab Oghli <omar-araboghli>`.
```
BenjaminBossan, line 16:
```suggestion
- All public ``scipy.special`` ufuncs (Universal Functions) are trusted by default
  by :func:`.io.load`. :pr:`295` by :user:`Omar Arab Oghli <omar-araboghli>`.

```

BenjaminBossan (commented):
Thanks, this looks almost good to go. Just a small nit left.

BenjaminBossan (approved):
Great work, thanks.

PR #297, opened 2023-02-10
MNT: Upgrade mypy to version 1.0.0

BenjaminBossan:
The upgrade worked out of the box. I made a small change, however, to adjust the signatures of `Card` to return the `Self` type, since this is now correctly recognized by mypy.

Below is an example to illustrate why it's useful:

```python
class Card1:
    def add(self) -> Card1:
        return self

class MyCard1(Card1):
    pass

reveal_type(MyCard1().add())  # returns Card1 but should be MyCard1

class Card2:
    def add(self) -> Self:
        return self

class MyCard2(Card2):
    pass

reveal_type(MyCard2().add())  # returns MyCard2
```

Another advantage of upgrading is that mypy 1.0.0 should be faster. More in [this post](https://mypy-lang.blogspot.com/2023/02/mypy-10-released.html).

Note that pre-commit config has to be updated for this change as well.

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import json
 import re
+import sys
 import textwrap
 import zipfile
 from collections.abc import Mapping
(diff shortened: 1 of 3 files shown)


BenjaminBossan:
ready for review @skops-dev/maintainers

skops/card/_model_card.py
```diff
@@ -19,6 +20,12 @@
 from skops.io import load
 from skops.utils.importutils import import_or_raise
 
+if sys.version_info >= (3, 11):
+    from typing import Self
+else:
+    from typing_extensions import Self
```
adrinjalali, line 26: is this another instance where mypy doesn't like it to be in `fixes.py`?

adrinjalali (commented):
Otherwise LGTM.

skops/card/_model_card.py
BenjaminBossan: Yes, it is indeed annoying.

PR #298, opened 2023-02-12
`ENH` add metric frame

### LINKED ISSUE #278: Helper to add fairlearn's MetricFrame to the model card
One can easily disaggregate metrics by sensitive attribute or arbitrary groups using fairlearn's [`MetricFrame`](https://fairlearn.org/v0.8/api_reference/fairlearn.metrics.html#fairlearn.metrics.MetricFrame). It'd be nice for a user to be able to easily add that to a model card.

Example usage of a MetricFrame looks like:

```py
from fairlearn.metrics import MetricFrame, selection_rate
from sklearn.metrics import accuracy_score
import pandas as pd
y_true = [1,1,1,1,1,0,0,1,1,0]
y_pred = [0,1,1,1,1,0,0,0,1,1]
sex = ['Female']*5 + ['Male']*5
metrics = {"selection_rate": selection_rate}
mf1 = MetricFrame(
     metrics=metrics,
     y_true=y_true,
     y_pred=y_pred,
     sensitive_features=sex)
```
The reported metric can be any sklearn score or a fairlearn metric.

lazarust:
Closes #278

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,7 +11,9 @@ skops Changelog
 
 v0.6
 ----
-- Added tabular regression example. :pr: `254` by `Thomas Lazarus`
+- Added tabular regression example. :pr:`254` by `Thomas Lazarus`_.
+- Added a new function :func:`skops.card.Card.add_metric_frame` to help users
+  add metrics to their model cards. :pr:`298` by `Thomas Lazarus`_.
 
 v0.5
 ----
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,11 +11,11 @@ skops Changelog
 
 v0.6
 ----
-- Added tabular regression example. :pr:`254` by `Thomas Lazarus <lazarust>`.
+- Added tabular regression example. :pr:`254` by :user:`Thomas Lazarus <lazarust>`.
 - All public ``scipy.special`` ufuncs (Universal Functions) are trusted by default
   by :func:`.io.load`. :pr:`295` by :user:`Omar Arab Oghli <omar-araboghli>`.
 - Add a new function :func:`skops.card.Card.add_metric_frame` to help users
-  add metrics to their model cards. :pr:`298` by `Thomas Lazarus <lazarust>`
+  add metrics to their model cards. :pr:`298` by :user:`Thomas Lazarus <lazarust>`
 - Add example of using model card without the skops template. :pr:`291` by
   `Benjamin Bossan`_.
 
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -11,11 +11,11 @@ skops Changelog
 
 v0.6
 ----
-- Added tabular regression example. :pr:`254` by `Thomas Lazarus`_.
+- Added tabular regression example. :pr:`254` by `Thomas Lazarus <lazarust>`.
 - All public ``scipy.special`` ufuncs (Universal Functions) are trusted by default
   by :func:`.io.load`. :pr:`295` by :user:`Omar Arab Oghli <omar-araboghli>`.
 - Added a new function :func:`skops.card.Card.add_metric_frame` to help users
-  add metrics to their model cards. :pr:`298` by `Thomas Lazarus`_.
+  add metrics to their model cards. :pr:`298` by `Thomas Lazarus <lazarust>`
 
 v0.5
 ----
--- a/skops/_min_dependencies.py
+++ b/skops/_min_dependencies.py
@@ -32,6 +32,7 @@ dependent_packages = {
     "xgboost": ("1.6", "tests", None),
     # TODO: remove condition when catboost supports python 3.11
     "catboost": ("1.0", "tests", "python_version < '3.11'"),
+    "fairlearn": ("0.8.0", "tests", None),
 }
 
 
--- a/skops/_min_dependencies.py
+++ b/skops/_min_dependencies.py
@@ -32,7 +32,7 @@ dependent_packages = {
     "xgboost": ("1.6", "tests", None),
     # TODO: remove condition when catboost supports python 3.11
     "catboost": ("1.0", "tests", "python_version < '3.11'"),
-    "fairlearn": ("0.8.0", "docs, tests", None),
+    "fairlearn": ("0.7.0", "docs, tests", None),
 }
 
 
--- a/skops/_min_dependencies.py
+++ b/skops/_min_dependencies.py
@@ -32,7 +32,7 @@ dependent_packages = {
     "xgboost": ("1.6", "tests", None),
     # TODO: remove condition when catboost supports python 3.11
     "catboost": ("1.0", "tests", "python_version < '3.11'"),
-    "fairlearn": ("0.8.0", "tests", None),
+    "fairlearn": ("0.8.0", "docs, tests", None),
 }
 
 
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -1155,7 +1155,8 @@ class Card:
         transpose=True,
     ) -> Card:
         """
-        Add a :class:`fairlearn.metrics.MetricFrame` table to the model card.
+        Add a :class:`fairlearn.metrics.MetricFrame` table to the model card. The table contains
+        the difference, group_ma, group_min, and ratio for each metric.
 
         Parameters
         ----------
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -1148,6 +1148,45 @@ class Card:
 
         return self
 
+    def add_fairlearn_metric_frame(
+        self,
+        metric_frame,
+        table_name: str = "Fairlearn MetricFrame Table",
+        transpose=True,
+    ) -> Card:
+        """
+        Add a Fairlearn MetricFrame table to the model card.
+
+        Parameters
+        ----------
+        metric_frame: MetricFrame
+            :func:`fairlearn.MetricFrame`.
+
+        transpose: bool, default=True
+            Whether to transpose the table or not.
+
+        table_name: str
+            The desired name of the table section in the model card.
+
+        Returns
+        -------
+        self: Card
+            The model card with the metric frame added.
+        """
+        frame_dict = {
+            "difference": metric_frame.difference(),
+            "group_max": metric_frame.group_max(),
+            "group_min": metric_frame.group_min(),
+            "ratio": metric_frame.ratio(),
+        }
+
+        if transpose is True:
+            pd = import_or_raise("pandas", "Pandas is used to pivot the table.")
+
+            frame_dict = pd.DataFrame(frame_dict).T
+
+        return self.add_table(folded=True, **{table_name: frame_dict})
+
     def _add_metrics(
         self,
         section: str,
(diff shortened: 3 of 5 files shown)


lazarust:
@BenjaminBossan I feel I should write a test for this, but I'm unsure where it should go. I was thinking of `skops/card/tests/test_card.py` inside the `TestTableSection` class. Thoughts? Is there somewhere else it would be better?

skops/card/_model_card.py
```diff
@@ -11,6 +11,7 @@
 from typing import Any, Iterator, Literal, Protocol, Sequence, Union
 
 import joblib
+from fairlearn.metrics import MetricFrame
```
BenjaminBossan, line 14: We should not import fairlearn at the top level, since it would mean that it's a required dependency. It should be imported inside the corresponding method, similar to how you import pandas there.

However, you should not import them directly, but use the helper function we introduced, like here: https://github.com/skops-dev/skops/blob/b324b8d3689037eda6f1a6b1b9650e7eda7dab64/skops/card/_model_card.py#L1130

skops/card/_model_card.py
```diff
@@ -1308,3 +1309,56 @@ def render(self) -> str:
             sections inserted.
         """
         return "\n".join(self._generate_card())
+
+    def add_metric_frame(
+        self, metrics: dict, y_true, y_pred, sensitive_features, pivot=True
+    ) -> Card:
+        """
+        Add a metric frame table to the model card.
```
BenjaminBossan, line 1317: About the name `add_metric_frame`, I'm just wondering out loud if we can find a better name. For users who don't know fairlearn, the name could be very confusing, and the docstring "Add a metric frame table to the model card" doesn't add much.

BenjaminBossan (commented):
I didn't do a thorough review yet, just high level.

> I feel I should write a test for this, but I'm unsure where it should go. I was thinking of skops/card/tests/test_card.py inside the TestTableSection class. Thoughts? Is there somewhere else it would be better?

I think it would be best to add a completely new `TestAddMetricFrame`. You can take inspiration from other tests though.

BenjaminBossan:
@lazarust I see that there are some `black` issues. If you set up your dev environment with pre-commit hooks as described [here](https://github.com/skops-dev/skops/blob/main/CONTRIBUTING.rst#setting-up-the-dev-environment), they should be caught before committing.

lazarust:
@BenjaminBossan Thanks for the tip! Sorry, I missed that from the Contribution Guide. I'm currently working on making the tests. I'll ping you whenever it's ready again!

lazarust:
@BenjaminBossan This is ready for you to take a look at!

BenjaminBossan (changes requested):
Thanks a lot for this. I have a couple of code comments, but before considering those, I would like to discuss a bigger design decision.

Right now, the user more or less passes the arguments for `MetricFrame` to the `add_fairlearn_metric_frame` method, which takes care of creating the `MetricFrame` instance. I would propose to change this so that the user has to create the `MetricFrame` instance themselves, then pass it to `add_fairlearn_metric_frame` ("inversion of control"), which does not need to construct it but just takes care of creating and adding the table.

This approach is similar to [`add_permutation_importances`](https://github.com/skops-dev/skops/blob/3e1f138c5a7c12863ca263beb2804ceede07d35a/skops/card/_model_card.py#L1097), which takes the computed permutation importances as input, instead of computing them inside the method.

Why do I think this could be better? Here are a few reasons:

With the current implementation, the user loses control over the instantiation of `MetricFrame`. Therefore, if they want to use something like `control_features` or `sample_params`, it's not possible. Of course, we could add more parameters to the signature of `add_fairlearn_metric_frame`, but it only gets bigger and bigger that way, and we have to keep it up to date when fairlearn changes.

Another advantage of having the user pass the instance is that we don't need to import `fairlearn` inside the method. If a user creates that object, they have already imported `fairlearn`.

One more advantage is that if a user has a custom `MetricFrame` class, they can pass that to the method, whereas right now, it's impossible to use. For testing, we could even create a mock object instead of using a real `MetricFrame`, and then skops would have no dependency on fairlearn at all! But I think adding a test dependency is fine here.

A disadvantage of my proposal is that users have to do a little bit of extra work by instantiating the object themselves.

Overall, I think this price is worth paying. What do you think? If we decide to make this change and you refactor the code accordingly, then many of my comments are obsolete. I think you will see which ones.

lazarust:
@BenjaminBossan Thanks for the review! I haven't looked at your code comments yet since the decision of whether to pass in the parameters or a `MetricFrame` affects a lot of it. 

>  Of course, we could add more parameters to the signature of add_fairlearn_metric_frame, but it only gets bigger and bigger that way, and we have to keep it up to date when fairlearn changes.

I agree! I was thinking of that while writing the method but was unsure if that was something we wanted to do. 

I do wonder if we require users to install and import `fairlearn` on their own if this feature just won't be known/used as much. I could be completely off base on that, but it is a thought I have. We could add an example of using this or just add it to the existing examples. 

Beyond that, I don't really see too many disadvantages that would outway the advantages.  What're your thoughts?

BenjaminBossan:
> I do wonder if we require users to install and import `fairlearn` on their own if this feature just won't be known/used as much.

We would do that anyway, fairlearn would not be installed by default when installing skops. And even if it were, users don't just discover and try out random packages that were installed :)

> We could add an example of using this or just add it to the existing examples.

This + documenting it well are definitely the way to go for users to discover and use the feature.

> Beyond that, I don't really see too many disadvantages that would outway the advantages. What're your thoughts?

Great, I think we agree then.

skops/card/_model_card.py
```diff
@@ -1308,3 +1308,69 @@ def render(self) -> str:
...
         return "\n".join(self._generate_card())
+
+    def add_fairlearn_metric_frame(
+        self,
+        metric_dict: dict,
+        y_true,
+        y_pred,
+        sensitive_features,
+        table_name: str,
+        pivot=True,
```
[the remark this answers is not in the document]
lazarust: Hmmm, I guess we don't really need the `pivot` if we don't want to. 

So if we don't transpose it the table looks like: 
|                |   difference |   group_max |   group_min |   ratio |
|:---------------|-------------:|------------:|------------:|--------:|
| selection_rate |          0.4 |         0.8 |         0.4 |     0.5 |

But when transposing it looks like: 

|            |   selection_rate |
|:-----------|-----------------:|
| difference |              0.4 |
| group_max  |              0.8 |
| group_min  |              0.4 |
| ratio      |              0.5 |

Personally, I see the transposed version as more useful so we could just make that one the version of the table that is generated. What do you think?

I'm fine changing the name of the parameter to `transpose` if we decide to keep it.

lazarust:
@BenjaminBossan Thanks for explaining this! I think I've addressed/responded to most of your comments.

I still need to handle scalars from the `MetricFrame` and update the doc string.

BenjaminBossan:
Great, thanks, please ping Adrin and me once it's ready for review.

lazarust:
@BenjaminBossan and @adrinjalali This is ready to review again!!

docs/changes.rst
```diff
@@ -11,9 +11,11 @@ skops Changelog
 
 v0.6
 ----
-- Added tabular regression example. :pr: `254` by `Thomas Lazarus`
+- Added tabular regression example. :pr:`254` by `Thomas Lazarus <lazarust>`.
```
```suggestion
- Added tabular regression example. :pr:`254` by
  :user:`Thomas Lazarus <lazarust>`.
```

same for the other one

skops/_min_dependencies.py
```diff
@@ -32,6 +32,7 @@
     "xgboost": ("1.6", "tests", None),
     # TODO: remove condition when catboost supports python 3.11
     "catboost": ("1.0", "tests", "python_version < '3.11'"),
+    "fairlearn": ("0.8.0", "docs, tests", None),
```
adrinjalali, line 35:
```suggestion
    "fairlearn": ("0.7.0", "docs, tests", None),
```

This should work as well.

skops/card/tests/test_card.py
```diff
@@ -1593,3 +1594,48 @@ def test_visibility_with_card_save(self, card):
...
+        metric_frame = metrics.MetricFrame(
+            y_true=y_true, y_pred=y_pred, sensitive_features=sex, metrics=metric_dict
+        )
+        card.add_fairlearn_metric_frame(
+            metric_frame=metric_frame,
+            transpose=transpose,
+            table_name="Metric Frame Table",
+        )
+
+        actual_table = card.select("Metric Frame Table").content.format()
```
adrinjalali, line 1625: love how the new `select` makes it so much easier to test things.

skops/card/_model_card.py
```diff
@@ -1155,6 +1155,46 @@ def add_permutation_importances(
 
         return self
 
+    def add_fairlearn_metric_frame(
+        self,
+        metric_frame,
+        table_name: str = "Fairlearn MetricFrame Table",
+        transpose=True,
+    ) -> Card:
```
adrinjalali, line 1163: should we be adding a `description` to our `add_*` methods? From the user's perspective, it seems odd that it's tricky to add a description here for sections. I think it'd make sense for all of them to allow a description as well as a title. WDYT? also cc @skops-dev/maintainers

skops/card/_model_card.py
```diff
@@ -1155,6 +1155,46 @@ def add_permutation_importances(
...
+            Whether to transpose the table or not.
+
+        table_name: str
+            The desired name of the table section in the model card.
+
+        Returns
+        -------
+        self: Card
+            The model card with the metric frame added.
+        """
```
adrinjalali, line 1183:
```suggestion

	Notes
	--------
	You can check `fairlearn's documentation
	<https://fairlearn.org/v0.8/user_guide/assessment/index.html>`__ on how to
	work with `MetricFrame`s.
        """
```

adrinjalali (commented):
This looks pretty good to me, and we should probably incorporate it in our existing examples, but that could be a separate PR.

skops/card/_model_card.py
BenjaminBossan: Yes, I agree it would be good to have, some existing methods also don't have it, e.g. `add_table`. It's probably easier to have a separate PR where this argument is added.

BenjaminBossan (commented):
Not much to add from my side. I think the transpose feature could be implemented without pandas, but it's not a big deal, most users will probably have pandas installed anyway.

adrinjalali:
> I think the transpose feature could be implemented without pandas, but it's not a big deal, most users will probably have pandas installed anyway.

`fairlearn` depends on `pandas` anyway.

BenjaminBossan:
> `fairlearn` depends on `pandas` anyway.

I didn't even think about that, then forget what I said :)

skops/card/_model_card.py
lazarust: @adrinjalali For clarification, should I add the `description` argument to this method in this PR? Or do it in another one?

lazarust:
@adrinjalali and @BenjaminBossan Thanks for reviewing this! I've addressed all the comments except for one I needed clarification on.

skops/card/_model_card.py
adrinjalali: we can keep it for another PR as @BenjaminBossan suggested.

BenjaminBossan: One more thing: Could you please change the type annotation to return `-> Self`? This is a change we recently made on all the other methods as well.

BenjaminBossan (changes requested):
Fantastic, just this minor comment, then we can merge.

lazarust:
@BenjaminBossan I've addressed your comment! Sorry, it took me a couple of days to get to this.

Issue #47, opened 2022-07-21
Remove token from hub_utils.tests.common.py and move to secrets


adrinjalali:
We should remove the token from the repo and move it to secrets, but we should make sure it works on PRs from forks.

merveenoyan:
@BenjaminBossan would you be up for this one?

BenjaminBossan:
I would be up but I have no idea yet how to actually solve the problem.

adrinjalali:
It's possible to have this work in PRs by doing some magic in GH actions, but then running the tests for users would require them to give a token. On the other hand, it's already the case that users would need to pass their token to create the docs. So maybe that's not a big issue?

BenjaminBossan:
> On the other hand, it's already the case that users would need to pass their token to create the docs

They could still use the checked in token for that.

> running the tests for users would require them to give a token

I don't really see a good solution for that. We could mark all tests that require a valid token (which amounts to the network tests I guess) and if users don't provide a token, skip them. If they make changes that don't require those tests, then we're fine, else they have to provide their own token . At the end of the day, all tests will be run on CI.

adrinjalali:
yeah I guess that works.

BenjaminBossan:
Just noticed we also haven the token [here](https://github.com/skops-dev/skops/blob/2c2dd6eddbd88c4774d297c387d1169533246b97/scripts/clean_skops.py#L12). This one could really be replaced by a GH secret, though I guess we would also need to invalidate it since it's already public, and as long as we still use the same token for tests, nothing is gained.

merveenoyan:
@BenjaminBossan AFAIK once you remove it on 🤗Hub it's invalidated. I think I don't have the user access to add tokens (I don't see the settings tab) maybe @adrinjalali can do it (and then we can replace the token with the secret).

adrinjalali:
yeah that one is the same as the one used by tests to create the repos. So changing it doesn't gain us anything.

adrinjalali:
Closing as `hub_utils` is deprecated away.

PR #303, opened 2023-02-20
DOC Add intelex inference example

ahuber21:
I want to give a self-contained example how scikit-learn-intelex speeds up model inference times. This effort was kicked off in #251 and the PR in this form is basically just to align the requirements.

I'm new to the discussion, so I'm including just the most basic things. Please let me know if this example is a step in the right direction.

Output from running on my machine:
```
$ python examples/use_intelex.py
Intel(R) Extension for Scikit-learn* enabled (https://github.com/intel/scikit-learn-intelex)
[skl] Inference took t_stock = 3.7e+00s and achieved 21.7% accuracy
[skl-ex] Inference took t_opt = 2.3e-01s and achieved 21.7% accuracy
t_stock / t_opt = 16.1
# ... more output about pushing and downloading ...
[skl] Inference took t_stock = 3.8e+00s and achieved 21.7% accuracy
[skl-ex] Inference took t_opt = 2.5e-01s and achieved 21.7% accuracy
t_stock / t_opt = 15.3
Intel(R) Extension for Scikit-learn* enabled (https://github.com/intel/scikit-learn-intelex)
[skl] Inference took t_stock = 3.8e+00s and achieved 21.7% accuracy
```

### DIFF
--- a/docs/examples.rst
+++ b/docs/examples.rst
@@ -15,3 +15,8 @@ Examples of interactions with the Hugging Face Hub
 - Text Classification:
     :ref:`sphx_glr_auto_examples_plot_text_classification.py` is an example of using skops to serialize a text 
     classification model and create a model card and a Hugging Face Hub repository.
+- Using Intel(R) Extension for scikit-learn:
+    :ref:`sphx_glr_auto_examples_plot_intelex.py` is an example of using
+    Intel(R) Extension for scikit-learn to speed up inference of classical
+    machine learning algorithms and how performance-optimized models work with
+    Hugging Face Hub.
--- a/examples/plot_intelex.py
+++ b/examples/plot_intelex.py
@@ -1,0 +1,277 @@
+"""
+Creating models that are accelerated by Intel(R) Extension for scikit-learn
+---------------------------------------------------------------------------
+
+Introduction
+============
+
+This guide demonstrates how under certain conditions, Intel(R) Extension for
+Scikit-learn (also ``scikit-learn-intelex``, or ``sklearnex``) can be used to
+speed up inference of Scikit-learn models.
+
+The extension supports most of Scikit-learn's classical machine learning
+algorithms, like k-nearest neighbors, support vector machines, linear/logistic
+regression, and more. Stock Scikit-learn implementations are used where no
+optimized version is available, making this package 100% compatible with
+existing code. Note while compatibility is assured by continuous testing,
+equivalence of results between the two packages is not guaranteed. In fact, due
+to independent implementations, intermediate results differ in many cases. An
+up-to-date list of supported algorithms can be found in the `official
+documentation <https://intel.github.io/scikit-learn-intelex/algorithms.html>`_.
+
+Intel(R) Extension for Scikit-learn accelerates Scikit-learn algorithms by using
+the latest hardware features and optimized caching and threading strategies.
+Find more details in Intel's blog posts on Medium (`1
+<https://medium.com/intel-analytics-software/save-time-and-money-with-intel-extension-for-scikit-learn-33627425ae4>`_,
+`2
+<https://medium.com/intel-analytics-software/accelerate-your-scikit-learn-applications-a06cacf44912>`_).
+In many cases, optimizations translate to hardware from other vendors, albeit
+with smaller performance gains.
+
+For this example, we train two simple
+:class:`sklearn.neighbors.KNeighborsClassifier` instances, one with and one
+without using ``sklearnex``, and compare inference times. Afterward, we upload
+both models to the Hugging Face Model Hub. Hugging Face Hub supports
+``sklearnex``-optimized models, meaning the achieved speedup will translate for
+Inference API users.
+"""
+
+# %%
+# Imports
+# =======
+# First, we import everything required for the rest of this document.
+
+import os
+import pickle
+from pathlib import Path
+from tempfile import NamedTemporaryFile, mkdtemp
+from time import perf_counter
+from uuid import uuid4
+
+from huggingface_hub import delete_repo, whoami
+from sklearn.datasets import make_classification
+from sklearn.metrics import log_loss
+from sklearn.model_selection import train_test_split
+from sklearn.neighbors import KNeighborsClassifier
+from sklearnex import patch_sklearn
+from sklearnex.neighbors import KNeighborsClassifier as KNeighborsClassifierOptimized
+
+from skops import card, hub_utils
+
+# %%
+# Data
+# ====
+# Next, we create some generic data. A dataset of 50k rows x 15 columns is big enough
+# to showcase a performance gain from using ``sklearnex``. Generally speaking,
+# larger datasets will benefit more from the ``sklearnex`` optimizations. More
+# details can be found in the official
+# `README <https://github.com/intel/scikit-learn-intelex/blob/master/README.md>`_.
+X, y = make_classification(
+    n_samples=50_000,
+    n_features=15,
+    n_informative=15,
+    n_redundant=0,
+    n_clusters_per_class=1,
+    shuffle=False,
+    random_state=42,
+)
+X_train, X_test, y_train, y_test = train_test_split(
+    X, y, test_size=0.3, random_state=42
+)
+
+# %%
+# Training the stock model
+# ========================
+# Now we can train a stock Scikit-learn
+# :class:`sklearn.neighbors.KNeighborsClassifier`
+
+clf = KNeighborsClassifier(3)
+start = perf_counter()
+clf.fit(X_train, y_train)
+print(f"Training finished in {perf_counter() - start:.2f}s")
+
+# %%
+# Training the optimized model
+# ============================
+# Now we fit the optimized algorithm. Note, that rather than loading the model
+# from ``sklearnex``, we could also load and call ``patch_sklearn()``. Find more
+# details in the `documentation
+# <https://intel.github.io/scikit-learn-intelex/#usage>`_.
+
+clf_opt = KNeighborsClassifierOptimized(3)
+start = perf_counter()
+clf_opt.fit(X_train, y_train)
+print(f"Training finished in {perf_counter() - start:.2f}s")
+
+# %%
+# We are not comparing the k-NN fit times, since this is not a compute-intensive
+# task and both are typically very fast.
+
+# %%
+# Comparing inference times
+# =========================
+# Now to the interesting part: We measure the execution time of
+# ``predict_proba()`` for the two models.
+
+start = perf_counter()
+y_proba = clf.predict_proba(X_test)
+t_stock = perf_counter() - start
+
+log_loss_score = log_loss(y_test, y_proba)
+print(
+    f"[stock scikit-learn] Inference took t_stock = {t_stock:.2f}s with a "
+    f"log-loss score of {log_loss_score:.1%}"
+)
+
+start = perf_counter()
+y_proba = clf_opt.predict_proba(X_test)
+t_opt = perf_counter() - start
+
+log_loss_score = log_loss(y_test, y_proba)
+print(
+    f"[sklearnex] Inference took t_opt = {t_opt:.2f}s with a log-loss score of"
+    f" {log_loss_score:.1%}"
+)
+
+print(f"t_stock / t_opt = {t_stock/t_opt:.1f}")
+
+# %%
+# We see that inference using ``sklearnex`` is a lot faster while achieving the
+# same log-loss score.
+
+# %%
+# Save and upload the models
+# ==========================
+# Let's save all required files to disk and initialize Hugging Face Model Hub
+# repositories.
+
+# replace with your own token or set it as an environment variable before
+# running the script
+token = os.environ["HF_HUB_TOKEN"]
+
+with NamedTemporaryFile(mode="bw", prefix="stock-", suffix=".pkl") as fp:
+    pickle.dump(clf, file=fp)
+
+    stock_repo = mkdtemp(prefix="stock-")
+    hub_utils.init(
+        model=fp.name,
+        requirements=["scikit-learn=1.2.1"],
+        dst=stock_repo,
+        task="tabular-classification",
+        data=X_test,
+    )
+
+
+with NamedTemporaryFile(mode="bw", prefix="opt-", suffix=".pkl") as fp:
+    pickle.dump(clf_opt, file=fp)
+
+    opt_repo = mkdtemp(prefix="opt-")
+    hub_utils.init(
+        model=fp.name,
+        requirements=["scikit-learn=1.2.1", "scikit-learn-intelex=2023.0.1"],
+        dst=opt_repo,
+        task="tabular-classification",
+        data=X_test,
+        use_intelex=True,
+    )
+
+# Create Model cards with the most basic information
+clf_card = card.Card(clf, metadata=card.metadata_from_config(Path(stock_repo)))
+clf_card.metadata.license = "mit"
+limitations = "This model is not ready to be used in production."
+model_description = (
+    "This is a `KNeighborsClassifier` model trained on synthetic data. It is "
+    "trained with the stock scikit-learn algorithm and part of a "
+    "demonstration, showing how Intel(R) Extension for scikit-learn can be "
+    "used to speed up model inference times."
+)
+model_card_authors = "skops_user"
+citation_bibtex = "**BibTeX**\n\n```\n@inproceedings{...,year={2020}}\n```"
+clf_card.add(
+    **{
+        "Citation": citation_bibtex,
+        "Model Card Authors": model_card_authors,
+        "Model description": model_description,
+        "Model description/Intended uses & limitations": limitations,
+    }
+)
+clf_card.save(Path(stock_repo) / "README.md")
+
+clf_opt_card = card.Card(clf_opt, metadata=card.metadata_from_config(Path(opt_repo)))
+model_description = (
+    "This is a `KNeighborsClassifier` model trained on synthetic data. It is "
+    "trained with the Intel(R) extension for scikit-learn optimized version of "
+    "the algorithm, and part of a demonstration, showing how Intel(R) "
+    "Extension for scikit-learn can be used to speed up model inference times."
+)
+clf_card.add(
+    **{
+        "Citation": citation_bibtex,
+        "Model Card Authors": model_card_authors,
+        "Model description": model_description,
+        "Model description/Intended uses & limitations": limitations,
+    }
+)
+clf_opt_card.save(Path(opt_repo) / "README.md")
+
+# Push everything to the Model hub
+user_name = whoami(token=token)["name"]
+uuid = uuid4()
+repo_id_stock = f"{user_name}/knn-example-stock-{uuid}"
+repo_id_opt = f"{user_name}/knn-example-intelex-{uuid}"
+
+print(f"Pushing skl model to: {repo_id_stock}")
+hub_utils.push(
+    repo_id=repo_id_stock,
+    source=stock_repo,
+    token=token,
+    commit_message="Add scikit-learn KNN model example",
+    create_remote=True,
+    private=False,
+)
+print(f"Pushing sklearnex model to: {repo_id_opt}")
+hub_utils.push(
+    repo_id=repo_id_opt,
+    source=opt_repo,
+    token=token,
+    commit_message="Add scikit-learn-intelex KNN model example",
+    create_remote=True,
+    private=False,
+)
+
+
+# %%
+# Loading non-optimized models
+# ============================
+# It is possible to load non-optimized models even after Intel(R) optimizations
+# were loaded with ``patch_sklearn()``. Note, however, that this will not result
+# in faster inference times, since loading a persisted model will always load
+# the objects exactly as they were saved.
+
+patch_sklearn()
+
+with NamedTemporaryFile(mode="bw+") as fp:
+    pickle.dump(clf, file=fp)
+    fp.seek(0)
+    clf = pickle.load(fp)
+
+start = perf_counter()
+clf.predict_proba(X_test)
+t_stock = perf_counter() - start
+
+log_loss_score = log_loss(y_test, y_proba)
+print(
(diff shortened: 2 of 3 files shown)


examples/use_intelex.py
```diff
@@ -0,0 +1,222 @@
+"""
+Creating models that are accelerated by Intel(R) Extension for scikit-learn
+--------------------------------------------------------------------------------
+
+This guide demonstrates how Intel(R) Extension for scikit-learn (sklearnex) can
+be used to speed up inference of scikit-learn models.
+We train two simple :class:`KNeighborsClassifier` instances, one with and one
```
BenjaminBossan, line 7: :class:`KNeighborsClassifier`

should probably be

    :class:`sklearn.neighbors.KNeighborsClassifier`

right now sphinx doesn't link it.

examples/use_intelex.py
```diff
@@ -0,0 +1,222 @@
...
+    f"[skl-ex] Inference took t_opt = {t_opt:.2f}s and achieved {accuracy:.1%} accuracy"
+)
+
+print(f"t_stock / t_opt = {t_stock/t_opt:.1f}")
+
+# %%
+# Loading non-optimized models
+# ============================
+# It is possible to load non-optimized models even after Intel(R) optimizations
+# were loaded with `path_sklearn()`. Note, however, that this will not result
```
BenjaminBossan, line 206:
```suggestion
# were loaded with `patch_sklearn()`. Note, however, that this will not result
```

examples/use_intelex.py
```diff
@@ -0,0 +1,222 @@
+"""
+Creating models that are accelerated by Intel(R) Extension for scikit-learn
+--------------------------------------------------------------------------------
+
+This guide demonstrates how Intel(R) Extension for scikit-learn (sklearnex) can
+be used to speed up inference of scikit-learn models.
```
BenjaminBossan, line 6: Maybe if there is a link to a document showing what models are supported, it would be nice to link it here.

BenjaminBossan (changes requested):
Thanks for adding this example. This is not an in depth review yet, just what I found on first look:

The example is not actually being run but just included as is into the docs, as can be seen here: 

https://skops--303.org.readthedocs.build/en/303/auto_examples/use_intelex.html#sphx-glr-auto-examples-use-intelex-py

I'm not completely sure why that happens here but not with the other examples, Possibly, this example needs to be linked explicitly in the docs somewhere? A good place would be in:

https://github.com/skops-dev/skops/blob/3e1f138c5a7c12863ca263beb2804ceede07d35a/docs/examples.rst

(which would be a good addition regardless of whether that is the root of the issue or not)

In general, I think this guide could be more useful if it was made a bit more accessible. E.g. explain in a few words what intelex does, link to its docs, and avoid too many abbreviations like "skl" and "sklex".

ahuber21:
Hi @BenjaminBossan 
thanks for the feedback! I'm happy to hear that the original suggestions was already a step in the right direction.
I added your requested changes and also added some content to the introduction in fixup commits that we can squash later.
Let's see if the code actually runs this time. 

Let me know what else you want to see!

BenjaminBossan:
The example is still not building correctly but I believe I have figured out the issue. It seems like the file name should always start with `plot_`. Could you please try if that fixes the issue?

Btw. you can build the docs locally by going into `docs/` and running `make html`, then open the html files in `_build/html`.

ahuber21:
Hi @BenjaminBossan thanks for the tip. The latest commit worked locally so I hope this time the example runs.

I had an interesting thought. At the bottom of my example I show that loading a pickled non-optimized model has no speedup, even when `sklearnex` is loaded. I'm curious if that is also true for `skops.io` methods. Essentially, if a new object is created, i.e. the constructor is called, Intel optimizations should apply. It would be interesting to see if there are any compatibility issues.
I think @adrinjalali already looked into that, maybe he can comment.

BenjaminBossan:
> The latest commit worked locally so I hope this time the example runs.

The doc build is failing, but it's still progress :) The trouble now is that building the docs has a dependency on sklearnex, which is not installed. Could you please add it to the [dependencies](https://github.com/skops-dev/skops/blob/main/skops/_min_dependencies.py) as a "doc" dependency?

> I'm curious if that is also true for `skops.io` methods. Essentially, if a new object is created, i.e. the constructor is called, Intel optimizations should apply.

I think this will depend on how exactly the patching works, which I don't know. Ideally, there would be an easy way to check on each estimator if it was patched, something like `hasattr(estimator, "_uses_intelex")` or `sklearnex.check_is_patched(estimator)`, I believe this would be really useful.

ahuber21:
Progress indeed. The `sklearnex.check_is_patched(estimator)` suggestion is very good. I'm wondering if we already have something like this. Let me ping @napetrov, maybe he knows more.

ahuber21:
By the way, the pipeline now complains about the missing HF token. I followed the approach that I found in `plot_hf_hub.py`, so I think the error is nothing new. How is this expected to work?

examples/plot_intelex.py
```diff
@@ -0,0 +1,273 @@
...
+``scikit-learn-intelex``, or ``sklearnex``) can be used to speed up inference of
+Scikit-learn models.
+
+The extension supports most of Scikit-learn's classical machine learning
+algorithms, like k-nearest neighbors, support vector machines, linear/logistic
+regression, and more. Stock Scikit-learn implementations are used where no
+optimized version is available, making this package 100% compatible with
+existing code. An up-to-date list of supported algorithms can be found in the
+`official documentation
+<https://intel.github.io/scikit-learn-intelex/algorithms.html>`_.
```
BenjaminBossan, line 18: Is there any check in sklearnex that a patched model produces the exact same output? So e.g. can I assume that `np.allclose(clf.predict_proba(X), clf_opt.predict_proba(X))`? Or is it more like the algorithm is the same but results can be different, e.g. due to random initialization? If it's the former, it would be nice to mention that.

examples/plot_intelex.py
```diff
@@ -0,0 +1,273 @@
...
+clf_opt.fit(X_train, y_train)
+
+# %%
+# Comparing inference times
+# =========================
+# Now to the interesting part: We measure the execution time of
+# ``predict_proba()`` for the two models.
+
+start = perf_counter()
+prediction = clf.predict_proba(X_test)
```
BenjaminBossan, line 108: Let's rename `prediction` to `y_proba`, which is the canonical name for the output from `predict_proba`. The same for variables further below.

examples/plot_intelex.py
```diff
@@ -0,0 +1,273 @@
...
+t_stock = perf_counter() - start
+
+accuracy = log_loss(y_test, prediction)
```
BenjaminBossan, line 111: Calling this "accuracy" is very confusing. I was surprised to find an "accuracy" of 21% on a binary classification task :D Let's either call it "log_loss_score" or something like that, or calculate the actual accuracy and not log loss.

examples/plot_intelex.py
```diff
@@ -0,0 +1,273 @@
...
+print(
+    f"[sklearnex] Inference took t_opt = {t_opt:.2f}s and achieved"
+    f" {accuracy:.1%} accuracy"
+)
+
+print(f"t_stock / t_opt = {t_stock/t_opt:.1f}")
+
+# %%
+# We see that inference using ``sklearnex`` is a lot faster while producing the
+# same result.
```
BenjaminBossan, line 131: Finding the "same result" doesn't really tell us much here IMHO. This is KNN with 3 neighbors, so each probability is either 0.0, 0.33, 0.67, or 1.0. Since there are only 4 discrete values, it's not too surprising to get the exact same output. The argument would be stronger with something like a linear regression.

In the end, however, this comes back to what I remarked earlier, whether we can expect the exact same output after patching or not. Clarifying this would help a lot.

examples/plot_intelex.py
```diff
@@ -0,0 +1,273 @@
...
+
+# %%
+# Save and upload the models
+# ==========================
+# Let's save all required files to disk and initialize Hugging Face Model Hub
+# repositories.
```
BenjaminBossan, line 137: It could be worth mentioning that sklearnex will be used on the Hub, e.g. for the Inference API. This might not be obvious to everyone.

BenjaminBossan (changes requested):
Thanks Andreas for adding this example. I have now given it a more thorough review. Overall it looks good, but I think a few things could be improved further. Please take a look at my comments.

> By the way, the pipeline now complains about the missing HF token. I followed the approach that I found in plot_hf_hub.py, so I think the error is nothing new. How is this expected to work?

Yes, that's unfortunately expected, see #47. As soon as this is merged, it should work, as it would use the HF token set up for this repo. When I tested this locally with my token, it worked.

examples/plot_intelex.py
```diff
@@ -0,0 +1,273 @@
...
+
+The extension supports most of Scikit-learn's classical machine learning
+algorithms, like k-nearest neighbors, support vector machines, linear/logistic
+regression, and more. Stock Scikit-learn implementations are used where no
+optimized version is available, making this package 100% compatible with
+existing code. An up-to-date list of supported algorithms can be found in the
+`official documentation
+<https://intel.github.io/scikit-learn-intelex/algorithms.html>`_.
+
+Intel(R) Extension for Scikit-learn uses the `Intel(R) oneAPI Data Analytics
```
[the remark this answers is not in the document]
ahuber21: I basically copied a paragraph from an older Medium blog post. In the updated version it is replaced by a more high level description and links to two Medium blog posts (for anyone who needs more details). I first wasn't sure if I'm supposed to link to Medium in this example, but it might just be the best solution.

examples/plot_intelex.py
ahuber21: sklearnex's backend is completely separate from scikit-learn, including a separate RNG. So there's no guarantee that identical outputs are achieved and, in fact, they're mostly different.
The agreement of model outputs is however tested in our CI https://github.com/intel/scikit-learn-intelex/blob/master/tests/test_examples.py

I'll mention this in the example

examples/plot_intelex.py
ahuber21: Hmm you're right. At some point while condensing the initial example down to something quick and easy this part lost all its meaning. If you would like to see something else I'm happy to make the change. I mean, it's at least showing that (a) there is a result and it actually works, and (b) it's not totally buggy. That, in combination with the comment I'll add in the next commit may be enough. Let me know what you think

ahuber21:
Hi Benjamin. Thanks! I hope all your comments are addressed with the latest commit.

examples/plot_intelex.py
```diff
@@ -0,0 +1,276 @@
...
+# %%
+# Data
+# ====
+# Next, we create some generic data. A dataset of 50k rows x 15 columns is big enough
+# to showcase a performance gain from using ``sklearnex``. Generally speaking,
+# larger datasets will benefit more from the ``sklearnex`` optimizations. More
+# details can be found in the official
+# `README <https://github.com/intel/scikit-learn-intelex/blob/master/README.md>`_.
+X, y = make_classification(
+    n_samples=50_000,
```
adrinjalali, line 66: we'd need to see how long it takes on the CI machines, might need to reduce the number if it's too slow.

examples/plot_intelex.py
```diff
@@ -0,0 +1,276 @@
...
+# Training the optimized model
+# ============================
+# We apply ``patch_sklearn()`` and reimport the model to load the patched
+# version. A message is shown, telling us that Intel(R) Extension for
+# Scikit-learn* has been enabled.
+patch_sklearn()
+from sklearn.neighbors import KNeighborsClassifier
+
+clf_opt = KNeighborsClassifier(3)
+clf_opt.fit(X_train, y_train)
```
adrinjalali, line 99: I think in a script, this is very confusing. It'd be nicer if we explicitly import the class from sklearnex for it to be clear which one it is, and leave a comment that users don't have to change imports, and they can only add the call to patch on top of the file before they run imports.

We can also here show training times and compare them.

examples/plot_intelex.py
```diff
@@ -0,0 +1,276 @@
...
+    f" {log_loss_score:.1%}"
+)
+
+print(f"t_stock / t_opt = {t_stock/t_opt:.1f}")
+
+# %%
+# We see that inference using ``sklearnex`` is a lot faster while achieving the
+# same log-loss score. In fact, the agreement between stock Scikit-learn and
+# ``sklearnex`` model outputs is guaranteed by CI tests, see
+# `here <https://github.com/intel/scikit-learn-intelex/blob/master/tests/test_examples.py>`_.
```
adrinjalali, line 133: I don't think it makes sense to link to a test file on another repo, we should link to a documentation page.

I didn't read the test file, but since the training procedures are different, the outputs can't be guaranteed to be the same. They can only be tested to be the same on a few toy examples, which doesn't guarantee equivalence.

examples/plot_intelex.py
```diff
@@ -0,0 +1,276 @@
...
+# Save and upload the models
+# ==========================
+# Let's save all required files to disk and initialize Hugging Face Model Hub
+# repositories.
+
+with open("model-stock.pkl", mode="bw") as fp:
+    pickle.dump(clf, file=fp)
+
+with open("model-optimized.pkl", mode="bw") as fp:
+    pickle.dump(clf_opt, file=fp)
```
adrinjalali, line 145: please use a tempfile rather than saving in the current directory.

examples/plot_intelex.py
```diff
@@ -0,0 +1,276 @@
...
+    data=X_test,
+    use_intelex=True,
+)
+
+# Create Model cards with the most basic information
+clf_card = card.Card(clf, metadata=card.metadata_from_config(Path(stock_repo)))
+clf_card.save(Path(stock_repo) / "README.md")
+
+clf_opt_card = card.Card(clf_opt, metadata=card.metadata_from_config(Path(opt_repo)))
+clf_opt_card.save(Path(opt_repo) / "README.md")
```
adrinjalali, line 176: we shouldn't have empty cards from our own examples, please include a few basic info about the model and the data and the performance here.

examples/plot_intelex.py
```diff
@@ -0,0 +1,276 @@
...
+    commit_message="Add scikit-learn-intelex KNN model example",
+    create_remote=True,
+    private=False,
+)
+
+# %%
+# Download and use the models from Model Hub
+# ==========================================
+# We download the models again and check whether our achieved performance boost
+# survived the round trip to the Hub.
```
adrinjalali, line 207: do we need to test this? I think this is well tested in other places

examples/plot_intelex.py
```diff
@@ -0,0 +1,276 @@
...
+
+# %%
+# We see that the model performance is still the same.
+
+# %%
+# Loading non-optimized models
+# ============================
+# It is possible to load non-optimized models even after Intel(R) optimizations
+# were loaded with ``patch_sklearn()``. Note, however, that this will not result
+# in faster inference times.
```
adrinjalali, line 251:
```suggestion
# in faster inference times, since loading a persisted model will always load the objects
# exactly as they were saved.
```

examples/plot_intelex.py
```diff
@@ -0,0 +1,276 @@
...
+
+
+# %%
+# Delete Repository
+# =================
+# At the end, we can delete the created repositories again using
+# ``delete_repo``. For more information please refer to the
+# documentation of ``huggingface_hub`` library.
+delete_repo(repo_id=repo_id_stock)
+delete_repo(repo_id=repo_id_opt)
```
adrinjalali, line 276: need to pass the token as well.

skops/_min_dependencies.py
```diff
@@ -11,6 +11,7 @@
 #     "tomli": ("1.1.0", "install", "python_full_version < '3.11.0a7'"),
 dependent_packages = {
     "scikit-learn": ("0.24", "install", None),
+    "scikit-learn-intelex": ("2023.0.1", "docs", None),
```
adrinjalali, line 14: do we really need a very recent release here? These are minimum version dependencies.

adrinjalali (commented):
Thanks for the PR @ahuber21 , it's in a pretty good shape.

I left a few minor comments. I haven't made sure the line lengths are correct. You don't need to "accept suggestion" here, you can take the texts and apply them the way you like.

examples/plot_intelex.py
ahuber21: On my machine the bottleneck is up- and downloading the models. I'm optimistic all the code will run in <10s.

examples/plot_intelex.py
ahuber21: I changed the imports. Training for kNN is <1s and in many cases `sklearnex` comes with a few percent of overhead. We're working on that, but for now I think it wouldn't help to add it.

examples/plot_intelex.py
ahuber21: sklearnex is tested on the stock scikit-learn tests and is therefore "as tested as scikit-learn". Of course the implementations differ and it is impossible to reproduce the computations exactly (which isn't the goal in the first place). But all algorithms are required to produce results that are in very good agreement (typically the diff is required to be <1E-5).
We can definitely discuss how this is phrased, but the overall message should be that it is save to use sklearnex as a drop-in replacement and that statistically equivalent predictions will be made from sklearnex algos.

examples/plot_intelex.py
```diff
@@ -0,0 +1,278 @@
...
+    f" {log_loss_score:.1%}"
+)
+
+print(f"t_stock / t_opt = {t_stock/t_opt:.1f}")
+
+# %%
+# We see that inference using ``sklearnex`` is a lot faster while achieving the
+# same log-loss score. In fact, the agreement between stock Scikit-learn and
+# ``sklearnex`` model outputs is guaranteed by CI tests, see
+# `here <https://github.com/intel/scikit-learn-intelex/blob/master/tests/test_examples.py>`_.
```
napetrov, line 134: Think that keeping this simple would make sense 
```suggestion
# We see that inference using ``sklearnex`` is a lot faster while achieving the
# same log-loss score.
```

examples/plot_intelex.py
ahuber21: Wanted to demonstrate that we can download and use optimized models. Should I remove it?

skops/_min_dependencies.py
ahuber21: tbh: I don't know, so I added what I tested. We can go with a 1y old release. As far as I can tell this should work too. But I wouldn't want to put an even older release there.

examples/plot_intelex.py
adrinjalali: in that case, it does help to add it, since right now we're only selectively showing data, not the whole truth, and that's not what we wanna do here.

examples/plot_intelex.py
adrinjalali: I don't mind, as you wish. I'm not sure if it's adding to the message here though.

adrinjalali:
Build failing, once you're done with edits, please ping me for a second review.

ahuber21:
Don't quite understand what happened with the ubuntu tests. To me it looked unrelated. Anyway, I just pushed an update without the "Download model and re-evaluate" section, so there'll be a new run.
From my side, I've added everything. Let me know what you think

examples/plot_intelex.py
ahuber21: removed

examples/plot_intelex.py
ahuber21: I have added a `perf_counter` for the fit stage for completeness. On my machine, both fit times were ~0.05s. The numbers printed, but doing anything else with them (like calculating ratios) won't help if we don't do multiple runs. I hope you're fine with how it is now. Let me know :)

examples/plot_intelex.py
```diff
@@ -0,0 +1,278 @@
...
+# <https://intel.github.io/scikit-learn-intelex/#usage>`_.
+
+clf_opt = KNeighborsClassifierOptimized(3)
+start = perf_counter()
+clf_opt.fit(X_train, y_train)
+print(f"Training finished in {perf_counter() - start:.2f}s")
+
+# %%
+# We are not comparing the k-NN fit times, since this is not a compute-intensive
+# task and typically very fast.
```
adrinjalali, line 109:
```suggestion
# We are not comparing the k-NN fit times, since this is not a compute-intensive
# task and both are typically very fast.
```

docs/examples.rst
```diff
@@ -15,3 +15,6 @@ Examples of interactions with the Hugging Face Hub
 - Text Classification:
     :ref:`sphx_glr_auto_examples_plot_text_classification.py` is an example of using skops to serialize a text 
     classification model and create a model card and a Hugging Face Hub repository.
+- Using Intel(R) Extension for scikit-learn:
+    :ref:`sphx_glr_auto_examples_plot_intelex.py` is an example of using Intel(R) Extension for scikit-learn to speed up
+    inference of classical machine learning algorithms and how performance-optimized models work with Hugging Face Hub.
```
adrinjalali, line 20:
```suggestion
- Using Intel(R) Extension for scikit-learn:
    :ref:`sphx_glr_auto_examples_plot_intelex.py` is an example of using
    Intel(R) Extension for scikit-learn to speed up inference of classical
    machine learning algorithms and how performance-optimized models work with
    Hugging Face Hub.
```

(just wrapping lines)

adrinjalali (commented):
Other than the nits, LGTM. WDYT @BenjaminBossan

ahuber21:
Force-pushed again because in the intro I still said that we download the models again. Nothing else changed.

adrinjalali:
@ahuber21 we squash and merge anyway, please avoid force pushing since it makes reviewing changes harder :)

BenjaminBossan (approved):
Yes, looks good, thanks Andreas

ahuber21:
> I think this will depend on how exactly the patching works, which I don't know. Ideally, there would be an easy way to check on each estimator if it was patched, something like hasattr(estimator, "_uses_intelex") or sklearnex.check_is_patched(estimator), I believe this would be really useful.

@BenjaminBossan this feature is now added in the latest release.

```python
from sklearnex import is_patched_instance
my_instance = pickle.load(open("some_file.pkl", "rb"))
is_patched_instance(my_instance)  # true if it was trained using sklearnex, false otherwise 
```

Issue #302, opened 2023-02-17
Add a method to model cards to generate an overview of sections


BenjaminBossan:
This would be nice to have.

Right now, when working with model cards, it's not so easy to get a good overview of what sections currently do or don't exist. The `repr` of the object gives some indication, but besides sections, it also displays the contents, which is distracting. It also truncates long strings, which can obscure the section names.

Having a nice overview of the section names, especially when it comes to nested subsections, would be really helpful. Especially when I want to add a subsection to an existing subsection, since I need to know the full name to the subsection, it requires some digging right now.

For now, I'm using something like this:

```python
from skops.card._model_card import Section

def iterate_key_section_content(
    data: dict[str, Section],
    parent_section: str = "",
    parent_keys: list[str] | None = None,
    level: int = 0,
):
    parent_keys = parent_keys or []

    for key, val in data.items():
        if not getattr(val, "visible", True):
            continue

        if parent_section:
            title = "/".join((parent_section, val.title))
        else:
            title = val.title

        yield title, level

        if val.subsections:
            yield from iterate_key_section_content(
                val.subsections,
                parent_section=title,
                parent_keys=parent_keys + [key],
                level=level + 1,
            )
```

With that function, I can do something like this:

```
>>> for title, level in iterate_key_section_content(model_card._data):
...     print(title)
Model description
Model description/Intended uses & limitations
Model description/Training Procedure
Model description/Training Procedure/Hyperparameters
Model description/Training Procedure/Model Plot
Model description/Evaluation Results
How to Get Started with the Model
Model Card Authors
Model Card Contact
Citation
```

lazarust:
@BenjaminBossan I can take this if it's not already taken!

BenjaminBossan:
@lazarust Yes, feel free to give it a shot. Use this thread to discuss designs.

lazarust:
@BenjaminBossan Here are some of my ideas. I'm thinking this should live in `skops/utils/modelutils.py`. For printing the structure I'm thinking of something like 

```
Model description
 |----Intended uses & limitations
 |----Training Procedure
 |          |----Hyperparameters
 |          |----Model Plot
 |----Evaluation Results
How to Get Started with the Model
Model Card Authors
Model Card Contact
Citation
```

I'm still thinking through how to dynamically generate this, but what're your thoughts?

lazarust:
[...]

PR #305, opened 2023-02-27
`ENH` Adds ability to create table of contents

lazarust:
Closes: #302

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -14,6 +14,8 @@ v0.6
 - Added tabular regression example. :pr: `254` by `Thomas Lazarus`
 - All public ``scipy.special`` ufuncs (Universal Functions) are trusted by default
   by :func:`.io.load`. :pr:`295` by :user:`Omar Arab Oghli <omar-araboghli>`.
+- Add :func:`.Card.create_toc` to create a table of contents for the model card in markdown format.
+  :pr:`305` by :user:`Thomas Lazarus <lazarust>`.
 
 v0.5
 ----
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -14,8 +14,8 @@ v0.6
 - Added tabular regression example. :pr: `254` by `Thomas Lazarus`
 - All public ``scipy.special`` ufuncs (Universal Functions) are trusted by default
   by :func:`.io.load`. :pr:`295` by :user:`Omar Arab Oghli <omar-araboghli>`.
-- Add :func:`.Card.create_toc` to create a table of contents for the model card in markdown format.
-  :pr:`305` by :user:`Thomas Lazarus <lazarust>`.
+- Add :func:`.Card.create_toc` to create a table of contents for the model card in
+  markdown format. :pr:`305` by :user:`Thomas Lazarus <lazarust>`.
 - Add example of using model card without the skops template. :pr:`291` by
   `Benjamin Bossan`_.
 
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -1321,7 +1321,6 @@ class Card:
         data: dict[str, Section],
         level: int = 0,
     ):
-        """Iterate through the key sections and yield the title and level."""
         for key, val in data.items():
             if not getattr(val, "visible", True):
                 continue
@@ -1337,6 +1336,7 @@ class Card:
 
     def create_toc(self) -> str:
         """Create a table of contents for the model card.
+
 
         Returns
         -------
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -1338,6 +1338,6 @@ class Card:
         """
         sections = []
         for title, level in self._iterate_key_section_content(self._data):
-            sections.append(f"{'    ' * level}- [{title}](#{title})")
+            sections.append(f"{'  ' * level}- {title}")
 
         return "\n".join(sections)
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -1321,6 +1321,7 @@ class Card:
         data: dict[str, Section],
         level: int = 0,
     ):
+        """Iterate through the key sections and yield the title and level."""
         for key, val in data.items():
             if not getattr(val, "visible", True):
                 continue
@@ -1336,7 +1337,6 @@ class Card:
 
     def create_toc(self) -> str:
         """Create a table of contents for the model card.
-
 
         Returns
         -------
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -1321,7 +1321,30 @@ class Card:
         data: dict[str, Section],
         level: int = 0,
     ):
-        """Iterate through the key sections and yield the title and level."""
+        """Iterate through the key sections and yield the title and level.
+
+        Parameters
+        ----------
+        data : dict[str, Section]
+            The card data to iterate through. This is usually the sections and subsections.
+
+        level : int, optional
+            The level of the section, by default 0. This keeps track of subsections.
+
+        Returns
+        -------
+        table_of_contents : str
+            Example:
+                - Model description
+                    - Intended uses & limitations
+                    - Training Procedure
+                        - Hyperparameters
+                        - Model Plot
+                    - Evaluation Results
+                - How to Get Started with the Model
+                - Model Card Authors
+                - Model Card Contact
+        """
         for key, val in data.items():
             if not getattr(val, "visible", True):
                 continue
@@ -1335,7 +1358,7 @@ class Card:
                     level=level + 1,
                 )
 
-    def create_toc(self) -> str:
+    def get_toc(self) -> str:
         """Create a table of contents for the model card.
 
         Returns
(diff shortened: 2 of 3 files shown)


BenjaminBossan (changes requested):
Regarding [not needing an extra utility function](https://github.com/skops-dev/skops/issues/302#issuecomment-1446235048): you're right, we should keep it simple for now, this is sufficient.

Regarding your implementation, I think it won't work if you plan to use the anchors as they need to be lower-cased and may not contain white spaces ([SO](https://stackoverflow.com/a/16426829)). Also not sure if we need links at all, as a user can't do anything with those, except when they're rendered, but as is, the rendered model card won't contain the TOC.

Moreover, four spaces looks excessive to me. I think 2 spaces should be enough to be rendered correctly.

Finally, could you please add some tests?

lazarust:
@BenjaminBossan This is ready for you to review again!

skops/card/_model_card.py
```diff
@@ -1315,3 +1315,36 @@ def render(self) -> str:
...
+            yield title, level
+
+            if val.subsections:
+                yield from self._iterate_key_section_content(
+                    val.subsections,
+                    level=level + 1,
+                )
+
+    def create_toc(self) -> str:
+        """Create a table of contents for the model card.
```
BenjaminBossan, line 1338: I think we need some clarification on this method. As a user, I might think that calling this method will "create a toc" in the sense that it will be automatically added to the model card when it is saved or rendered. The text should therefore clarify that this method only returns the toc for convenience but doesn't modify the model card.

skops/card/_model_card.py
```diff
@@ -1315,3 +1315,36 @@ def render(self) -> str:
...
+
+
```
BenjaminBossan, line 1340:
```suggestion
```

skops/card/tests/test_card.py
```diff
@@ -1593,3 +1593,26 @@ def test_visibility_with_card_save(self, card):
...
+        card = Card(model=model)
+        return card
+
+    def test_toc(self, card):
+        card.add_model_plot()
+        card.add_hyperparams()
+        card.add_metrics(accuracy=0.1)
+        card.add_get_started_code()
+        toc = card.create_toc()
+        exptected_toc = (
```
BenjaminBossan, line 1611: This test would be easier to read if the expected result were not concatenated like this. E.g. something like this:

```
expected_toc = [
    "- Model description",
    "  - Intended uses & limitations,
    ...,
]
assert toc == "\n".join(expected_toc)
```

As is, it would be quite hard to spot if a section is missing for instance.

Also, could you please add a custom section with `model_card.add(...)` and delete an existing section with `model_card.delete(...)`, just to be sure?

skops/card/_model_card.py
```diff
@@ -1315,3 +1315,36 @@ def render(self) -> str:
...
         return "\n".join(self._generate_card())
+
+    def _iterate_key_section_content(
+        self,
+        data: dict[str, Section],
+        level: int = 0,
+    ):
+        for key, val in data.items():
+            if not getattr(val, "visible", True):
+                continue
```
BenjaminBossan, line 1326: Let's add a test to cover this line. This visibility thing is not something we really advertise yet, but still good to test. To make a section invisible, select it and set it's visibility to `False`:

```
section = model_card.select(<section name>)
section.visible = False
```

Then this section should not appear in the toc.

BenjaminBossan (changes requested):
From my point of view, we're pretty close now. Just a few small changes.

Maybe @adrinjalali you could also take a quick look to see if this feature makes sense?

skops/card/_model_card.py
```diff
@@ -1315,3 +1315,36 @@ def render(self) -> str:
             sections inserted.
         """
         return "\n".join(self._generate_card())
+
+    def _iterate_key_section_content(
+        self,
+        data: dict[str, Section],
+        level: int = 0,
+    ):
+        for key, val in data.items():
```
adrinjalali, line 1324: needs a docstring, not easy to understand what it does.

skops/card/_model_card.py
adrinjalali: Maybe instead of this design, we should have something line

```
	def include_toc(visible/expanded=True):
```

which adds the ToC during render time, and also allows it to be hidden/collapsed.

BenjaminBossan: The primary use case I had in mind when thinking about this is if I'm interactively building up my model card. I would like to get a quick overview of what is and isn't there. Adding a toc to the final model card would not help, as it would require me to render and read the md, which is cumbersome.

Now it could be interesting to _also_ have a method to add the toc to the card. But as mentioned elsewhere, at least on GH and HF Hub, the toc is created automatically, so it's not necessary there. We could still add it, but I wouldn't use a method, but an init argument (since the toc should be lazily created when the card is rendered).

lazarust: > We could still add it, but I wouldn't use a method, but an init argument (since the toc should be lazily created when the card is rendered).

I agree, if we're going to have it on the model card we should just add an argument.

docs/changes.rst
```diff
@@ -14,6 +14,8 @@ v0.6
 - Added tabular regression example. :pr: `254` by `Thomas Lazarus`
```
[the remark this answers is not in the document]
lazarust: That one should be fixed in this PR I did earlier #298

lazarust:
@BenjaminBossan and @adrinjalali I've addressed all your comments other than including the Table of Contents in the markdown

skops/card/_model_card.py
adrinjalali: in that case, why not a read-only property `toc` which would return the current `toc`?

BenjaminBossan: What would be the advantage of having a property compared to a method? I'd slightly prefer a method here, e.g. in case we want to later add arguments to it.

adrinjalali: I'm not sure what arguments you'd pass here, but if you want a method, then it should be `get_toc`

BenjaminBossan: Not sure either, but at least that leaves the door open. I'm fine with `get_toc`.

skops/card/_model_card.py
```diff
@@ -1315,3 +1315,36 @@ def render(self) -> str:
             sections inserted.
         """
         return "\n".join(self._generate_card())
+
+    def _iterate_key_section_content(
+        self,
+        data: dict[str, Section],
+        level: int = 0,
+    ):
+        """Iterate through the key sections and yield the title and level."""
```
adrinjalali, line 1324: please explain `data` and `level` as well.

skops/card/_model_card.py
adrinjalali: @lazarust could you please rename the method to `get_toc`?

skops/card/_model_card.py
```diff
@@ -1315,3 +1315,36 @@ def render(self) -> str:
...
+                    level=level + 1,
+                )
+
+    def create_toc(self) -> str:
+        """Create a table of contents for the model card.
+
+        Returns
+        -------
+        toc : str
+            The table of contents for the model card formatted as a markdown string.
```
adrinjalali, line 1344: this is not clear as what the output would be, a tiny example would be nice here.

lazarust:
@adrinjalali I've addressed your comments! Sorry, it took a while!

skops/card/_model_card.py
```diff
@@ -1315,3 +1315,59 @@ def render(self) -> str:
             sections inserted.
         """
         return "\n".join(self._generate_card())
+
+    def _iterate_key_section_content(
```
adrinjalali, line 1319: I wonder if this should be a local func inside `get_toc`

skops/card/_model_card.py
```diff
@@ -1315,3 +1315,59 @@ def render(self) -> str:
...
+            yield title, level
+
+            if val.subsections:
+                yield from self._iterate_key_section_content(
+                    val.subsections,
+                    level=level + 1,
+                )
+
+    def get_toc(self) -> str:
+        """Create a table of contents for the model card.
```
adrinjalali, line 1362:
```suggestion
        """Get the table of contents for the model card.
```

skops/card/_model_card.py
```diff
@@ -1315,3 +1315,59 @@ def render(self) -> str:
...
+
+        Returns
+        -------
+        toc : str
+            The table of contents for the model card formatted as a markdown string.
+        """
```
adrinjalali, line 1368: I think the example output should be here in the public method doc, for the users to see how it will look like.

adrinjalali (commented):
Thanks @lazarust

skops/card/_model_card.py
BenjaminBossan: I think it's better to have it as a private method, because I can imagine other uses for it than just for `get_toc`.

lazarust:
@adrinjalali and @BenjaminBossan This is ready for y'all to look at again!

adrinjalali (approved):
Thanks @lazarust , I'll let @BenjaminBossan have the final look.

BenjaminBossan (approved):

BenjaminBossan:
Thanks a lot!

PR #307, opened 2023-02-27
FEAT Add skops space creator app

BenjaminBossan:
## Description

This is the code for a streamlit app whose goal it is to let users easily build HF spaces for sklearn models. In particular, it offers a GUI to interface with our `skops.card.Card` class, giving users a wysiwyg experience.

The app is currently hosted on my personal account:

https://huggingface.co/spaces/BenjaminB/skops-model-card-creator2

We want to move it to skops and update it automatically.

## Comments

I have adopted the CI to create this space using `skops-ci`. If the space itself is faulty, the CI job will still finish successfully, but at least one can check it and do some manual testing.

Adding automatic tests will prove quite difficult, because, AFAICT, streamlit doesn't have anything like a testing client. Automatic testing would require some heavy lifting from the likes of Selenium or Playwright.

It is possible to write tests for the parts of the code that are streamlit-agnostic, e.g. in `tasks.py`. However, that's the least brittle part of the code, so the additional value is low.

I had some strange issues with pre-commit and mypy. I don't want mypy to check the streamlit app -- it is a bit hacky in some places, which results in mypy errors. Therefore, I added an exclusion rule to `pyproject.toml`, which is indeed respected when I run mypy manually. However, pre-commit will still check the space, even though it's told to use the `pyproject.toml`. Thus I had to exclude the space code explicitly in the `.pre-commit-config.yaml`. If anyone knows what the issue is, please let me know.

## TODO

Right now, there is no job that automatically builds this space and pushes it to the sklearn orga. I assume for that, we should create a workflow similar to `clean-skops-user.yml` that runs regularly and that uses a secret token so that random people cannot override the app.


adrinjalali:
So this is now importing the token. It's odd cause the effective change is adding `-e`. Which suggests somehow the installation is odd when `-e` is missing. I tried locally with a new env, installed skops from pypi, and the test files exist and I can import the token. I'm not sure why it doesn't work on the CI.

BenjaminBossan:
@skops-dev/maintainers I made all changes as discussed here and with Adrin on Slack. Whether the "main" deployment works, we will only see once we merge to main :) The test space is deployed here:

https://huggingface.co/spaces/skops-ci/skops-space-creator-054a9631-14a8-47a0-ace1-f914c62ae3dc

Before going further, some points:

1. some changes in #310 may affect the space (I think: facilitate some stuff). It's probably easier to wait for it to land first
2. I'm not happy with the name "skops space creator", any better suggestions are welcome
3. do we already want to document this space (+ the converter space) in the docs or not yet?

BenjaminBossan:
> There's a warning on the bottom of the space you link here:

That's what I mentioned about flakiness of streamlit, refreshing should remove the error, and everything works despite of it. I add the button with that key [only once](https://github.com/BenjaminBossan/skops/blob/84a8eb0db506b6e19bf6bb14f5511483bdb0ef35/spaces/skops_space_creator/start.py#L164-L188), I really don't have an idea how the key is duplicated (would the button not appear twice then??).

> To me it's more of a scikit-learn model card creator/editor, or skops model card manager or something. It's not creating a space, it's creating a model card.

Although I agree that this the main purpose, it's also not quite all of it. First of all, you can also create a model repo, not just the model card, second, it technically can also be used to create/edit any kind of model card. Also, I think having "skops" appear prominently would be good.

adrinjalali:
@BenjaminBossan should we move this forward?

BenjaminBossan:
> should we move this forward?

Yes, what is the next step?

adrinjalali:
You have some items here: https://github.com/skops-dev/skops/pull/307#issuecomment-1453651097

- I think we can document this later
- maybe call it skops README generator?
- do the modifications based on the now-merged PR

BenjaminBossan:
Hmm, I get `AttributeError: 'Card' object has no attribute 'get_toc'` on the [deployed space](https://huggingface.co/spaces/skops-ci/skops-model-card-creator-b98ff038-9aba-4fc2-aac9-edbb6ba6325f), not sure how that's possible...

adrinjalali:
> Hmm, I get AttributeError: 'Card' object has no attribute 'get_toc' on the [deployed space](https://huggingface.co/spaces/skops-ci/skops-model-card-creator-b98ff038-9aba-4fc2-aac9-edbb6ba6325f), not sure how that's possible...

Would a _factory reset_ on the space help? but this is odd since they're unique spaces anyway. Hmm...

BenjaminBossan:
> Would a _factory reset_ on the space help? but this is odd since they're unique spaces anyway. Hmm...

Exactly. I reran the whole workflow, the [new space](https://huggingface.co/spaces/skops-ci/skops-model-card-creator-d8479217-2499-4bf3-8c71-ff9b97d81441) has the same issue.

adrinjalali:
Something in the manifest/setup must be flaky, we're seeing too many of these issues.

adrinjalali:
Should we merge this, have it deployed, and then try to fix it?

BenjaminBossan:
> Should we merge this, have it deployed, and then try to fix it?

We could try that and just don't advertise this feature until it's fixed.

adrinjalali (approved):
Let's merge and iterate then :)

PR #311, opened 2023-03-02
DOC add roadmap

adrinjalali:
@BenjaminBossan and I had a meeting and talked about the project's roadmap, and here's the result, and we thought it'd be nice to have it as a part of our recorded documentation, and update it every few months.

We also talked about how we can increase the visibility and usage of the project, which I don't think belongs to the project roadmap section, but putting it here to also talk about it:

The status at the moment:
- pypi downloads have been pretty stable around 50-100 downloads a day.
- We've been having continuous engagement from community members on the repo.
- The scikit-learn core team is excited about the project, and sends people to us for certain things like persistence.

Ideas on increasing usage:
- outreach: we're giving talks around
    - PyCon Italy is the next scheduled one
    - Did one in PyData Berlin meetup
    - Did another talk with Linux Foundation's ML Security Committee
- usage/conversion:
    - the model card app would make people happy probably
    - we need clearer communication on why people should use skops format, people are still confused as what the difference between `.skops` and `.onnx` is, for example.
    - many people know of ONNX, convenience methods for that could bring them here.
    - people always look for how to deploy their models online, if we have something related to that, it'd bring more people here. But this is a very large topic and whole bunch of companies/teams work on it.
    - we could try to convince a place like sagemaker to support `.skops` format. (the format would need to be more mature for that)

cc @skops-dev/maintainers

### DIFF
--- a/docs/roadmap.rst
+++ b/docs/roadmap.rst
@@ -25,7 +25,7 @@ In particular:
   visualizations about different aspects of the model. This work is started
   in :pr:`307`.
 - There are issues with model cards that make it a bit tricky to work with
-  them, such as linking images. These issues need to be resolves so that
+  them, such as linking images. These issues need to be resolved so that
   new users don't encounter too many hurdles when they first start using
   the tools.
 
--- a/docs/roadmap.rst
+++ b/docs/roadmap.rst
@@ -1,0 +1,102 @@
+.. _roadmap:
+
+Project's Roadmap
+=================
+
+``skops`` is a project which deals with certain challenges related to
+operationalizing scikit-learn based models. To that end, the following areas
+are in our focus, and are already implemented or will be in worked on in the
+near future.
+
+Model Cards
+-----------
+
+We have now :ref:`tools <model_card>` to generate model cards, but there
+are still some rough edges which need to be smoothed out. Model cards have
+the potential to be adopted by upstream scikit-learn, which would be
+possible since we have almost no external dependency there. You can see the
+list of open issues on model cards `here
+<https://github.com/skops-dev/skops/issues?q=is%3Aissue+is%3Aopen+label%3A%22model+cards%22+>`__.
+In particular:
+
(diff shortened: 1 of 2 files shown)


adrinjalali:
rendered here: https://skops--311.org.readthedocs.build/en/311/roadmap.html

docs/roadmap.rst
```diff
@@ -0,0 +1,102 @@
...
+<https://github.com/skops-dev/skops/issues?q=is%3Aissue+is%3Aopen+label%3A%22model+cards%22+>`__.
+In particular:
+
+- An app to create or modify a model card: This would also include easy
+  ways for users to add inspection of the models to the card. They can
+  potentially upload the model and the data, and easily generate good
+  visualizations about different aspects of the model. This work is started
+  in :pr:`307`.
+- There are issues with model cards that make it a bit tricky to work with
+  them, such as linking images. These issues need to be resolves so that
```
BenjaminBossan, line 28:
```suggestion
  them, such as linking images. These issues need to be resolved so that
```

docs/roadmap.rst
```diff
@@ -0,0 +1,102 @@
...
+  new users don't encounter too many hurdles when they first start using
+  the tools.
+
+Persistence
+-----------
+When deploying models, persistence is a key aspect. Since :ref:`pickles are
+insecure <persistence>`, we helps users in two ways to replace pickle
+files. One is our own ``.skops`` format, and another one is through tools
+to make it easier to convert models to ONNX. There are existing tools to
```
BenjaminBossan, line 37: Should probably be "sklearn models"?

BenjaminBossan (approved):
Great write-up. Apart from the two minor comments, nothing to add from my side.

docs/roadmap.rst
adrinjalali: Ideally they'd be able to convert non-sklearn models as well, by implementing custom converters. But that's not well documented.

PR #312, opened 2023-03-07
CI add codecov token to the config

adrinjalali:
The codecov token we have is not actually used in PRs since it's a token on GH, and they're not exposed to `pull_request` triggered actions.

The recommended action seems to be adding it hard coded, which doesn't really pose any security risks: https://github.com/codecov/codecov-action/issues/837#issuecomment-1453877750

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -5,6 +5,9 @@ codecov:
   notify:
     after_n_builds: 12
     wait_for_ci: true
+  # the coverage is uploaded with or without a token. This only helps with some
+  # codecov errors. It's a recommended action from here:
+  # https://github.com/codecov/codecov-action/issues/837#issuecomment-1453877750
   token: 2b8d4d69-6de6-4e1d-840a-5ccf9d849565
 ignore:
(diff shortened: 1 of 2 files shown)


.codecov.yml
```diff
@@ -5,5 +5,6 @@ codecov:
   notify:
     after_n_builds: 12
     wait_for_ci: true
+  token: 2b8d4d69-6de6-4e1d-840a-5ccf9d849565
```
BenjaminBossan, line 8: How about a comment about why it's okay to include the token here?

BenjaminBossan (commented):
Ah I see, it's the same issue again about exposing secrets to forks. I wonder if GH could fix that without leaking the secret, but even if possible, we can't wait for it.

.codecov.yml
adrinjalali: Added comment.

adrinjalali:
I don't think it's something GH wants to "fix".

BenjaminBossan (approved):

BenjaminBossan:
> I don't think it's something GH wants to "fix".

Yeah, at this point probably not.

PR #313, opened 2023-03-07
BUG: Fix bug in tests for fairlearn + 1 more test

BenjaminBossan:
We merged #298 and #310 shortly after each other but they contained an incompatibility that broke the fairlearn tests (the code itself was fine). This PR fixes this incompatibility.

To be clear, the only change needed to fix the tests is the following:

```python
- actual_table = card.select("Metric Frame Table").content.format()
+ actual_table = card.select("Metric Frame Table").format()
```

On top, I added the `description` argument to `add_fairlearn_metric_frame`, to be consistent with all the other methods (also changed in #310), and I also added as a test for it. Since we now have 2 tests, I moved the `metric_frame` variable to a fixture.

Finally, 2 small fixes:

- Added type annotation to transpose argument
- Changed order of arguments in docstring to match order in signature

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -1236,7 +1236,8 @@ class Card:
         self,
         metric_frame,
         table_name: str = "Fairlearn MetricFrame Table",
-        transpose=True,
+        transpose: bool = True,
+        description: str | None = None,
     ) -> Self:
         """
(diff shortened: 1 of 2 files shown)


BenjaminBossan:

adrinjalali (approved):
Thanks!

PR #314, opened 2023-03-09
FIX Fix issue with setting model_diagram to False

### LINKED ISSUE #292: BUG: After setting model_diagram=False on instantiated model card, the diagram is still shown
It would be nice to be able to disable the model diagram after the model card has been instantiated by setting the attribute to `False`. However, this does not happen:

```python
from sklearn.dummy import DummyClassifier
from skops import card

clf = DummyClassifier()
c = card.Card(clf)
print(c.model_diagram)  # True

c.model_diagram = False
c.render()  # still contains the model diagram
```

It is not difficult to work around this, as a user I can simply do:

```python
c.delete("Model description/Training Procedure/Model Plot")
```

but it's inconvenient.

BenjaminBossan:
Resolves #292, supersedes #296. The discussed solution was sufficiently different from the one proposed in #296 that it was better to open a new PR than to rework that one.

The new solution is the following: Instead of dealing with the issue what should happen if the user sets `model_card.model_diagram = False` after initializing with `True`, just remove the attribute from the `Card` class. This way, there cannot be the false impression that the diagram can be removed from the card by setting the attribute to `False`.

If users really need to remove the model diagram after enabling it at first, they can just delete the corresponding section (or make it invisible).

On top of that, a feature was added that passing a str to `model_diagram` will use that str as the section name. Previously, always the default section would be used (for skops template) or no section added (for custom templates).


BenjaminBossan:
@skops-dev/maintainers Ready for review (I can't re-trigger docs build, which failed because of huggingface_hub 0.13.0).

skops/card/_model_card.py
```diff
@@ -380,8 +380,15 @@ class Card:
         ``Path``/``str`` of the model or the actual model instance that will be
         documented. If a ``Path`` or ``str`` is provided, model will be loaded.
 
-    model_diagram: bool, default=True
-        Set to True if model diagram should be plotted in the card.
+    model_diagram: bool or str, default=True
+        If using the skops template, setting this to ``True`` will add the model
+        diagram, as generated by sckit-learn, to the default section. Passing a
```
adrinjalali, line 385: I think we should mention what the name of the section is, for them to be able to select and modify it later.

skops/card/_model_card.py
```diff
@@ -487,21 +494,27 @@ def __init__(
...
+    def _populate_template(self, model_diagram: bool | str):
+        """If initialized with a template, use it to populate the card.
+
+        Parameters
+        ----------
+        model_diagram: bool or str
+            If model_diagram is truthy and str, use str as section name. If
+            model_diagram is True, use default section for default template or
+            ignore for non-default template. If model_diagram is False, don't
+            add it.
```
adrinjalali, line 515: I think this logic makes more sense (and update the code accordingly)? 

```suggestion
            If model_diagram is str, use it as section name. If
            model_diagram is True, use default section for default template or
            raise for non-default template. If model_diagram is False, don't
            add it.
```

adrinjalali (commented):
This looks much better :)

skops/card/_model_card.py
BenjaminBossan: > If model_diagram is True ... raise for non-default template

That's what I started with. The problem is that if I initialize a `Card` object with a custom template, I would immediately get the error and need to also set `model_diagram=False`. I would find this very annoying as a user, since I never asked for the model diagram. I would prefer it if I could initialize the `Card` and only get the error if I explicitly call `model_card.add_model_plot` (same behavior as now).

BenjaminBossan: I could do that, but I wonder if this isn't overly complicated for this edge case. Especially if we want to allow passing arbitrary strings to `model_diagram`, having a special case for `"auto"` is a bit odd.

adrinjalali: I'd say using `"auto"` as a section name is not wise in a model  card anyway ;)

BenjaminBossan: Done

skops/card/_model_card.py
```diff
@@ -380,8 +380,17 @@ class Card:
         ``Path``/``str`` of the model or the actual model instance that will be
         documented. If a ``Path`` or ``str`` is provided, model will be loaded.
 
-    model_diagram: bool, default=True
-        Set to True if model diagram should be plotted in the card.
+    model_diagram: bool or "auto" or str, default="auto"
+        If using the skops template, setting this to ``True`` or ``"auto"`` will
+        add the model diagram, as generated by sckit-learn, to the default
+        section. Passing a string to ``model_diagram`` will instead use that
```
adrinjalali, line 386: We should add the section name here.

```suggestion
        section, i.e. {SECTION_NAME}. Passing a string to ``model_diagram`` will instead use that
```

adrinjalali (commented):
Otherwise LGTM.

skops/card/_model_card.py
BenjaminBossan: Done

adrinjalali (approved):

PR #316, opened 2023-03-10
DOC make sklearn's estimator run in parallel in the intelex example

adrinjalali:
On my local machine, this changes the speedup from 15 to 3. sklearn is not that slow.


BenjaminBossan:
Does intelex also use all cores? Otherwise, it's not an apples to apples comparison.

Moreover, I see now that we initially missed incorrect reporting of log loss:

> with a log-loss score of 21.7%

doesn't make sense, the f-string should not use `%`. Would you mind fixing that too?

examples/plot_intelex.py
```diff
@@ -120,7 +120,7 @@
 log_loss_score = log_loss(y_test, y_proba)
 print(
     f"[stock scikit-learn] Inference took t_stock = {t_stock:.2f}s with a "
-    f"log-loss score of {log_loss_score:.1%}"
+    f"log-loss score of {log_loss_score:.1f}"
```
BenjaminBossan, line 123:
```suggestion
    f"log-loss score of {log_loss_score:.3f}"
```

Let's put 3 digits, otherwise, it's just `0.2`.

examples/plot_intelex.py
```diff
@@ -130,7 +130,7 @@
 log_loss_score = log_loss(y_test, y_proba)
 print(
     f"[sklearnex] Inference took t_opt = {t_opt:.2f}s with a log-loss score of"
-    f" {log_loss_score:.1%}"
+    f" {log_loss_score:.1f}"
```
BenjaminBossan, line 133:
```suggestion
    f" {log_loss_score:.3f}"
```

Same comment as above.

BenjaminBossan (changes requested):
Looks good, just some minor formatting thing. There is one more `:.1%` at the bottom of the doc.

BenjaminBossan (approved):
Looks good, thx.

PR #317, opened 2023-03-10
FEAT Function to visualize skops files

### LINKED ISSUE #301: Utility function to better understand schema of a skops file
This is not high priority, but more of a nice to have.

The `schema.json` generated by the skops persistence is more or less human readable. This is good, because a curious user may want to inspect it before loading a model, so that they have an idea of what they're about to load. It does, however, contain a lot of fluff that is not necessary to understand what's going on. Also, it requires a few lines of code to even read the `schema.json`. I wonder if we could provide some helper function to assist with that.

Here is an example:

```json
# contents of a schema.json I have lying around:
{
  "__class__": "StackingRegressor",
  "__module__": "sklearn.ensemble._stacking",
  "__loader__": "ObjectNode",
  "content": {
    "__class__": "dict",
    "__module__": "builtins",
    "__loader__": "DictNode",
    "content": {
      "estimators": {
        "__class__": "list",
        "__module__": "builtins",
        "__loader__": "ListNode",
        "content": [
          {
            "__class__": "tuple",
            "__module__": "builtins",
            "__loader__": "TupleNode",
            "content": [
              {
                "__class__": "str",
                "__module__": "builtins",
                "__loader__": "JsonNode",
                "content": "\"knn@5\"",
                "is_json": true,
                "__id__": 140165569561648
              },
              {
                "__class__": "Pipeline",
                "__module__": "sklearn.pipeline",
                "__loader__": "ObjectNode",
                "content": {
                ...
# 20000 more lines
```
Maybe we could turn it into:

```python
>>> schema = sio.load_simple_schema(...)
>>> print(schema)
{
  "class": "sklearn.ensemble._stacking.StackingRegressor",
  "content": {
    "class": "builtins.dict",
    "content": {
      "estimators": {
        "class": "builtins.list",
        "content": [
          {
            "class": "builtins.tuple",
            "content": [
              {
                "class": "builtins.str",
                "content": "knn@5",
              },
              {
                "class": "sklearn.pipeline.Pipeline",
                "content": ...
...
```

That is, instead of showing module and class separately, we show it in a single line. We also strip other fields like `__loader__`, `__id__`, `key_types`, etc. This results in a much more compact and readable view of the schema.

On top of that, we could think about having a more graphical view of the resulting tree, like with the [`tree command`](https://upload.wikimedia.org/wikipedia/commons/0/0d/ReactOS-0.4.13_tree_command_667x434.png) (not sure how this would work with circular references though...).

```python
>>> sio.visualize_schema(...)
sklearn.ensemble._stacking.StackingRegressor
├── estimators
│   ├── builtins.tuple
│   │   └── builtins.str "knn@5"
│   │   └── sklearn.pipeline.Pipeline
│   │   │   └── ...
...
```

Of course, this would not be a "security feature", in the sense that looking at this simplified schema would be necessary or sufficient to determine of the object is secure.

BenjaminBossan:
Resolves #301

This is not finished, just a basis for discussion.

Here is the output for a simple `MinMaxScaler(feature_range=(-555, 123))`:

```python
sklearn.preprocessing._data.MinMaxScaler
├-attrs: builtins.dict
├---feature_range: builtins.tuple
├-----content: json-type(-555)
├-----content: json-type(123)
├---copy: json-type(true)
├---clip: json-type(false)
├---_sklearn_version: json-type("1.2.0")
```

And below is the output for a bigger `Pipeline` discussed [here](https://github.com/skops-dev/skops/issues/301#issuecomment-1456413627), but with the option `show="untrusted"`, i.e. only showing untrusted types:

```python
sklearn.pipeline.Pipeline [UNSAFE]
├-attrs: builtins.dict [UNSAFE]
├---steps: builtins.list [UNSAFE]
├-----content: builtins.tuple [UNSAFE]
├-------content: sklearn.pipeline.FeatureUnion [UNSAFE]
├---------attrs: builtins.dict [UNSAFE]
├-----------transformer_list: builtins.list [UNSAFE]
├-------------content: builtins.tuple [UNSAFE]
├---------------content: sklearn.preprocessing._data.StandardScaler [UNSAFE]
├-----------------attrs: builtins.dict [UNSAFE]
├-------------------n_samples_seen_: numpy.int64 [UNSAFE]
├-------------content: builtins.tuple [UNSAFE]
├---------------content: sklearn.pipeline.Pipeline [UNSAFE]
├-----------------attrs: builtins.dict [UNSAFE]
├-------------------steps: builtins.list [UNSAFE]
├---------------------content: builtins.tuple [UNSAFE]
├-----------------------content: sklearn.preprocessing._function_transformer.FunctionTransformer [UNSAFE]
├-------------------------attrs: builtins.dict [UNSAFE]
├---------------------------func: numpy.ufunc => numpy.core._multiarray_umath.sqrt [UNSAFE]
```

I.e. all the parent nodes are considered untrusted if a child is untrusted. This is now achieved by leveraging `node.get_unsafe_set()`.

Before going too deep with this, I would like to discuss what other users and @skops-dev/maintainers think about this. Some open points that come to my mind:

1. Should we check schema of nodes or not? Especially if they are nodes whose children are not visited*.
2. Are we fine with the visual representation? Could it be prettier? Json-types are currently shown as `json-type(123)` for instance.
3. Right now, some nodes have special dispatch for visualization, e.g. for functions (or else we would only show `numpy.ufunc`, not what ufunc is actually used). I wonder if we should move this to the `Node` classes?

*Why are some children not visited? Because they can't. E.g. for ndarray, the children are `io.BytesIO`, not nodes.


BenjaminBossan:
Okay, I made an update to only mark nodes as `[UNSAFE]` if the nodes themselves are unsafe, not if they are safe but one of their children is unsafe. This is how the example above is now shown:

```
root: sklearn.pipeline.Pipeline
├-attrs: builtins.dict
├---steps: builtins.list
├-----content: builtins.tuple
├-------content: sklearn.pipeline.FeatureUnion
├---------attrs: builtins.dict
├-----------transformer_list: builtins.list
├-------------content: builtins.tuple
├---------------content: sklearn.preprocessing._data.StandardScaler
├-----------------attrs: builtins.dict
├-------------------n_samples_seen_: numpy.int64 [UNSAFE]
├-------------content: builtins.tuple
├---------------content: sklearn.pipeline.Pipeline
├-----------------attrs: builtins.dict
├-------------------steps: builtins.list
├---------------------content: builtins.tuple
├-----------------------content: sklearn.preprocessing._function_transformer.FunctionTransformer
├-------------------------attrs: builtins.dict
├---------------------------func: numpy.ufunc => numpy.core._multiarray_umath.sqrt [UNSAFE]
```

BenjaminBossan:
A variant using ansi colors:

image

skops/io/_visualize.py
```diff
@@ -0,0 +1,262 @@
...
+    if show == "all":
+        should_print = True
+    elif show == "untrusted":
+        should_print = not node_and_children_are_safe
+    else:  # only trusted
+        should_print = node_is_safe
+    return should_print
+
+
+def _check_node_and_children_safe(node: Node, trusted: bool | Sequence[str]) -> bool:
```
adrinjalali, line 75: I think it'd be okay to move this to nodes and cache them.

skops/io/_visualize.py
```diff
@@ -0,0 +1,262 @@
...
+    elif trusted is False:
+        node_and_children_are_safe = not node.get_unsafe_set()
+    else:
+        node_and_children_are_safe = not (node.get_unsafe_set() - set(trusted))
+    return node_and_children_are_safe
+
+
+# use singledispatch so that we can register specialized visualization functions
+@singledispatch
+def format_node(node: Node) -> str:
```
adrinjalali, line 90: also don't mind moving these to nodes.

adrinjalali (commented):
I think this is great, and better than what I originally had. I personally like this version the best: https://github.com/skops-dev/skops/pull/317#issuecomment-1468133355, but that's because I'm colorblind-ish and the colors in the last version bother me a bit, but it'd be nice to have that as an argument on how to format things.

skops/io/_visualize.py
BenjaminBossan: Oh yeah, there is already `node.is_safe()` >__<

BenjaminBossan:
@skops-dev/maintainers This is now ready for review

As an example, here is the top part of the visualization of the following estimator:

```python
pipeline = Pipeline([
    ("features", FeatureUnion([
        ("scaler", StandardScaler()),
        ("scaled-poly", Pipeline([
            ("polys", FeatureUnion([
                ("poly1", PolynomialFeatures()),
                ("poly2", PolynomialFeatures(degree=3, include_bias=False))
            ])),
            ("square-root", FunctionTransformer(unsafe_function)),
            ("scale", MinMaxScaler()),
        ])),
    ])),
    ("clf", LogisticRegression(random_state=0, solver="liblinear")),
]).fit([[0, 1], [2, 3], [4, 5]], [0, 1, 2])

file = sio.dumps(pipeline)
sio.visualize_tree(file)
```

image

Lots of stuff can be customized, e.g. if colors should be used (and which), how to mark untrusted types, etc. The visualization requires `rich`, but users can provide custom printing functions if they want to avoid that.

skops/io/_visualize.py
```diff
@@ -0,0 +1,326 @@
...
+
+    # RECURSE
+    yield from walk_tree(
+        node.children,
+        node_name=node_name,
+        level=level + 1,
+    )
+
+
+def visualize_tree(
```
adrinjalali, line 255: I don't think it matters that it's a _tree_. What about just `visualize`? Or `inspect` (although there's the `inspect` module)?

adrinjalali (commented):
This looks pretty nice. We do need to add `rich` to `setup.py` with a key though, somehow in extras. I would have liked to keep your own implementation of `sink`, so that if `rich` is not installed, or a parameter is passed in a certain way, we just use that.

BenjaminBossan:
> I would have liked to keep your own implementation of sink, so that if rich is not installed, or a parameter is passed in a certain way, we just use that.

@adrinjalali I experimented a bit and could create a custom implementation that does not rely on `rich.Tree`. It produces the exact same output, so I'd switch over to it. However, I would not go down the road of reimplementing the colorization stuff, since that can become quite tricky. Therefore, I would propose that we don't use colors if rich is not installed and do use colors if it is installed. Of course, users can still pass `use_colors=False` to prevent colorization even if rich is installed. I think in that case, we don't need an optional install flag for rich, as there would be no error if rich is not installed. WDYT?

adrinjalali:
> It produces the exact same output, so I'd switch over to it.

Perfect. 

> I think in that case, we don't need an optional install flag for rich, as there would be no error if rich is not installed. WDYT?

That's fair, but I'd still add `rich` to an `extras` group kinda thing in `setup.py`, so that users can easily have a way of installing `skops` with all optional dependencies.

skops/io/_visualize.py
```diff
@@ -0,0 +1,326 @@
...
+from ..utils.importutils import import_or_raise
+from ._audit import Node, get_tree
+from ._general import FunctionNode, JsonNode
+from ._numpy import NdArrayNode
+from ._scipy import SparseMatrixNode
+from ._utils import LoadContext
+
+
+@dataclass
+class NodeInfo:
```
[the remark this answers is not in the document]
BenjaminBossan: added

BenjaminBossan (commented):
I addressed your points, please review again.

adrinjalali (approved):
This is awesome!

PR #319, opened 2023-03-14
Refactor model card init

BenjaminBossan:
_This PR is just a basis for discussion_

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -443,6 +443,8 @@ class Card:
     >>> X, y = load_iris(return_X_y=True)
     >>> model = LogisticRegression(solver="liblinear", random_state=0).fit(X, y)
     >>> model_card = Card(model)
+    >>> model_card.populate_template()
+    Card(...)
     >>> model_card.metadata.license = "mit"
     >>> y_pred = model.predict(X)
     >>> model_card.add_metrics(**{
@@ -491,7 +493,6 @@ class Card:
     def __init__(
         self,
         model,
-        model_diagram: bool | Literal["auto"] | str = "auto",
         metadata: ModelCardData | None = None,
         template: Literal["skops"] | dict[str, str] | None = "skops",
         trusted: bool = False,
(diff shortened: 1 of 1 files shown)


skops/card/_model_card.py
```diff
@@ -504,9 +505,10 @@ def __init__(
         self._data: dict[str, Section] = {}
         self._metrics: dict[str, str | float | int] = {}
 
-        self._populate_template(model_diagram=model_diagram)
-
-    def _populate_template(self, model_diagram: bool | Literal["auto"] | str):
+    def populate_template(
```
adrinjalali, line 508: users shouldn't have to call this one, do they?

adrinjalali (commented):
I think we can remove the `        if self.template == Templates.skops.value:` from this method.

Also, we can remove `            if self.template == Templates.skops.value:` like statements from the other methods.

skops/card/_model_card.py
BenjaminBossan: If they didn't, then the card would always be empty, would it not?

BenjaminBossan (commented):
> I think we can remove the `if self.template == Templates.skops.value`: from this method.

You mean that the code below that should be called unconditionally? This would require there to be defaults for the section names in case a non-default template is being used.

> Also, we can remove `if self.template == Templates.skops.value:` like statements from the other methods.

If we set default section names, we could do that. However, we could do that regardless of the change in this PR.

skops/card/_model_card.py
adrinjalali: when I create a card based on a template, I expect at least the default sections to be there.

adrinjalali:
> This would require there to be defaults for the section names in case a non-default template is being used.

Yes.

skops/card/_model_card.py
BenjaminBossan: Okay, but then it's not different from what we already have. I think this means that we could add the default section names to the `add_` methods but that's it. This PR could be closed.

adrinjalali: looks like it, and simplify their code.

BenjaminBossan:
Conclusion: Not a worthy change, but we can set default section names for the `add_` methods, which would allow to get rid of the template checks inside of them, resulting in simplified code.

PR #321, opened 2023-03-16
ENH: Default section for Card.add_* methods

BenjaminBossan:
## Description

The `add_*` methods for `Card` now have default sections (the one used in the skops template) and no longer set descriptions by default. `None` is no longer a valid argument for section.

This is useful because it allows us to completely remove all the template checks in the diverse `add_*` methods. Before this change, many `add_*` methods had to check the template and if it was a custom template, would raise an error when the section is not indicated.

All methods have been homogenized in that they no longer add a description by default and put a placeholder there if empty. This means that after this PR is merged, the same code will sometimes produce a different looking model card.

## Comment

For the `Card` class, this is a nice simplification of the code. The main change affects the unit tests. All tests that used to raise when using a custom template without a section argument no longer raise but use the default template. Furthermore, since some default descriptions have been removed, other tests had to be touched too. Apart from that, some minor cleanups in the tests.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -22,6 +22,9 @@ v0.6
   `Benjamin Bossan`_.
 - Fix: skops persistence now also works with many functions from the
   :mod:`operator` module. :pr:`287` by `Benjamin Bossan`_.
+- ``add_*`` methods on :class:`.Card` now have default section names (but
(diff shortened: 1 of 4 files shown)


BenjaminBossan:

adrinjalali (approved):
Love this, merging the latest `main` to double check before merging.

PR #322, opened 2023-03-16
ENH Dealing with skops persistence protocol updates

BenjaminBossan:
~This PR is a POC that is intended as a basis for discussing how we can deal with updating the skops persistence protocol.~

This is a working implementation that shows one way to deal with updates in the skops persistence protocol in a backwards compatible manner that should, hopefully, scale.

As an example, the update to how functions are persisted is rolled into this PR to show that it works. Therefore, this PR supersedes #320.

## Problem description

The problem in general occurs if we make a change to the state that is stored with an object. For instance, we could store an additional field. When we try to load old state without that key, there would be an error. Therefore, we have to somehow adjust our loading code to deal with this possibility.

## Rejected solution

We want to avoid having a bunch of conditionals in our loading code that checks what protocol is used and then does this or that thing. This approach would become messy and error prone quite quickly, because we intend to support old protocols for a very long time.

## Proposed solution

This proposal here would allow to write a clean implementation of the new deserialization code without regard for backwards compatibility. The old code, however, would not be deleted but instead moved to a different module and registered with its old protocol number. Then, while loading, we would check the protocol stored in the schema and use the old code if there is a match for it. If there is no match, we assume we can safely use the current protocol instead. For this to work, the key used for mapping the state to the loader is changed to be the tuple `(loader, protocol)`, instead of just the `loader`, as has been the case so far. This is a hopefully transparent way to "dispatch" on both the loader and the protocol.

This is just a draft and I haven't tested it on a real example. If we agree that this is the way forward, I would expand on it and test it properly. But first we should agree that this is the way to go and check if there are no problems with this approach.

## Description of the mechanism for changing the protocol

Below are the instructions with how to deal with a change in protocol:

Every time that a backwards incompatible change to the skops format is made for the first time within a release, the protocol should be bumped to the next higher number. The old version of the Node, which knows how to deal with the old state, should be preserved and registered. Let's give an example:

- There is a BC breaking change in `FunctionNode`.
- Since it's the first BC breaking change in the skops format in this release, bump `skops.io._protocol.PROTOCOL` from version X to Y.
- Move the old `FunctionNode` code into `skops/io/old/_general_vX.py`, where 'X' is the old protocol.
- Register the `_general_vX.FunctionNode` in `NODE_TYPE_MAPPING` inside of `_persist.py` (on top of registering the current version).
- Write a test in `test_persist_old.py` that shows that the old state can still be loaded.

Now, if a user loads a `FunctionNode` state with version X using skops with version Y, the old code will be used instead of the new one. For all other node types, if there is no loader for version X, skops will automatically use version Y instead.

@skops-dev/maintainers RFC


adrinjalali (commented):
Other than adding some docs in the `_general_v0.py` and some tests to check for old protocol, I think this looks perfect.

BenjaminBossan:
> Other than adding some docs in the `_general_v0.py` and some tests to check for old protocol, I think this looks perfect.

One way to go about that would be that I wrap #320 into this PR and have the old + new `FunctionNode` code as a test case. Otherwise, I would have to come up with a fake example. WDYT?

adrinjalali:
adding #320 here sounds good to me.

BenjaminBossan:
@skops-dev/maintainers Despite the best efforts of codecov and rtd, I managed to get CI green. Ready for review.

skops/hub_utils/_hf_hub.py
```diff
@@ -264,6 +264,7 @@ def _create_config(
         does not support it. For more info, see
         https://intel.github.io/scikit-learn-intelex/.
     """
+
```
adrinjalali, line 267: why is this file changed here? 😁

skops/io/tests/_utils.py
```diff
@@ -173,3 +178,62 @@ def assert_method_outputs_equal(estimator, loaded, X):
             X_out1 = getattr(estimator, method)(X)
             X_out2 = getattr(loaded, method)(X)
             assert_allclose_dense_sparse(X_out1, X_out2, err_msg=err_msg, atol=ATOL)
+
+
+def downgrade_state(*, data: bytes, keys: list[str], old_state: dict, protocol: int):
```
adrinjalali, line 183: this is a bit cryptic, an example in the docstring on how it's supposed to be used would be nice.

skops/io/old/_general_v0.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Any, Sequence
+
+from skops.io._audit import Node
+from skops.io._trusted_types import SCIPY_UFUNC_TYPE_NAMES
+from skops.io._utils import LoadContext, _import_obj
+
+PROTOCOL = 0
+
+
+class FunctionNode(Node):
```
adrinjalali, line 12: we should actually this this one, shouldn't we? we're not at the moment.

adrinjalali (approved):

adrinjalali:
I'm feeling much better about our persistence now :)

PR #326, opened 2023-03-23
FIX: Workflow file for persistence performance

BenjaminBossan:
- wrong name
- trying to fix install

### DIFF
--- a/.github/workflows/persistence-performance.yml
+++ b/.github/workflows/persistence-performance.yml
@@ -15,9 +15,9 @@ jobs:
     timeout-minutes: 10
 
     steps:
-    - uses: actions/checkout@v3
+    - uses: actions/checkout@v3.5
     - name: Set up Python
-      uses: actions/setup-python@v4
+      uses: actions/setup-python@v4.5
     - name: Install requirements
       run: |
         pip install .[tests]
--- a/.github/workflows/persistence-performance.yml
+++ b/.github/workflows/persistence-performance.yml
@@ -17,7 +17,7 @@ jobs:
     steps:
     - uses: actions/checkout@v3
     - name: Set up Python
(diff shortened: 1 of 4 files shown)


BenjaminBossan:

adrinjalali:
Could you please set this to trigger on PR so that we can test it here and see if the install actually fixes the issue?

BenjaminBossan:
(don't merge yet)

@adrinjalali While the fundamental issue still exists, I found a "solution" by copying the script to the project root inside the GH action. WDYT?

adrinjalali:
Since this is to "fix" import issues, would it work if we do an `os.chdir` or something inside the script instead, to run it from the root, before running the import statements? I'm not sure which one's better.

But I'm happy to merge and revert once https://github.com/actions/setup-python/issues/634 is resolved maybe?

.github/workflows/persistence-performance.yml
```diff
@@ -17,10 +18,13 @@ jobs:
...
+        python-version: "3.11"
     - name: Install requirements
       run: |
         pip install .[tests]
         pip --version
         pip list
     - name: Run persistence performance checks
-      run: python scripts/check_persistence_performance.py
+      run: |
+        python3.11 scripts/check_persistence_performance.py
```
adrinjalali, line 30: this probably can be `python3` or even `python`, right? You can probably revert these two lines to what it was before actually.

BenjaminBossan:
@adrinjalali This works now, here is an output: https://github.com/skops-dev/skops/actions/runs/4531268169/jobs/7981214101?pr=326

I have removed the PR hook again, so we'll be 100% certain in 1 week :)

adrinjalali (approved):

PR #327, opened 2023-03-24
Test: debugging gh action

BenjaminBossan:
Testing, *do not merge*

 Run two tasks that are identical, except that one uses `actions/setup-python@v3` and the other uses `actions/setup-python@v4`.


adrinjalali:
I guess we can close this one, we know the issue.

PR #330, opened 2023-03-28
ENH Copy plots on Card save

### LINKED ISSUE #300: Annoyance with adding figures to model cards
There is a small annoyance I found when adding figures to model cards. The problem is that the path to the file on the storage device and the path saved in the model card can differ. This can easily result in model cards references figures that it cannot find. Let me give an example:

I want to create a model card to upload it later to the HF Hub. For this, I create a temp folder and initialize a repo:

```python
hub_dir = Path(mkdtemp())
hub_utils.init(..., dst=hub_dir)
```

Then I create my model card and add a figure like so:

```python
model_card = card.Card(...)
fig_path = hub_dir / "my_plot.png"
fig.savefig(fig_path)
model_card.add_plot(**{"My beautiful plot": fig_path})
...
hub_utils.push(..., source=hub_dir)
```

The issue is that the plot is stored as `My beautiful plot`, i.e. using its absolute path. When I now upload the model card to the Hub, the Hub will search for the plot in said temporary directory, which does not exist. On the Hub, what I actually want is to use the relative path, i.e., `My beautiful plot`. But I cannot use the relative path instead of the absolute path either, because then the figure is not uploaded when calling, `hub_utils.push(..., source=hub_dir)`, since it's not in `hub_dir`, it is in my current working directory.

Now I can work around this by changing my code to

```python
model_card = card.Card(...)
fig.savefig(hub_dir / "my_plot.png")  # <= use abs path here
model_card.add_plot(**{"My beautiful plot": "my_plot.png"})  # <= use rel path here
```

but this is an ugly gotcha. Also, even though it will work correctly on the Hub, when I render the model card locally to preview it, it will fail because the paths mismatch.

Even worse, however, is when I want to add permutation importances. When I do

```python
pi = permutation_importance(model, X, y)
model_card.add_permutation_importances(pi, plot_file="my-permutation-importances.png", ...)
```

there will be a problem because the method stores the file locally using the `plot_file` argument, and it will use the same argument in the model card. If I put the relative path to `hub_dir` there, the file will not be found when the model card is rendered on the Hub. And when I put the relative path there, the file will not be included in the upload. The only solution is not to use `add_permutation_importances` but to create the figure by hand and call `add_plot` as discussed above, or I have to manually create a copy the file at the correct location.

I guess the problem could be avoided if the current working directory and the temporary directory for pushing are always identical, but requiring this is also annoying and I don't think we can assume it's always needed.

My guess why we haven't really encountered this problem yet is that in our examples, we either create model cards with plots but don't upload them to the Hub, or we do upload model cards to the Hub but they don't contain plots. We don't have an example that combines both (I'm working on exactly that, which is how I stumbled upon this).

Anyway, I don't have a solution to this problem. We can better document it and maybe warn users if a repo is created with a model card that references plots that are not inside the repo. But that's not very satisfying. Anyone got better ideas?

lazarust:
#### Reference Issues/PRs

Fixes #300 

Implements copying files to the path of the repository.


lazarust:
@BenjaminBossan @adrinjalali Here's the draft PR, there's not much because I wanted to make sure I was thinking of this correctly. 

From what I understood from the thread on the issue, we're wanting to copy the files to the repo path whenever `add_plot` method is called. Is that understanding correct?

Also, I've been trying to think of how best to test this. Do y'all have any ideas?

Sorry, this PR is so small. I just wanted to make sure I was heading in the right direction before jumping into the deep end.

skops/card/_model_card.py
```diff
@@ -1057,6 +1064,9 @@ def add_plot(
                 folded=folded,
             )
             self._add_single(section_name, section)
+
+            if path_in_repo is not None:
+                shutil.copy(plot_path, path_in_repo)
```
adrinjalali, line 1069: this should be in the above section, and modify the `plot_path` variable accordingly.

adrinjalali (commented):
The copying should probably happen when the user calls `save` on a `Card` object. And instead of `path_in_repo`, we should probably call it sth like `relative_path` or something, and then copy all plots when `save` is called, to a path relative to where the path to `save` is given. WDYT?

skops/card/_model_card.py
adrinjalali: yes 👍🏼

lazarust: @adrinjalali Hmm, after thinking through this some I'm unsure if copying on model card save is the best. If we do it when a user calls `add_plot()` then we'd already have the path of the plot in the context. From what I understand of how `.save()` is working we'd have to loop through all the generated content, find the plots, and then save them.

That feels a little extra to me since we could just save the copy when a user adds the plot. What are your thoughts? Am I misunderstanding how `.save()` is working?

adrinjalali: You're right that `save` would do that in an iteration. But from the user's experience, to me it makes more sense. This way, users only provide a relative path to the destination of the model card, and at the end we save them all. It should also provide a more consistent experience since they can't mess up addresses and at the end not have them in the repo.

adrinjalali:
I'd be interested in what @BenjaminBossan thinks here too though.

skops/card/_model_card.py
BenjaminBossan: I agree with Adrin that copying on `save` has some advantages that make it worth the extra effort. On top of what was said, it would help to avoid the problem of a user adding a plot, then overwriting the plot file, but that change not being reflected in the final model card.

Implementation-wise, I agree with Thomas that looping through all sections _again_ is a bit wasteful. However, I think this feature can be implemented without a second loop. For that, I would add an argument to `_generate_card` to optionally copy the figures. This argument is passed down to `_generate_content`, which can take care of the copying. This way, we don't iterate through all sections twice and avoid duplicating the logic (which can be problematic besides performance, e.g. the current implementation does not consider invisible sections).

Another implementation detail, instead of `if hasattr(section, "path")`, I would have probably done `if isinstance(section, PlotSection)`, but I'm open to be convinced otherwise :)

lazarust:
@BenjaminBossan I've moved the logic into `_generate_content` and this is ready to review! Thanks! 👍🏽

skops/card/tests/test_card.py
```diff
@@ -1887,3 +1887,20 @@ def test_toc_with_invisible_section(self, card):
...
         assert toc == "\n".join(exptected_toc)
+
+
+class TestCardSaveWithPlots:
+    def test_copy_plots(self, destination_path, model_card):
+        import matplotlib.pyplot as plt
+
+        plt.plot([4, 5, 6, 7])
+        plt.savefig(Path(destination_path) / "fig1.png")
+        model_card = model_card.add_plot(fig1="fig1.png")
```
BenjaminBossan, line 1898: The plot path here is already incorrect, is it not? Because you save the figure at `Path(destination_path) / "fig1.png"` but reference `"fig1.png"`.
(btw we may want to verify in a future PR that a plot exists when calling `add_plot`)

skops/card/_model_card.py
```diff
@@ -1300,7 +1301,7 @@ def _generate_metadata(self, metadata: ModelCardData) -> Iterator[str]:
             yield aRepr.repr(f"metadata.{key}={val},").strip('"').strip("'")
 
     def _generate_content(
-        self, data: dict[str, Section], depth: int = 1
+        self, data: dict[str, Section], depth: int = 1, parent_path=None, plot_path=None
```
BenjaminBossan, line 1304: Could you please add the type annotation?

skops/card/_model_card.py
```diff
@@ -1387,19 +1400,21 @@ def __repr__(self) -> str:
...
+            self._data, parent_path=parent_path, plot_path=plot_path
+        ):
             if line:
                 yield "\n" + line
 
         # add an empty line add the end
         yield ""
 
-    def save(self, path: str | Path) -> None:
+    def save(self, path: str | Path, plot_path=None) -> None:
```
BenjaminBossan, line 1417: Could you please add the type annotation?

BenjaminBossan (changes requested):
Thanks for doing the adjustments.

I wonder if we can simplify the implementation a bit. Right now, the code considers the possibility that `plot_path != destination_path` but I wonder if that's relevant. If I store my model card into `foo/bar/`, would I ever want my plots to be copied to `spam/eggs/`? I can't think of a use case, but maybe I miss something?

If this is indeed not required, the implementation could be simplified by only having one new argument to `_generate_content`, which is the destination path. The path of the figure is already known from `plot_section.path`, so everything required for copying would be there.

From the user point of view, I'd also be unsure what the point of the `plot_path` argument is in `Card.save`, I think it needs a bit more explanation. Personally, I would also change the argument to be something like `copy_files: bool`. If `True`, it copies all the required files (which, at this point, are only the plots, but who knows what the future brings) to the same directory as the `README.md`. WDYT?

PS: It would be nice to extend the tests so that we have 2 plots, one with a relative path (as is right now) and one with an absolute path, to ensure that both work.

adrinjalali:
> I can't think of a use case, but maybe I miss something?

+1

skops/card/tests/test_card.py
lazarust: I was basing this off of this `TestAddPlot` test https://github.com/skops-dev/skops/blob/3ab308bd9beab44e816272790aaa64e04d3f6910/skops/card/tests/test_card.py#L1101

lazarust:
@BenjaminBossan Thanks for the feedback! I refactored this to copy the files into the same path as the README. I also went ahead and added a second plot to the test. 

(Side note it looks like there are `34 warnings` when running pytest. Is that something that needs fixing? If so I can start working on a PR)

skops/card/tests/test_card.py
BenjaminBossan: So the point of this PR is what was described in #300. A passage from there that fits this situation is:

> Now I can work around this by changing my code to

```python
model_card = card.Card(...)
fig.savefig(hub_dir / "my_plot.png")  # <= use abs path here
model_card.add_plot(**{"My beautiful plot": "my_plot.png"})  # <= use rel path here
```

> but this is an ugly gotcha. Also, even though it will work correctly on the Hub, when I render the model card locally to preview it, it will fail because the paths mismatch.

So the test you're referring to does exactly that, using a mix of relative and absolute path, which is exactly what we want to avoid with the new feature. That's why I think it's important for this test to not use that pattern. We could adjust the other tests as well, but that's really not so important there, since what they're testing is not concerned with the paths. I hope that clears it up.

skops/card/tests/test_card.py
```diff
@@ -1887,3 +1887,22 @@ def test_toc_with_invisible_section(self, card):
...
+
+        plt.plot([7, 6, 5, 4])
+        plt.savefig(Path(destination_path) / "fig2.png")
+        model_card = model_card.add_plot(fig2="fig2.png")
+
+        model_card.save(Path(destination_path) / "README.md")
```
BenjaminBossan, line 1904: So here, you're not using the new feature here, right? And the test still passes. This is because you saved the figures in the correct destination right from the start. But that's what we want the new feature to do.

skops/card/tests/test_card.py
```diff
@@ -1887,3 +1887,22 @@ def test_toc_with_invisible_section(self, card):
...
+    def test_copy_plots(self, destination_path, model_card):
+        import matplotlib.pyplot as plt
+
+        plt.plot([4, 5, 6, 7])
+        plt.savefig(Path(destination_path) / "fig1.png")
+        model_card = model_card.add_plot(fig1="fig1.png")
+
+        plt.plot([7, 6, 5, 4])
+        plt.savefig(Path(destination_path) / "fig2.png")
+        model_card = model_card.add_plot(fig2="fig2.png")
```
BenjaminBossan, line 1902: The second plot added here does not really test anything that the first doesn't test. You probably added it because I wrote:

> It would be nice to extend the tests so that we have 2 plots, one with a relative path (as is right now) and one with an absolute path, to ensure that both work.

What I mean is that this second plot should add the plot using its absolute path but this still uses the relative path.

skops/card/_model_card.py
```diff
@@ -1300,7 +1301,11 @@ def _generate_metadata(self, metadata: ModelCardData) -> Iterator[str]:
             yield aRepr.repr(f"metadata.{key}={val},").strip('"').strip("'")
 
     def _generate_content(
-        self, data: dict[str, Section], depth: int = 1
+        self,
+        data: dict[str, Section],
+        depth: int = 1,
+        path: str | Path = "",
+        copy_files: bool = False,
```
BenjaminBossan, line 1308: Personally, I'd find the implementation cleaner if we only used one argument: I would make `path` optional and if it is `None`, it means don't copy, if it is not `None`, copy to said path. There is no scenario where `path` is not `None` but `copy_files` is `True`, right?

Perhaps we can also come up with a better name than `path`, not exactly sure what, perhaps `destination_path` or something along those lines. And the docstring below should also be extended to explain the new argument (doesn't have to be a full list of parameters, just one sentence).

BenjaminBossan (approved):
Thanks for implementing the suggested changes. I think there is still some work to do on the tests and the implementation could be simplified. Please take a look.

BenjaminBossan:
Sorry, I accidentally submitted "Approve" when I wanted to submit "Request changes", I don't know how to change it now -__-

skops/card/tests/test_card.py
lazarust: Yeah 🤦🏽 Sorry I missed that

BenjaminBossan:
@lazarust Thanks for working on the test, but it looks like something is broken now. Could you please check?

lazarust:
@BenjaminBossan Should be working now!

BenjaminBossan:
Thanks for fixing the test. The one failing CI job is unrelated, so please ignore it. 

From the comments of my last review, not everything was addressed, right? Will you do that?

lazarust:
@BenjaminBossan Sorry about that, I've removed the `copy_files` parameter from everywhere other than `Card.save()`

skops/card/_model_card.py
```diff
@@ -1410,13 +1425,18 @@ def save(self, path: str | Path) -> None:
         path: str, or Path
             Filepath to save your card.
 
+        plot_path: str
+            Filepath to save the plots.
```
BenjaminBossan, line 1429: Could you please elaborate on this? E.g. mention under what circumstances users may want to use this parameter.

skops/card/_model_card.py
```diff
@@ -1410,13 +1425,18 @@ def save(self, path: str | Path) -> None:
...
+
         Notes
         -----
         The keys in model card metadata can be seen `here
         <https://huggingface.co/docs/hub/models-cards#model-card-metadata>`__.
         """
         with open(path, "w", encoding="utf-8") as f:
-            f.write("\n".join(self._generate_card()))
+            if not isinstance(path, Path):
```
BenjaminBossan, line 1437: There is a bit of redundancy here: You cast `path` to a `Path` object here but `_generate_card` and `_generate_content` both accept `str` and the latter does `Path(destination_path)`. So either you just pass `path` without conversion, or those methods should not accept `str` and don't perform a 2nd round of conversion.

skops/card/_model_card.py
```diff
@@ -1410,13 +1425,18 @@ def save(self, path: str | Path) -> None:
...
+                path = Path(path)
+            f.write("\n".join(self._generate_card(path.parent if copy_files else None)))
```
BenjaminBossan, line 1439: Just a nit, but this is more readable IMO:

```suggestion
            destination_path = path.parent if copy_files else None
            f.write("\n".join(self._generate_card(destination_path=destination_path)))
```

BenjaminBossan (changes requested):
Great, thanks for working in my suggestion and merging main to fix the CI issue. I found a few very minor things to improve, but after this it's good to merge.

lazarust:
@BenjaminBossan Sorry this took so long for me to get back to! This should be ready for you to take another look!

BenjaminBossan (approved):
Thanks a lot, we're always happy to have your contributions.

PR #331, opened 2023-03-28
FIX: Bug in random generator persistence

### LINKED ISSUE #329: Bug: numpy default rng persistence
When trying to determine the unsafe set of a numpy random `Generator`, there is an error. To reproduce:

```python
def test_default_rng():
    import skops.io as sio
    rng = np.random.default_rng()
    sio.loads(sio.dumps(rng), trusted=[])
```

We test this class but only with `trusted=True`, which is why we didn't encounter the error in our test suite.

I haven't dug into the source of the problem, but possibly it's related to `Generator` being a [cython class](https://github.com/bashtage/randomgen/blob/6e20bd0a3b4022f6406e70b31bfb4f2ca87ef211/randomgen/pcg64.pyx#L44).

Maybe we should tighten our tests by relying as little as possible on `trusted=True`.

BenjaminBossan:
Fixes #329 

One of its children was not persisted correctly (pure dict instead of `DictNode`), which made it impossible to check the untrusted types on it. This fix correctly stores the state as a node.

Note that the old implementation does correctly load the `Generator` instance, but it doesn't work with anything but `trusted=True`.

Tests were extended to uncover this bug.

The old persistence scheme was preserved for protocol vesion 0 -- even though it is buggy, it can work under some circumstances.


BenjaminBossan:

adrinjalali (approved):

PR #332, opened 2023-03-28
DOC Add section on compatibility of skops format

BenjaminBossan:
Added paragraphs:

1. Promise to not break backwards compatibility
2. Warning about using skops main branch for persistence

NB: I made a short note that users may want to do a load-dump cycle of skops files to upgrade them to the latest format. I wonder if we should extend the skops CLI to make this work from command line. WDYT?

### DIFF
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -222,7 +222,7 @@ Backwards compatibility
 -----------------------
 
 Compatibility across skops versions
-===================================
+###################################
 
 The skops persistence format is in flux, as we steadily work on improving it,
 making it more secure and supporting more types. When we make a change that is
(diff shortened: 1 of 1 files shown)


docs/persistence.rst
```diff
@@ -218,8 +218,32 @@ Therefore, you should only load a skops file containing a model of any of those
...
+versions still load as always, even if they contained a bug (in which case we
+will warn about it). Therefore, it is generally safe to assume that your skops
+files will keep on working with future versions of skops.
+
+You may want to periodically load and dump old skops files using newer versions
+of skops to benefit from the updates to the protocol.
+
+One caveat to the backwards compatibility promise is that the skops files have
+to be created based on a release version of skops. If you create skops files
+using a skops version installed from the main branch, it is possible to end up
```
adrinjalali, line 241:
```suggestion
using a skops version installed from the development branch, it is possible to end up
```

or put main in backticks to make it code like. Otherwise the sentence might sound odd to the reader.

adrinjalali (approved):

PR #334, opened 2023-03-28
RFC: Store more objects as pure json

BenjaminBossan:
## Description

Right now, when using skops persistence, only some primitive types are stored as json: str, numbers, and null. However, in theory we should also be able to store *some* lists and dicts. The advantage of that is that the storage is much more compact, helping with disk size and readability.

One issue is that we need to check if the lists and dicts can indeed be stored as pure json objects. To check this, we make a json roundtrip and verify that the loaded objects still equals the original object. This could of course impact performance.

On top of that, we also need to explicitly check the types. This is because a numpy float will `==` a Python float of the same value, so the json roundtrip might suggest that a dict with numpy floats was restored successfully. But that's not true, we get a Python float back, which is not the same. Checking the types explicitly solves the issue.

The presence of numpy types is, however, also a reason why this feature helps less than I intially thought: Many sklearn objects will store numpy objects as attributes, so their `__dict__` cannot be stored as a pure json object. This is from anecdotal observation, I haven't run the the numbers.

AFAICT, this change should be backwards compatibly, since the old, non "compact" list and dict nodes are still loaded as they were previously.

## Additional

Somewhat unrelatedly, I "fixed" a "bug" where the `constructor` of a `ReduceNode` was stored directly in the node's `children`. This made it so that the object could not be visualized, because the visualization code assumes that all children are nodes. Not sure if we want to make that assumption or not.

## Results

To illustrate how much more compact some objects can be after the change, here is the `state` of `DecisionTreeClassifier`'s `__dict__` with the proposed change:

```python
{'__class__': 'dict',
 '__loader__': 'DictNode',
 '__module__': 'builtins',
 '_is_json': True,
 'content': '{"criterion": "gini", "splitter": "best", "max_depth": 1, '
            '"min_samples_split": 2, "min_samples_leaf": 1, '
            '"min_weight_fraction_leaf": 0.0, "max_features": null, '
            '"max_leaf_nodes": null, "random_state": null, '
            '"min_impurity_decrease": 0.0, "class_weight": null, "ccp_alpha": '
            '0.0, "_sklearn_version": "1.2.2"}'}
```

And this is what it currently looks like:

```python
{'__class__': 'dict',
 '__loader__': 'DictNode',
 '__module__': 'builtins',
 'content': {'_sklearn_version': {'__class__': 'str',
                                  '__id__': 140710339249776,
                                  '__loader__': 'JsonNode',
                                  '__module__': 'builtins',
                                  'content': '"1.2.2"',
[...]

PR #335, opened 2023-03-29
RFC: Persist more objs as pure json

BenjaminBossan:
Alternative to #334 

@skops-dev/maintainers RFC

The two implementations result in the same outcome but are implemented differently: This one checks during `get_state` if we can use pure json and if so, returns a state that will be loaded with `JsonNode`. The other implementation will not dispatch to other `get_state` functions, instead the different nodes (`JsonNode`, `DictNode`, `ListNode`) will take care of loading the json if applicable.

I think this implementation is cleaner than the other one. A disadvantage of this implementation is that there needed to be a change in how numpy object dtype arrays are persisted, which will require a protocol change.

RFC, especially regarding:

1. Do we want to go down any of these paths?
2. If so, is this implementation better or #334?
3. Anything that could be improved?

## Description

Right now, when using skops persistence, only some primitive types are stored as json: str, numbers, and null. However, in theory we should also be able to store some lists and dicts. The advantage of that is that the storage is much more compact, helping with disk size and readability.

One issue is that we need to check if the lists and dicts can indeed be stored as pure json objects. To check this, we make a json roundtrip and verify that the loaded objects still equals the original object.

On top of that, we also need to explicitly check the types. This is because a numpy float will == a Python float of the same value, so the json roundtrip might suggest that a dict with numpy floats was restored successfully. But that's not true, we get a Python float back, which is not the same. Checking the types explicitly solves the issue.

The presence of numpy types is, however, also a reason why this feature helps less than I intially thought: Many sklearn objects will store numpy objects as attributes, so their __dict__ cannot be stored as a pure json object.

### Disk size

I stored the dumped objects on disk, once with main and once with this refactor:

```
$ du .
1000	./refactor-unfitted
1564	./normal-unfitted
22856	./normal-fitted
22664	./refactor-fitted
```

As can be seen, for unfitted objects, we get 1/3 reduction in size, but for fitted objects, there is no significant difference. This confirms the suspicion that for fitted objects, we almost always have some numpy types or something else that would prevent a pure json object.

The largest objects are all based on ensembles of trees, and since those are not json-serializable, there is no size benefit. For smaller objects, there is a consistent size reduction with the refactored code.

### Speed

I ran the `check_persistence_performance.py` script to check if we see a speed regression.

Before refactor:

```
10 largest differences:
                             name  pickle (s)  skops (s)  abs_diff   rel_diff
0      GradientBoostingClassifier    0.003439   0.254904  0.251465  74.130723
1            ExtraTreesClassifier    0.003807   0.243327  0.239520  63.909274
2       GradientBoostingRegressor    0.003362   0.242606  0.239244  72.170761
3          RandomForestClassifier    0.003718   0.235812  0.232094  63.424460
4                 IsolationForest    0.003818   0.214642  0.210824  56.219939
5            RandomTreesEmbedding    0.003389   0.209714  0.206325  61.883283
6             ExtraTreesRegressor    0.003325   0.202813  0.199488  60.994368
7           RandomForestRegressor    0.003200   0.185760  0.182561  58.057241
8  HistGradientBoostingClassifier    0.001485   0.134543  0.133058  90.599054
9   HistGradientBoostingRegressor    0.001372   0.130100  0.128728  94.794223
No estimator was found to be unacceptably slow

```

After:

```
10 largest differences:
                             name  pickle (s)  skops (s)  abs_diff   rel_diff
0      GradientBoostingClassifier    0.003333   0.252640  0.249307  75.802290
1       GradientBoostingRegressor    0.003426   0.251955  0.248529  73.539743
2            ExtraTreesClassifier    0.004077   0.250550  0.246473  61.452242
3          RandomForestClassifier    0.003669   0.242138  0.238469  65.987126
4                 IsolationForest    0.003406   0.212873  0.209466  62.496016
5            RandomTreesEmbedding    0.003437   0.212860  0.209423  61.939385
6             ExtraTreesRegressor    0.003329   0.195246  0.191918  58.655659
7           RandomForestRegressor    0.003228   0.189995  0.186767  58.860496
8   HistGradientBoostingRegressor    0.001317   0.127843  0.126526  97.072401
9  HistGradientBoostingClassifier    0.001313   0.126494  0.125180  96.315720
No estimator was found to be unacceptably slow
```

No significant differences here.

### Readability

Using `visualize`, arguably the new variant is a bit less readable when there is a big json object, though one could argue that those are always safe so closer inspection is not necessary (we might want to truncate thos). One reason for that is the `visualize` function already takes care of removing noise.

Below are the results for an unfitted `DecisionTreeClassifier`:

Before:

```
root: sklearn.tree._classes.DecisionTreeClassifier
└── attrs: builtins.dict
    ├── criterion: json-type("gini")
    ├── splitter: json-type("best")
    ├── max_depth: json-type(null)
    ├── min_samples_split: json-type(2)
    ├── min_samples_leaf: json-type(1)
    ├── min_weight_fraction_leaf: json-type(0.0)
    ├── max_features: json-type(null)
    ├── max_leaf_nodes: json-type(null)
    ├── random_state: json-type(0)
    ├── min_impurity_decrease: json-type(0.0)
    ├── class_weight: json-type(null)
    ├── ccp_alpha: json-type(0.0)
    └── _sklearn_version: json-type("1.2.2")
```

After:

```
root: sklearn.tree._classes.DecisionTreeClassifier
└── attrs: json-type({"criterion": "gini", "splitter": "best", "max_depth": null, "min_samples_split": 2, "min_samples_leaf": 1, "min_weight_fraction_leaf": 0.0, "max_features": null, "max_leaf_nodes": null, "random_state": 0, "min_impurity_decrease": 0.0, "class_weight": 
null, "ccp_alpha": 0.0, "_sklearn_version": "1.2.2"})
```

For the schema itself, the difference is quite big though. Before:

```
{'__class__': 'DecisionTreeClassifier',
 '__id__': 140451943710096,
 '__loader__': 'ObjectNode',
 '__module__': 'sklearn.tree._classes',
 'content': {'__class__': 'dict',
             '__id__': 140450810121792,
             '__loader__': 'DictNode',
             '__module__': 'builtins',
             'content': {'_sklearn_version': {'__class__': 'str',
                                              '__id__': 140451943909104,
                                              '__loader__': 'JsonNode',
                                              '__module__': 'builtins',
                                              'content': '"1.2.2"',
                                              'is_json': True},
                         'ccp_alpha': {'__class__': 'str',
                                       '__id__': 140450862391504,
                                       '__loader__': 'JsonNode',
                                       '__module__': 'builtins',
                                       'content': '0.0',
                                       'is_json': True},
                         'class_weight': {'__class__': 'str',
                                          '__id__': 94237493811136,
                                          '__loader__': 'JsonNode',
                                          '__module__': 'builtins',
                                          'content': 'null',
                                          'is_json': True},
                         'criterion': {'__class__': 'str',
                                       '__id__': 140450799343856,
                                       '__loader__': 'JsonNode',
                                       '__module__': 'builtins',
                                       'content': '"gini"',
                                       'is_json': True},
                         'max_depth': {'__class__': 'str',
                                       '__id__': 94237493811136,
                                       '__loader__': 'JsonNode',
                                       '__module__': 'builtins',
                                       'content': 'null',
                                       'is_json': True},
                         'max_features': {'__class__': 'str',
                                          '__id__': 94237493811136,
                                          '__loader__': 'JsonNode',
                                          '__module__': 'builtins',
                                          'content': 'null',
                                          'is_json': True},
                         'max_leaf_nodes': {'__class__': 'str',
                                            '__id__': 94237493811136,
                                            '__loader__': 'JsonNode',
                                            '__module__': 'builtins',
                                            'content': 'null',
                                            'is_json': True},
                         'min_impurity_decrease': {'__class__': 'str',
                                                   '__id__': 140450862391504,
                                                   '__loader__': 'JsonNode',
                                                   '__module__': 'builtins',
                                                   'content': '0.0',
                                                   'is_json': True},
                         'min_samples_leaf': {'__class__': 'str',
                                              '__id__': 140451945382128,
                                              '__loader__': 'JsonNode',
                                              '__module__': 'builtins',
                                              'content': '1',
                                              'is_json': True},
                         'min_samples_split': {'__class__': 'str',
                                               '__id__': 140451945382160,
                                               '__loader__': 'JsonNode',
                                               '__module__': 'builtins',
                                               'content': '2',
                                               'is_json': True},
                         'min_weight_fraction_leaf': {'__class__': 'str',
                                                      '__id__': 140450862391504,
                                                      '__loader__': 'JsonNode',
                                                      '__module__': 'builtins',
                                                      'content': '0.0',
                                                      'is_json': True},
                         'random_state': {'__class__': 'str',
                                          '__id__': 140451945382096,
                                          '__loader__': 'JsonNode',
                                          '__module__': 'builtins',
                                          'content': '0',
                                          'is_json': True},
                         'splitter': {'__class__': 'str',
                                      '__id__': 140451931580208,
                                      '__loader__': 'JsonNode',
                                      '__module__': 'builtins',
                                      'content': '"best"',
                                      'is_json': True}},
             'key_types': {'__class__': 'list',
                           '__id__': 140450788312704,
                           '__loader__': 'ListNode',
                           '__module__': 'builtins',
                           'content': [{'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'},
                                       {'__class__': 'str',
                                        '__id__': 94237493788224,
                                        '__loader__': 'TypeNode',
                                        '__module__': 'builtins'}]}}}
```

After:

```
{'__class__': 'DecisionTreeClassifier',
 '__id__': 140560901242256,
 '__loader__': 'ObjectNode',
 '__module__': 'sklearn.tree._classes',
 'content': {'__class__': 'str',
             '__id__': 140559769484800,
             '__loader__': 'JsonNode',
             '__module__': 'builtins',
             'content': '{"criterion": "gini", "splitter": "best", '
                        '"max_depth": null, "min_samples_split": 2, '
                        '"min_samples_leaf": 1, "min_weight_fraction_leaf": '
                        '0.0, "max_features": null, "max_leaf_nodes": null, '
                        '"random_state": 0, "min_impurity_decrease": 0.0, '
                        '"class_weight": null, "ccp_alpha": 0.0, '
                        '"_sklearn_version": "1.2.2"}',
             'is_json': True}}
```

### DIFF
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -2,6 +2,7 @@ from __future__ import annotations
 
 import io
 import json
+import math
 import operator
 import uuid
 from functools import partial
@@ -28,7 +29,54 @@ from ._utils import (
 from .exceptions import UnsupportedTypeException
 
 
+def _check_pure_json(obj: Any) -> bool:
+    # TODO
+    type_ = type(obj)
+
+    if type_ in (int, float, str, bool, type(None)):
+        return True
+
+    if type_ == list:
+        return all(_check_pure_json(item) for item in obj)
+
+    if type_ == dict:
+        return all(
+            (type(key) == str) and _check_pure_json(val) for key, val in obj.items()
+        )
+
+    return False
+
+
+def _get_pure_json(obj: Any) -> str | None:
+    """TODO"""
+    if not _check_pure_json(obj):
+        return None
+
+    obj_json = json.dumps(obj)
+    loaded = json.loads(obj_json)
+    if loaded == obj:
+        return obj_json
+    if isinstance(obj, float) and math.isnan(obj) and math.isnan(loaded):
+        # special case for nan because nan != nan
+        return obj_json
+    return None
+
+
+def _json_get_state(obj_str: str, save_context: SaveContext) -> dict[str, Any]:
+    return {
+        "__class__": "str",
+        "__module__": "builtins",
+        "__loader__": "JsonNode",
+        "content": obj_str,
+        "is_json": True,
+    }
+
+
 def dict_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
+    pure_json_obj = _get_pure_json(obj)
+    if pure_json_obj is not None:
+        return _json_get_state(pure_json_obj, save_context)
+
     res = {
         "__class__": obj.__class__.__name__,
         "__module__": get_module(type(obj)),
@@ -75,6 +123,10 @@ class DictNode(Node):
 
 
 def list_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
+    pure_json_obj = _get_pure_json(obj)
+    if pure_json_obj is not None:
+        return _json_get_state(pure_json_obj, save_context)
+
     res = {
         "__class__": obj.__class__.__name__,
         "__module__": get_module(type(obj)),
@@ -329,18 +381,9 @@ def object_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
     # This method is for objects which can either be persisted with json, or
     # the ones for which we can get/set attributes through
     # __getstate__/__setstate__ or reading/writing to __dict__.
-    try:
-        # if we can simply use json, then we're done.
-        obj_str = json.dumps(obj)
-        return {
-            "__class__": "str",
-            "__module__": "builtins",
-            "__loader__": "JsonNode",
-            "content": obj_str,
-            "is_json": True,
-        }
-    except Exception:
-        pass
+    pure_json_obj = _get_pure_json(obj)
+    if pure_json_obj is not None:
+        return _json_get_state(pure_json_obj, save_context)
 
     res = {
         "__class__": obj.__class__.__name__,
--- a/skops/io/_numpy.py
+++ b/skops/io/_numpy.py
@@ -25,7 +25,7 @@ def ndarray_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
         # recursively call get_state on it.
         if obj.dtype == object:
             obj_serialized = get_state(obj.tolist(), save_context)
-            res["content"] = obj_serialized["content"]
+            res["content"] = obj_serialized
             res["type"] = "json"
             res["shape"] = get_state(obj.shape, save_context)
         else:
@@ -67,9 +67,7 @@ class NdArrayNode(Node):
             }
         elif self.type == "json":
             self.children = {
-                "content": [  # type: ignore
-                    get_tree(o, load_context) for o in state["content"]  # type: ignore
-                ],
+                "content": get_tree(state["content"], load_context),
                 "shape": get_tree(state["shape"], load_context),
             }
         else:
@@ -87,7 +85,7 @@ class NdArrayNode(Node):
             # We explicitly set the dtype to "O" since we only save object
             # arrays in json.
             shape = self.children["shape"].construct()
-            tmp = [o.construct() for o in self.children["content"]]
+            tmp = self.children["content"].construct()
 
             # TODO: this is a hack to get the correct shape of the array. We
             # should find _a better way_ to do this.
--- a/skops/io/_sklearn.py
+++ b/skops/io/_sklearn.py
@@ -104,12 +104,12 @@ class ReduceNode(Node):
         self.children = {
             "attrs": get_tree(state["content"], load_context),
             "args": get_tree(reduce["args"], load_context),
-            "constructor": constructor,
         }
+        self.constructor = constructor
 
     def _construct(self):
         args = self.children["args"].construct()
-        constructor = self.children["constructor"]
+        constructor = self.constructor
         instance = constructor(*args)
         attrs = self.children["attrs"].construct()
         if not attrs:
--- a/skops/io/tests/test_persist.py
+++ b/skops/io/tests/test_persist.py
@@ -536,7 +536,7 @@ def test_metainfo():
     # additionally, check following metainfo: class, module, and version
     expected = {
         "builtin_": {
-            "__class__": "list",
+            "__class__": "str",
             "__module__": "builtins",
         },
         "stdlib_": {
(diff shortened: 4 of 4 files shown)


adrinjalali:
Looking at it, I think I see a bug:

```
                         'min_impurity_decrease': {'__class__': 'str',
                                                   '__id__': 140450862391504,
                                                   '__loader__': 'JsonNode',
                                                   '__module__': 'builtins',
                                                   'content': '0.0',
                                                   'is_json': True},
```
why is the type `str` here? It should be `float` I think, shouldn't it?

What I'm thinking is, here with the change we lose dtypes of the objects, but I see now the dtypes are wrong as they are. Hmm...

adrinjalali:
What I mean is, ideally we'd let json read the object, but then cast it with the type given in the type slot. Otherwise we're probably changing the types of numbers.

BenjaminBossan:
> What I mean is, ideally we'd let json read the object, but then cast it with the type given in the type slot. Otherwise we're probably changing the types of numbers.

I see. Yes, I was wondering if may sometimes get the wrong numerical type, but at least our unit tests, where we check type, don't detect any incorrect conversion. If we do want to coerce types, we can't proceed with the refactor, no matter which version.

adrinjalali:
We can probably at least optimize the dict node, so that we dump the values with json, but we keep dtypes of values as well as keys maybe?

BenjaminBossan:
We could do that for `DictNode` (and also for `ListNode`, which may also persist valid json). If I understand your suggestion correctly, though, the end result would not be much more compact or readable than the existing solution. So maybe let's just stick with the existing solution? And maybe set the actual value's type in `__class__` instead of `str` (which would be BC breaking, so require to preserve the old implementation).

adrinjalali:
Yeah I think that'd make sense.

PR #336, opened 2023-04-01
#224: trusting np.ufuncs and np.dtypes

### LINKED ISSUE #224: Add {numpy, scipy}.ufuncs to the trusted list

omar-araboghli:
#### Reference Issues/PRs
closes #224 

Trusting `numpy.ufuncs` and `numpy.dtypes` by default.

#### Any other comments?
1. I compared the ufuncs list obtained in this PR with the [available ufuncs](https://numpy.org/doc/stable/reference/ufuncs.html#available-ufuncs). They are almost the same except the aliases. Take `true_divide` as an example that is an alias for `divide` and unfortunately cannot be found with the `dir(...)` and `isinstance(...)` approach. Is there a way to find numpy aliases ?
2. `test_can_persist_fitted` didn't run successfully until I added `NUMPY_DTYPE_TYPE_NAMES` to the default trusted list of `NdArrayNode`. I am wondering why isn't `DTypeNode` the correct place ? Is there any other Nodes I am missing ?

### DIFF
--- a/skops/io/_trusted_types.py
+++ b/skops/io/_trusted_types.py
@@ -24,3 +24,25 @@ SCIPY_UFUNC_TYPE_NAMES = sorted(
         ]
     )
 )
+
+NUMPY_UFUNC_TYPE_NAMES = sorted(
+    set(
+        [
+            get_type_name(getattr(np, attr))
+            for attr in dir(np)
+            if isinstance(getattr(np, attr), np.ufunc)
+            and get_type_name(getattr(np, attr)).startswith("numpy")
+        ]
+    )
+)
+
+NUMPY_DTYPE_TYPE_NAMES = sorted(
+    set(
+        [
+            get_type_name(dtype)
+            for dtypes in np.sctypes.values()
+            for dtype in dtypes
+            if get_type_name(dtype).startswith("numpy")
+        ]
+    )
+)
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -13,6 +13,8 @@ import numpy as np
 from ._audit import Node, get_tree
 from ._protocol import PROTOCOL
 from ._trusted_types import (
+    NUMPY_DTYPE_TYPE_NAMES,
+    NUMPY_UFUNC_TYPE_NAMES,
     PRIMITIVE_TYPE_NAMES,
     SCIPY_UFUNC_TYPE_NAMES,
     SKLEARN_ESTIMATOR_TYPE_NAMES,
@@ -197,7 +199,9 @@ class FunctionNode(Node):
     ) -> None:
         super().__init__(state, load_context, trusted)
         # TODO: what do we trust?
-        self.trusted = self._get_trusted(trusted, default=SCIPY_UFUNC_TYPE_NAMES)
+        self.trusted = self._get_trusted(
+            trusted, default=SCIPY_UFUNC_TYPE_NAMES + NUMPY_UFUNC_TYPE_NAMES
+        )
         self.children = {}
 
     def _construct(self):
@@ -278,7 +282,9 @@ class TypeNode(Node):
     ) -> None:
         super().__init__(state, load_context, trusted)
         # TODO: what do we trust?
-        self.trusted = self._get_trusted(trusted, PRIMITIVE_TYPE_NAMES)
+        self.trusted = self._get_trusted(
+            trusted, PRIMITIVE_TYPE_NAMES + NUMPY_DTYPE_TYPE_NAMES
+        )
         # We use a bare Node type here since a Node only checks the type in the
         # dict using __class__ and __module__ keys.
         self.children = {}
--- a/skops/io/_numpy.py
+++ b/skops/io/_numpy.py
@@ -8,6 +8,7 @@ import numpy as np
 from ._audit import Node, get_tree
 from ._general import function_get_state
 from ._protocol import PROTOCOL
+from ._trusted_types import NUMPY_DTYPE_TYPE_NAMES
 from ._utils import LoadContext, SaveContext, get_module, get_state, gettype
 from .exceptions import UnsupportedTypeException
 
@@ -60,7 +61,7 @@ class NdArrayNode(Node):
     ) -> None:
         super().__init__(state, load_context, trusted)
         self.type = state["type"]
-        self.trusted = self._get_trusted(trusted, [np.ndarray])
+        self.trusted = self._get_trusted(trusted, [np.ndarray] + NUMPY_DTYPE_TYPE_NAMES)
         if self.type == "numpy":
             self.children = {
                 "content": io.BytesIO(load_context.src.read(state["file"]))
--- a/skops/io/tests/test_persist.py
+++ b/skops/io/tests/test_persist.py
@@ -53,7 +53,13 @@ import skops
 from skops.io import dump, dumps, get_untrusted_types, load, loads
 from skops.io._audit import NODE_TYPE_MAPPING, get_tree
 from skops.io._sklearn import UNSUPPORTED_TYPES
-from skops.io._trusted_types import SCIPY_UFUNC_TYPE_NAMES, SKLEARN_ESTIMATOR_TYPE_NAMES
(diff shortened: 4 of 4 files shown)


skops/io/tests/test_audit.py
```diff
@@ -151,9 +151,9 @@ def test_complex_pipeline_untrusted_set():
     clf = Pipeline([
         ("features", FeatureUnion([
             ("scaler", StandardScaler()),
-            ("sqrt", FunctionTransformer(
-                    func=np.sqrt,
-                    inverse_func=np.square,
+            ("np.funcs", FunctionTransformer(
+                    func=np.split,
+                    inverse_func=np.angle,
```
adrinjalali, line 156: This doesn't actually make sense now, does it? We can probably remove this case here, and bellow we have tests to make sure ufuncs are trusted. WDYT @BenjaminBossan

skops/io/tests/test_audit.py
```diff
@@ -162,7 +162,9 @@ def test_complex_pipeline_untrusted_set():
 
     untrusted = get_untrusted_types(data=dumps(clf))
     type_names = [x.split(".")[-1] for x in untrusted]
-    assert type_names == ["sqrt", "square"]
+
+    # choosing random numpy functions that are yet not considered as default trusted ones.
```
adrinjalali, line 166: line too long

adrinjalali (commented):
Thanks @omar-araboghli

skops/io/_trusted_types.py
```diff
@@ -24,3 +24,25 @@
...
+            get_type_name(getattr(np, attr))
+            for attr in dir(np)
+            if isinstance(getattr(np, attr), np.ufunc)
+            and get_type_name(getattr(np, attr)).startswith("numpy")
+        ]
+    )
+)
+
+NUMPY_DTYPE_TYPE_NAMES = sorted(
+    set(
```
BenjaminBossan, line 40: Same argument

skops/io/tests/test_audit.py
BenjaminBossan: Yes, I think this can be removed, as other tests cover what this covers.

BenjaminBossan (changes requested):
> Take true_divide as an example that is an alias for divide and unfortunately cannot be found with the dir(...) and isinstance(...) approach. Is there a way to find numpy aliases

Not that I know of, but it's better to trust too little than too much, so it's fine for now.

> test_can_persist_fitted didn't run successfully until I added NUMPY_DTYPE_TYPE_NAMES to the default trusted list of NdArrayNode

This might be a bug, let me investigate before we merge this.

BenjaminBossan:
> > test_can_persist_fitted didn't run successfully until I added NUMPY_DTYPE_TYPE_NAMES to the default trusted list of NdArrayNode
> 
> This might be a bug, let me investigate before we merge this.

Okay, I believe this is not a bug. The reason is that we use `NdArrayNode` not only for numpy arrays, but also for numpy generics. If e.g. a numpy float is saved, it will use `NdArrayNode`, which is why the numpy dtypes have to be trusted there. This is indeed a bit confusing, maybe another name for the class would have been better.

Another source of confusion (at least for me) is what is meant by `dtype`. Here we mean e.g. a `numpy.float64`. However, there is also the "dtype" type, e.g. `numpy.dtype[float64]`, which is persisted using the `DtypeNode`.

At the end of the day, I think everything is working here as expected, but the naming can cause confusion. What's your opinion @adrinjalali?

adrinjalali:
Changing the names here wouldn't be backward compatible, and will probably take some time, and I also don't have very good ideas for their names. So I think we can leave that for a future iteration.

BenjaminBossan:
> Changing the names here wouldn't be backward compatible, and will probably take some time, and I also don't have very good ideas for their names. So I think we can leave that for a future iteration.

True, we should probably add a comment though.

@omar-araboghli I think the docs also need to be updated here: https://skops.readthedocs.io/en/stable/persistence.html#usage

skops/io/_utils.py
```diff
@@ -200,3 +201,36 @@ def get_type_paths(types: Any) -> list[str]:
         types = [types]
 
     return [get_type_name(t) if not isinstance(t, str) else t for t in types]
+
+
+def get_public_type_names(module: ModuleType, _type: Type) -> list[str]:
```
adrinjalali, line 206: maybe rename `_type` to `oftype`?

skops/io/_utils.py
```diff
@@ -200,3 +201,36 @@ def get_type_paths(types: Any) -> list[str]:
...
+        The sorted list of type names, all as strings,
+         e.g. ``["numpy.core._multiarray_umath.absolute"]``.
+    """
+    module_name, _, _ = module.__name__.rpartition(".")
+
+    return sorted(
+        {
+            type_name
+            for attr in dir(module)
+            if (isinstance(obj := getattr(module, attr), _type))
```
adrinjalali, line 233: `isinstance` or `issubclass`?

adrinjalali (commented):
Otherwise it's looking great I think.

skops/io/_utils.py
omar-araboghli: changing it to 

```suggestion
            if (issubclass((obj := getattr(module, attr)).__class__, _type))
```

gets the same results for both numpy and  scipy. Since the ufuncs are classes and not objects (instances), `issubclass` makes more sense to use here. Thanks for the suggestion!

adrinjalali (approved):

BenjaminBossan (changes requested):
Looks quite good, thx. Only two minor issues for me, otherwise the PR can be merged.

BenjaminBossan:
@omar-araboghli is this ready for review?

omar-araboghli:
> @omar-araboghli is this ready for review?

Indeed! Sorry for not pingging you.

BenjaminBossan (approved):
Thanks for implementing the suggestions, this looks good now. You'd need to merge the current main and fix the potential merge conflict in `changes.rst`, then this is good to go.

It would have been nice if `NUMPY_DTYPE_TYPE_NAMES` could also use `get_public_type_names` but it wouldn't work as is. If you happen to come up with a clever solution to that works for both, it would be good to make that change, but it's not a blocker.

BenjaminBossan:
I just noticed that [`test_external.py`](https://github.com/skops-dev/skops/blob/main/skops/io/tests/test_external.py) can also be cleaned up, as it contains a couple of numpy dtypes that no longer need to be trusted explicitly.

omar-araboghli:
@BenjaminBossan just resolved the conflict and cleaned `test_externals.py` as well as `test_visualize.py` up. I unfortunately couldn't come up with a generic solution to have `NUMPY_DTYPE_TYPE_NAMES` use `get_public_type_names` since numpy doesn't have a superclass covering them. Getting them from `np.sctypes` seems to be the simplest way to go with. Any hint on that or shall we keep it as it as for now ?

skops/io/_utils.py
```diff
@@ -200,3 +201,36 @@ def get_type_paths(types: Any) -> list[str]:
         types = [types]
 
     return [get_type_name(t) if not isinstance(t, str) else t for t in types]
+
+
+def get_public_type_names(module: ModuleType, oftype: Type) -> list[str]:
+    """
+    Helper function that gets the type names of all
+    public objects of the given ``_type`` from the given ``module``,
```
BenjaminBossan, line 209:
```suggestion
    public objects of the given ``oftype`` from the given ``module``,
```

BenjaminBossan (changes requested):
Great work cleaning up the remaining instances, this looks very good now. There is a small typo in one of the docstrings, besides that this can be merged.

> I unfortunately couldn't come up with a generic solution

Then it's fine, it would have been nice but it's not a requirement.

omar-araboghli:
@BenjaminBossan Thanks for the review and great catch! The typo is now resolved, thanks to @adrinjalali :)

PR #340, opened 2023-04-05
MNT: A little hardening of the auditing of Nodes

BenjaminBossan:
## Description

Two measures to harden the auditing (a little bit):

- Type annotate the `Node`'s children to prevent setting invalid types.
- Change all the tests that use `loads` to only load trusted types instead of using `trusted=True`

The latter is important because when setting `trusted=True`, the whole machinery of checking types is not executed, so any bugs that may be contained there will not be revealed. In particular, this shows that for persisting methods, we had a child with a str type and that would raise an error, i.e. so far, loading method types was not possible for users who passed `trusted!=True`.

## Additional changes

As a consequence of the last point, the auditing code has been changed to accept str as type. Alternatively, we can make the change explained here:

https://github.com/skops-dev/skops/pull/338#discussion_r1156151738

i.e. not storing the method name in children.

Another "victim" of this PR is that the so far dead code of checking for primitive types inside of `get_unsafe_set` has been removed. This code was supposed to check if the type is a primitive type but it was defective. `get_module(child)` would raise an error if an instance of the type would be passed. We could theoretically fix that code, but it would still be dead code because primitive types are stored as json.

Another small change is to exclude the code in `skops/io/old` from mypy checks. Otherwise, we would have to update its type signatures if signatures in the persistence code change (as they did here).

### DIFF
--- a/skops/io/_audit.py
+++ b/skops/io/_audit.py
@@ -2,7 +2,7 @@ from __future__ import annotations
 
 import io
 from contextlib import contextmanager
-from typing import Any, Generator, Literal, Optional, Sequence, Type, Union
+from typing import Any, Dict, Generator, List, Literal, Optional, Sequence, Type, Union
 
 from ._protocol import PROTOCOL
 from ._utils import LoadContext, get_module, get_type_paths
@@ -10,7 +10,7 @@ from .exceptions import UntrustedTypesFoundException
 
 NODE_TYPE_MAPPING: dict[tuple[str, int], Node] = {}
 VALID_NODE_CHILD_TYPES = Optional[
-    Union["Node", list["Node"], dict[str, "Node"], Type, str, io.BytesIO]
+    Union["Node", List["Node"], Dict[str, "Node"], Type, str, io.BytesIO]
 ]
 
 
--- a/skops/io/_audit.py
+++ b/skops/io/_audit.py
@@ -2,14 +2,16 @@ from __future__ import annotations
 
 import io
 from contextlib import contextmanager
-from typing import Any, Generator, Literal, Sequence, Type, Union
+from typing import Any, Generator, Literal, Optional, Sequence, Type, Union
 
 from ._protocol import PROTOCOL
-from ._trusted_types import PRIMITIVE_TYPE_NAMES
 from ._utils import LoadContext, get_module, get_type_paths
 from .exceptions import UntrustedTypesFoundException
 
 NODE_TYPE_MAPPING: dict[tuple[str, int], Node] = {}
+VALID_NODE_CHILD_TYPES = Optional[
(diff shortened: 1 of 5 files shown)


BenjaminBossan:

skops/io/_audit.py
```diff
@@ -269,15 +271,11 @@ def get_unsafe_set(self) -> set[str]:
...
                         # if the child is a type, we check its safety
                         res.add(get_module(child) + "." + child.__name__)
-                elif isinstance(child, io.BytesIO):
+                elif isinstance(child, (io.BytesIO, str)):
                     # We trust BytesIO objects, which are read by other
-                    # libraries such as numpy, scipy.
-                    continue
-                elif check_type(
-                    get_module(child), child.__class__.__name__, PRIMITIVE_TYPE_NAMES
-                ):
```
adrinjalali, line 278: are we removing them cause now primitives are trusted by default?

skops/io/_visualize.py
```diff
@@ -168,7 +168,7 @@ def pretty_print_tree(
 
 
 def walk_tree(
-    node: Node | dict[str, Node] | list[Node],
+    node: VALID_NODE_CHILD_TYPES | dict[str, VALID_NODE_CHILD_TYPES],
```
adrinjalali, line 171: isn't `dict(str, Node)` included in `VALID_NODE_CHILD_TYPES`?

adrinjalali (commented):
Defining type annotations like `VALID_NODE_CHILD_TYPES` makes the code much less readable to me, but I guess we can let it be.

skops/io/_audit.py
BenjaminBossan: This is what I was referring to here:

> Another "victim" of this PR is that the so far dead code of checking for primitive types inside of `get_unsafe_set` has been removed. This code was supposed to check if the type is a primitive type but it was defective. `get_module(child)` would raise an error if an instance of the type would be passed. We could theoretically fix that code, but it would still be dead code because primitive types are stored as json.

skops/io/_visualize.py
BenjaminBossan: Yes, but this is `dict[str, VALID_NODE_CHILD_TYPES]`, so it could be something where the key is not a `Node`, like `{"foo": None}`

BenjaminBossan (commented):
> Defining type annotations like VALID_NODE_CHILD_TYPES makes the code much less readable to me, but I guess we can let it be.

Yes, it is a tradeoff with readability. I originally had the type definition on the line that it's used, which really adds a lot of noise when reading the `Node` code, I think this is an okay compromise, as it the eye can quickly skip it. We could use a shorter alias if that helps.

adrinjalali (approved):
Cool!

PR #343, opened 2023-04-10
ENH: Added CLI command to update skops files

### LINKED ISSUE #333: CLI: Command to update a skops file
The skops persistence format is updated over time, which may lead to bugs being fixed, performance being improved, reduced file size, etc. Therefore, as a user, I may want to upgrade my skops files periodically to ensure I can take advantage of the changes. Right now, I would need to do that manually but it would be nice for the skops CLI to offer this possibility.

Some optional features that come to mind:

- allow updating multiple files at once
- report if files where upgraded or nothing was changed
- some kind of check that upgraded file produces same object
- probably out of scope: upgrade to specified protcol version (or even downgrade)

EdAbati:
#### Reference Issues/PRs
Closes #333 

Taking inspiration from the implementation from `skops convert`, this adds the `update` command to the CLI.
It allows to update a `skops` file by simply 'loading' it and 'dumping' it again.

I have also added MacOS files in the gitignore.

### DIFF
--- a/skops/cli/_update.py
+++ b/skops/cli/_update.py
@@ -1,0 +1,118 @@
+from __future__ import annotations
+
+import argparse
+import logging
+import os
+import pathlib
+
+# import pickle
+from typing import Optional
+
+from skops.cli._utils import get_log_level
+
+# from skops.io import dumps, get_untrusted_types
+
+
+def _update_file(
+    input_file: os.PathLike,
+    output_file: os.PathLike,
+    logger: logging.Logger = logging.getLogger(),
+) -> None:
+    """Function that is called by ``skops update`` entrypoint.
+
+    Loads a skops model from the input path, converts to the latest skops format, and saves to
+    output file.
+
+    Parameters
+    ----------
+    input_file : os.PathLike
+        Path of input .pkl model to load.
+
+    output_file : os.PathLike
+        Path to save .skops model to.
+
+    """
+    # model_name = pathlib.Path(input_file).stem
+
+    # logger.debug(f"Converting {model_name}")
+
+    # with open(input_file, "rb") as f:
+    #     obj = pickle.load(f)
+    # skops_dump = dumps(obj)
+
+    # untrusted_types = get_untrusted_types(data=skops_dump)
+
+    # if not untrusted_types:
+    #     logger.info(f"No unknown types found in {model_name}.")
+    # else:
+    #     untrusted_str = ", ".join(untrusted_types)
+
+    #     logger.warning(
+    #         f"While converting {input_file}, "
+    #         "the following unknown types were found: "
+    #         f"{untrusted_str}. "
+    #         f"When loading {output_file} with skops.load, these types must be "
+    #         "specified as 'trusted'"
+    #     )
+
+    # with open(output_file, "wb") as out_file:
+    #     logger.debug(f"Writing to {output_file}")
+    #     out_file.write(skops_dump)
+    raise NotImplementedError
+
+
+def format_parser(
+    parser: Optional[argparse.ArgumentParser] = None,
+) -> argparse.ArgumentParser:
+    """Adds arguments and help to parent CLI parser for the `update` method."""
+
+    if not parser:  # used in tests
+        parser = argparse.ArgumentParser()
+
+    parser_subgroup = parser.add_argument_group("update")
+    parser_subgroup.add_argument("input", help="Path to an input file to update. ")
+
+    parser_subgroup.add_argument(
+        "-o",
+        "--output-file",
+        help=(
+            # TODO: decide what to do with this. Default name? or compulsory?
+            "Specify the output file name for the updated skops file. "
+            "If not provided, will default to using the same name as the input file, "
+            "and saving to the current working directory with the suffix '.skops'."
+        ),
+        default=None,
+    )
+    parser_subgroup.add_argument(
+        "-v",
+        "--verbose",
+        help=(
+            "Increases verbosity of logging. Can be used multiple times to increase "
+            "verbosity further."
+        ),
+        action="count",
+        dest="loglevel",
+        default=0,
+    )
+    return parser
+
+
+def main(
+    parsed_args: argparse.Namespace,
+) -> None:
+    output_file = parsed_args.output_file
+    input_file = parsed_args.input
+
+    logging.basicConfig(
+        format="%(levelname)-8s: %(message)s", level=get_log_level(parsed_args.loglevel)
+    )
+
+    if not output_file:
+        # No filename provided, defaulting to base file path
+        file_name = pathlib.Path(input_file).stem
+        output_file = pathlib.Path.cwd() / f"{file_name}.skops"
+
+    _update_file(
+        input_file=input_file,
+        output_file=output_file,
+    )
--- a/skops/cli/_update.py
+++ b/skops/cli/_update.py
@@ -2,63 +2,35 @@ from __future__ import annotations
 
 import argparse
 import logging
-import os
-import pathlib
-
-# import pickle
-from typing import Optional
+from pathlib import Path
+from typing import Optional, Union
 
 from skops.cli._utils import get_log_level
-
-# from skops.io import dumps, get_untrusted_types
+from skops.io import dump, load
 
 
 def _update_file(
-    input_file: os.PathLike,
-    output_file: os.PathLike,
+    input_file: Union[str, Path],
+    output_file: Union[str, Path],
     logger: logging.Logger = logging.getLogger(),
 ) -> None:
     """Function that is called by ``skops update`` entrypoint.
 
-    Loads a skops model from the input path, converts to the latest skops format, and saves to
-    output file.
+    Loads a skops model from the input path, converts to the current skops format, and
+    saves to output file.
 
     Parameters
     ----------
-    input_file : os.PathLike
-        Path of input .pkl model to load.
+    input_file : Union[str, Path]
+        Path of input skops model to load.
 
-    output_file : os.PathLike
-        Path to save .skops model to.
+    output_file : Union[str, Path]
+        Path to save the updated skops model to.
 
     """
-    # model_name = pathlib.Path(input_file).stem
-
-    # logger.debug(f"Converting {model_name}")
-
-    # with open(input_file, "rb") as f:
-    #     obj = pickle.load(f)
-    # skops_dump = dumps(obj)
-
-    # untrusted_types = get_untrusted_types(data=skops_dump)
-
-    # if not untrusted_types:
-    #     logger.info(f"No unknown types found in {model_name}.")
-    # else:
-    #     untrusted_str = ", ".join(untrusted_types)
-
-    #     logger.warning(
-    #         f"While converting {input_file}, "
-    #         "the following unknown types were found: "
-    #         f"{untrusted_str}. "
-    #         f"When loading {output_file} with skops.load, these types must be "
-    #         "specified as 'trusted'"
-    #     )
-
-    # with open(output_file, "wb") as out_file:
-    #     logger.debug(f"Writing to {output_file}")
-    #     out_file.write(skops_dump)
-    raise NotImplementedError
+    input_model = load(input_file, trusted=True)
+    dump(input_model, output_file)
+    logger.debug(f"Updated skops file written in {output_file}")
 
 
 def format_parser(
@@ -75,13 +47,7 @@ def format_parser(
     parser_subgroup.add_argument(
         "-o",
         "--output-file",
-        help=(
-            # TODO: decide what to do with this. Default name? or compulsory?
-            "Specify the output file name for the updated skops file. "
-            "If not provided, will default to using the same name as the input file, "
-            "and saving to the current working directory with the suffix '.skops'."
-        ),
-        default=None,
+        help="Specify the output file name for the updated skops file.",
     )
     parser_subgroup.add_argument(
         "-v",
@@ -100,18 +66,12 @@ def format_parser(
 def main(
     parsed_args: argparse.Namespace,
 ) -> None:
-    output_file = parsed_args.output_file
-    input_file = parsed_args.input
+    output_file = Path(parsed_args.output_file)
+    input_file = Path(parsed_args.input)
 
     logging.basicConfig(
         format="%(levelname)-8s: %(message)s", level=get_log_level(parsed_args.loglevel)
     )
-
-    if not output_file:
-        # No filename provided, defaulting to base file path
-        file_name = pathlib.Path(input_file).stem
-        output_file = pathlib.Path.cwd() / f"{file_name}.skops"
-
     _update_file(
         input_file=input_file,
         output_file=output_file,
--- a/.gitignore
+++ b/.gitignore
@@ -117,6 +117,8 @@ node_modules
 # Vim
 *.swp
 
+# MacOS
+.DS_Store
 
 exports
 trash
--- a/skops/cli/entrypoint.py
+++ b/skops/cli/entrypoint.py
@@ -1,6 +1,7 @@
 import argparse
 
 import skops.cli._convert
+import skops.cli._update
 
 
 def main_cli(command_line_args=None):
@@ -31,6 +32,10 @@ def main_cli(command_line_args=None):
         "convert": {
             "method": skops.cli._convert.main,
             "format_parser": skops.cli._convert.format_parser,
+        },
+        "update": {
+            "method": skops.cli._update.main,
+            "format_parser": skops.cli._update.format_parser,
         },
     }
 


BenjaminBossan (changes requested):
Thanks a lot for working on this.

Before going further, we need a few more things: tests and documentation (see `docs/persistence.rst`).

IMO it would also be nice to have a feature to check the protocol of the loaded file and if it's the same as the current protocol, to skip the conversion. If it's too complicated, we can also leave this open for a future PR, though.

EdAbati:
Hi @BenjaminBossan 

I finally moved this PR from "Draft" to "ready for review". 🙂

You can see that, for now, I have not implemented any extra checks that we discussed here and in the issue. I wanted to make sure I’m on the right track before adding anything else.

The interfaces of the functions are slightly different from the ones in `_convert`. I am passing a `logger` argument to make testing easier.

BenjaminBossan:
I did a quick pass and it looks good to me overall. @adrinjalali would you be fine with this PR as is, i.e. no extra checks etc.?

adrinjalali:
I'm happy as is, without the checks, but we should figure out the checks, we've needed them in a few places now.

BenjaminBossan (changes requested):
Thanks a lot, this is a very clean implementation, well done.

There are only a few minor code comments from my side, please take a look.

Regarding the more general points, I have two suggestions (not must haves):

1. How about making it possible to call the command without `-o` to just overwrite the existing file? I.e. `skops update my_model.skops` would write the new, updated skops file to `my_model.skops`.

2. There is no test for the actual work that is being done. So e.g. if the `_update_file` function would just create a copy, the tests would still pass fine. I know that creating a file in an old skops format to test the updating feature is not trivial. Maybe you could come up with something clever? Say, create the skops file, manually set the protocol to an older version, then checking after the update that the protocol is the current version now. WDYT?

skops/cli/tests/test_entrypoint.py
```diff
@@ -39,3 +39,23 @@ def test_convert_works_as_expected(
...
 
         assert caplog.at_level(logging.WARNING)
+
+    @mock.patch("skops.cli._update._update_file")
+    def test_update_works_as_expected(
+        self,
+        update_file_mock: mock.MagicMock,
+    ):
+        """
+        Intended as a unit test to make sure,
```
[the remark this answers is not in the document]
EdAbati: good point 😅 I copied and pasted from the test above. I will change

EdAbati:
Thank you very much for your review!

> 1. How about making it possible to call the command without `-o' to just overwrite the existing file? I.e. `skops update my_model.skops` would write the new, updated skops file to `my_model.skops`.

I actually did that initially, but then I thought that if we did that by default, it could be a negative surprise for certain users that want to keep the original file. Another idea was to attach a prefix or a suffix to the file name provided as input if no `output-file` is provided. For example: `my-model.skops` could be updated by default to `my-model-updated.skops`. In the end, this is a design decision; I'm happy to do what you think is best :) 

> 2. There is no test for the actual work that is being done. So e.g. if the `_update_file` function would just create a copy, the tests would still pass fine. I know that creating a file in an old skops format to test the updating feature is not trivial. Maybe you could come up with something clever? Say, create the skops file, manually set the protocol to an older version, then checking after the update that the protocol is the current version now. WDYT?

You are right! I added the protocol check inside the `update` function so that we can also test it :) Let me know if it's what you had in mind

skops/cli/_update.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+    input_model = load(input_file, trusted=True)
+    with zipfile.ZipFile(input_file, "r") as zip_file:
+        input_file_schema = json.loads(zip_file.read("schema.json"))
+
+    if input_file_schema["protocol"] != PROTOCOL:
+        dump(input_model, output_file)
+        logger.info(f"Updated skops file written to {output_file}")
+
+    logger.info(f"Input file is already up to date to the current protocol: {PROTOCOL}")
```
BenjaminBossan, line 42:
```suggestion
    logger.info(f"Input file is already up to date to the current protocol: {PROTOCOL}")
```

Should this not be in the `else` clause? Also, I prefer a wording that says what was actually done, something like "File was not updated because ...".

skops/cli/tests/test_update.py
```diff
@@ -0,0 +1,130 @@
...
+        skops_path: pathlib.Path,
+        new_skops_path: pathlib.Path,
+        dump_file,
+    ):
+        mock_logger = mock.MagicMock()
+        _update._update_file(
+            input_file=skops_path, output_file=new_skops_path, logger=mock_logger
+        )
+
+        with pytest.raises(FileNotFoundError):
```
BenjaminBossan, line 73: How about asserting that the file does not exist, instead of trying to load and raising an error? Also, could we check that the log info contains the "Input file is already up to date..." message?

BenjaminBossan (changes requested):
Thanks for implementing the feedback. Clever way of patching the protocol.

I think there is a bug with one of the log messages, please take a look at my comment. Otherwise, I just left some comments on how to potentially tighten the tests.

> I actually did that initially, but then I thought that if we did that by default, it could be a negative surprise for certain users that want to keep the original file. Another idea was to attach a prefix or a suffix to the file name provided as input if no `output-file` is provided. For example: `my-model.skops` could be updated by default to `my-model-updated.skops`. In the end, this is a design decision; I'm happy to do what you think is best :)

Good points. My intuition for being okay with allowing to overwrite the existing file is that other terminal commands also allow to just overwrite existing files without warning (say, `mv`). Adding a prefix/suffix would be a way to prevent accidents but has its own complications (what if `my-model-updated.skops` already exists?). @adrinjalali wanna break the tie?

BenjaminBossan:
> only print info about the diffs by default, no change in any file, no new files created

Hmm, how exactly would that diff look like for a skops file?

> `-i` would change file in place
> 
> `-o` creates a new file

I can get behind the idea to allow inplace but requiring an extra argument. Not sure about `-i`, as it is often used to mark the input (especially when `-o` is used for the output). If there is some precedence for a given argument name, we should probably use that. *Maybe* `-f` or `--force` to force overwriting but IMO it doesn't quite fit here. `cp` has `-n` (no clobber) to _prevent_ overwriting an existing file, making overwriting the default, but I don't like that too much. If nothing else comes up, we could require the full `--inplace`.

adrinjalali:
> Hmm, how exactly would that diff look like for a skops file?

we can start by reporting if there'd be any change at all or not, we can improve the diff maybe later if needed. It could be something like the diff of the json schema, and reporting any other saved file (like numpy files) if they're changed.

> I can get behind the idea to allow inplace but requiring an extra argument. Not sure about `-i`, as it is often used to mark the input (especially when `-o` is used for the output). If there is some precedence for a given argument name, we should probably use that. *Maybe* `-f` or `--force` to force overwriting but IMO it doesn't quite fit here. `cp` has `-n` (no clobber) to _prevent_ overwriting an existing file, making overwriting the default, but I don't like that too much. If nothing else comes up, we could require the full `--inplace`.

Input is usually just an argument, w/o a parameter. `grep` uses `-i` for case insensitive for instance. But I don't really mind, I'm also happy with `--inplace`.

EdAbati:
Hi everyone, thank you very much for your feedback. 🙂

I have updated the PR, now the arguments of the CLI are:
```
$ skops update --help
usage: Skops update [-h] [-o OUTPUT_FILE] [--inplace] [-v] input

options:
  -h, --help            show this help message and exit

update:
  input                 Path to an input file to update.
  -o OUTPUT_FILE, --output-file OUTPUT_FILE
                        Specify the output file name for the updated skops file.
  --inplace             Update and overwrite the input file in place.
  -v, --verbose         Increases verbosity of logging. Can be used multiple times to increase verbosity further.
  ```

File: docs/persistence.rst [hunk 4]
```diff
@@ -129,8 +129,13 @@ for more details.
...
 
-Skops has a command line interface to convert scikit-learn models persisted with
-``Pickle`` to ``Skops`` files.
+Skops has a command line interface to:
+
+- convert scikit-learn models persisted with ``Pickle`` to ``Skops`` files.
+- update ``Skops`` files to the latest version.
+
+``skops convert``
+-----------------
```
BenjaminBossan, line 138: The headings of the docs are messed up:

Screenshot from 2023-05-02 11-32-00

I believe the problem is the use of `----` here. If you look at the top of the file, the title uses `====` (level 1) and the section under the title, "Usage", uses `----` (level 2), and the subsection of "Usage", "Command Line Interface", uses `####` (level 3). Thus, the two subcommands, "skops convert" and "skops update", cannot also use `----` (level 2), they have to use a different type of heading, e.g. `~~~~` (level 4).

I would also be fine to elevate "Command Line Interface" to the same level as "Usage", i.e. to use `----`, but this would still require changing the `----` from the subcommands. Also, "Visualization" below would also need to be elevated (to level 2) in that case.

Yeah, the way rst handles sections is ... not optimal.

skops/cli/_update.py
```diff
@@ -0,0 +1,122 @@
...
+def _update_file(
+    input_file: str | Path,
+    output_file: str | Path | None = None,
+    inplace: bool = False,
+    logger: logging.Logger = logging.getLogger(),
+) -> None:
+    """Function that is called by ``skops update`` entrypoint.
+
+    Loads a skops model from the input path, converts to the current skops format, and
+    saves to output file.
```
BenjaminBossan, line 23: This sentence needs a slight update with the `inplace` flag.

skops/cli/_update.py
```diff
@@ -0,0 +1,122 @@
...
+                "Cannot specify both an output file path and the inplace flag. Please"
+                " choose whether you want to create a new file or overwrite the input"
+                " file."
+            )
+
+    input_model = load(input_file, trusted=True)
+    with zipfile.ZipFile(input_file, "r") as zip_file:
+        input_file_schema = json.loads(zip_file.read("schema.json"))
+
+    if input_file_schema["protocol"] != PROTOCOL:
```
BenjaminBossan, line 53: I'm just wondering about the case of the input file having a _higher_ protocol than the current one. Might be a niche application, but could have it's uses. The code would already work as is, but the message might be confusing. I wonder: Do we want to change the wording to "upgrade" / "downgrade" depending on whether the protocol was higher or lower?

skops/cli/_update.py
```diff
@@ -0,0 +1,122 @@
...
+    if inplace:
+        if output_file is None:
+            output_file = input_file
+        else:
+            raise ValueError(
+                "Cannot specify both an output file path and the inplace flag. Please"
+                " choose whether you want to create a new file or overwrite the input"
+                " file."
+            )
+
```
BenjaminBossan, line 48: Basically, for the case that `output_file is None and not inplace`, further below, you print a message that the file can be updated if the protocol is different. So basically you added a new functionality, which is to check if a file needs an update. This is fine for me, I see how it can be useful, but it should be documented, otherwise I think users will not discover it ("To check if a file needs to be updated, ...").

I would also prefer to reorder this function a bit, with all argument checks coming at the top and then using early returns to get rid of some of the nesting.

So basically:

1. Check if both inplace and output file (so the same as right now)
2. If protocol is identical, log message and early return
3. Check if neither inplace, nor output file, check protocol, log message, early return

WDYT?

skops/cli/_update.py
```diff
@@ -0,0 +1,122 @@
...
+    parser_subgroup.add_argument(
+        "-o",
+        "--output-file",
+        help="Specify the output file name for the updated skops file.",
+        default=None,
+    )
+    parser_subgroup.add_argument(
+        "--inplace",
+        help="Update and overwrite the input file in place.",
+        action="store_true",
```
BenjaminBossan, line 90: Do we want inplace to be the default?

skops/cli/_update.py
```diff
@@ -0,0 +1,122 @@
...
+        Whether to update and overwrite the input file in place.
+
+    logger : logging.Logger, default=logging.getLogger()
+        Logger to use for logging.
+    """
+    if inplace:
+        if output_file is None:
+            output_file = input_file
+        else:
+            raise ValueError(
```
BenjaminBossan, line 43: AFAICT, there is no test case for this, right? Could you please add one? Not sure why codecov didn't pick this up...

BenjaminBossan (changes requested):
Thanks a lot, this is getting close to being finished.

I have a few comments and suggestions, please take a look.

skops/cli/_update.py
adrinjalali: hehe, I think it's fair to not support downgrade at least for now?

BenjaminBossan: Hmm, not sure, I could imagine edge cases (like if we mess up an upgrade of the protocol) where users might want to downgrade. Probably rare (we never mess up), but should it be impossible?

adrinjalali: I just mean to keep it simple to start with, and add that usecase if we need to.

BenjaminBossan: I'd be okay with either solution, but it should be clear to the user what's happening. Treating upgrade and downgrade identically is bound to create confusion.

docs/persistence.rst
EdAbati: Thank you for pointing this out! I forgot to check the rendered docs, changed it now :)

skops/cli/_update.py
EdAbati: This is already tested actually. I have changed the test function name to make it a bit clearer: https://github.com/skops-dev/skops/pull/343/commits/7846077e3cf9448d2eb2a1147f37bbe3776b72bb

BenjaminBossan (changes requested):
Thanks for the fixes. I missed the test for the error, my bad. There is a merge conflict, apart from that, the PR is ready to be merged.

skops/cli/_update.py
EdAbati: You are right, I was also not a fan of the if-else nesting, thank you for the tip. Let me know if this is what you had in mind

skops/cli/_update.py
EdAbati: I like the idea that `skops update` doesn't do anything (only logs) if someone doesn't specify neither `inplace` nor `output-file`. But I can change it up to you :)

skops/cli/_update.py
EdAbati: I have added a if-statement to check for a _higher_ protocol and exit with a message if true. WDYT?

skops/cli/_update.py
EdAbati: You are right, I was also not a fan of the if-else nesting, thank you for the tip. Let me know if this is what you had in mind

EdAbati:
Hi both so sorry for the delay, I have been very busy lately. 🥲

I checked all the remaining comments and fixed the conflicts. Let me know if anything else should be updated.

skops/cli/_update.py
```diff
@@ -0,0 +1,131 @@
...
+
+    if output_file is None:
+        logger.info(
+            f"File can be updated to the current protocol: {PROTOCOL}. Please"
+            " specify an output file path or use the `inplace` flag to create the"
+            " updated Skops file."
+        )
+        return None
+
+    dump(input_model, output_file)
```
BenjaminBossan, line 75: This is an extreme edge case, but I wanted to bring it up: If the user cancels the process during the `dump`ing operation, could they end up with a broken file? Since we write to the zip archive progressively, I think it could happen. Sometimes, I would start a command and see I did a mistake and would cancel it, so it's not impossible.

Now when the user uses inplace, they could even bust their original file, ending up with no working version. It is an extreme edge case for this to happen, not sure if we want to take measures against it. @adrinjalali WDYT?

To prevent this behavior, we could write the output to a temp file and then move it to `output_file`.

skops/cli/_update.py
BenjaminBossan: That behavior would be fine. It would also be acceptable IMO to update inplace by default, since it would only be done if the current version is higher, but leaving it like this is also good.

Just one question: Right now, if using the defaults (no change to log level), a user would typically not get feedback at all, right? Do we want the default behavior to be not do anything, nor message anything?

BenjaminBossan (commented):
This looks good, thanks a lot for all your work. I don't have any blockers for merging, just two comments, please take a look. If the general thought is that they don't require changes, I'm fine with you merging @adrinjalali

skops/cli/_update.py
EdAbati: > Just one question: Right now, if using the defaults (no change to log level), a user would typically not get feedback at all, right? Do we want the default behavior to be not do anything, nor message anything?

Ah yes you are right. Because the messages are logged as "info" and the default logging level should be "warning". They will not show up if one doesn't set `-v`.

Actually we could log certain messages as "warning", not sure if "info" is the correct level for somenthing like [this](https://github.com/skops-dev/skops/pull/343/files#diff-fd06e712870d44acb9dcfbfac570deb5224468b6e9d10277a9c68f4276859d54R68) for example.

BenjaminBossan: Yeah, maybe it's better to elevate that message to "warning". As a user, if I call `skops update mymodel.skops` and there is no message whatsoever, I'd assume that the model was successfully updated (if an update was necessary).

EdAbati: Cool I changed some levels here: https://github.com/skops-dev/skops/pull/343/commits/b44dca493e64debdff6767fa6bec74617bebb0dd

skops/cli/_update.py
EdAbati: Hi both :) is this what you had in mind? https://github.com/skops-dev/skops/pull/343/commits/3a30721bbdbcc72fa17abc36624c580c51a49943

skops/cli/_update.py
```diff
@@ -0,0 +1,136 @@
...
+            " updated Skops file."
+        )
+        return None
+
+    tmp_output_file = f"{output_file}.tmp"
+    logger.debug(f"Writing updated skops file to temporary path: {tmp_output_file}")
+    dump(input_model, tmp_output_file)
+    logger.debug(f"Moving updated skops file to output path: {output_file}")
+    os.replace(tmp_output_file, output_file)
+    logger.info(f"Updated skops file written to {output_file}")
```
adrinjalali, line 81: we probably don't need so much debug logging.

This should use a temp file from (https://docs.python.org/3/library/tempfile.html), and in a context manager to protect from random OS failures. Otherwise LGTM.

EdAbati: Changed it now. I used `tempfile.TemporaryDirectory` because I understood that `tempfile.NamedTemporaryFile` that cannot be moved when opened on Windows (_"Whether the name can be used to open the file a second time, while the named temporary file is still open, varies across platforms (it can be so used on Unix; it cannot on Windows)."_).

Am I overcomplicating this? 🤔

adrinjalali:
The issues on the CI are numpy / persistence related. Merging.

adrinjalali:
@BenjaminBossan somehow here it shows you've requested changes

BenjaminBossan (approved):
Thanks for all the changes, this LGTM now.

BenjaminBossan:
@EdAbati Could you please rebase on/merge with `main` to ensure the tests pass (#373 should have fixed the issues)? It's probably fine now, I just want to be sure.

EdAbati:
Done now! 🤞
Thank you for the feedback and apologies for taking a bit long to update the PR :)

EdAbati:
There is still a small issue with SciPy, that it shouldn't be caused by this PR

EdAbati:
Made a separate PR in an attempt to fix the issues with the `test_persist.py` https://github.com/skops-dev/skops/pull/374

PR #351, opened 2023-04-24
Add trust check to persistence performance test

BenjaminBossan:
## Description

Our script to check that skops persistence is not too slow compared to pickle used `trusted=True` so far. However, this is not the recommended way of using skops -- ideally the check for trusted types should be performed, which theoretically could slow down persistence. This PR adds the check for trusted types to the script.

## Results

Testing locally, there is only a small difference. For instance, for the slowest couple of tests, the added time was only 0.05 sec or less. On average across all estimators, the difference was 0.004 sec. or 30% slower. This relatively small increase means that the check for trusted types is fast compared to constructing the objects.


BenjaminBossan:

PR #352, opened 2023-04-24
FIX: bug when visualizing byte nodes

BenjaminBossan:
The `visualize` function failed when trying to visualize models that had byte attributes. This is because the node's children would contain raw bytes, which the function doesn't know how to visualize. Therefore, children of byte and byte array nodes are now skipped (in addition to the node types already being skipped).

To test this bug, I added a visualization test to the external packages like LightGBM, which make use of bytes. I'm not sure if some of the sklearn estimators could also be candidates, but it would surely be overkill to test visualizing all of them, whereas the overhead is not so big for the external packages.

### DIFF
--- a/skops/io/_visualize.py
+++ b/skops/io/_visualize.py
@@ -8,10 +8,20 @@ from typing import Any, Callable, Iterator, Literal, Sequence
 from zipfile import ZipFile
 
 from ._audit import VALID_NODE_CHILD_TYPES, Node, get_tree
-from ._general import FunctionNode, JsonNode, ListNode
+from ._general import BytearrayNode, BytesNode, FunctionNode, JsonNode, ListNode
 from ._numpy import NdArrayNode
 from ._scipy import SparseMatrixNode
 from ._utils import LoadContext
+
+# The children of these types are not visualized
+SKIPPED_TYPES = (
+    BytearrayNode,
+    BytesNode,
+    FunctionNode,
+    JsonNode,
+    NdArrayNode,
+    SparseMatrixNode,
+)
 
 
 @dataclass
@@ -269,7 +279,7 @@ def walk_tree(
     # TODO: For better security, we should check the schema if we return early,
     # otherwise something nefarious could be hidden inside (however, if there
     # is, the node should be marked as unsafe)
-    if isinstance(node, (NdArrayNode, SparseMatrixNode, FunctionNode, JsonNode)):
+    if isinstance(node, SKIPPED_TYPES):
         return
 
     yield from walk_tree(
--- a/skops/io/tests/test_external.py
+++ b/skops/io/tests/test_external.py
@@ -54,6 +54,11 @@ def rank_data(clf_data):
     return X, y, group
 
 
+def _null(*args, **kwargs):
+    # used to prevent printing anything to stdout when calling visualize
+    return
+
+
 class TestLightGBM:
     """Tests for LGBMClassifier, LGBMRegressor, LGBMRanker"""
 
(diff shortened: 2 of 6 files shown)


BenjaminBossan:

I had to set the `PYTHONIOENCODING` env var for Windows tests to pass, no idea why it's necessary all of a sudden.

skops/io/_visualize.py
```diff
@@ -8,11 +8,21 @@
...
-from ._general import FunctionNode, JsonNode, ListNode
+from ._general import BytearrayNode, BytesNode, FunctionNode, JsonNode, ListNode
 from ._numpy import NdArrayNode
 from ._scipy import SparseMatrixNode
 from ._utils import LoadContext
 
+# The children of these types are not visualized
+SKIPPED_TYPES = (
+    BytearrayNode,
+    BytesNode,
```
adrinjalali, line 19: shouldn't these two have a custom repr so that the user sees something somewhat interesting? like the first few bytes in base64 or something?

BenjaminBossan:
@adrinjalali I addressed your comment, please review again

skops/io/tests/test_external.py
```diff
@@ -303,9 +338,12 @@ def test_regressor(self, catboost, cb_regr_data, trusted, boosting_type):
 
         X, y = cb_regr_data
         estimator.fit(X, y, cat_features=[0, 1])
-        loaded = loads(dumps(estimator), trusted=trusted)
+        dumped = dumps(estimator)
+        loaded = loads(dumped, trusted=trusted)
         assert_method_outputs_equal(estimator, loaded, X)
 
+        visualize(dumped, trusted=trusted)
```
adrinjalali, line 345: these are going to print a bunch of stuff in the terminal when we run tests, shouldn't we be capturing the output? It also makes me realize, maybe we want to have an output arg for `visualize`?

BenjaminBossan: > these are going to print a bunch of stuff in the terminal when we run tests

pytest automatically captures stdout unless used with the `-s` option, no?

> maybe we want to have an output arg for `visualize`?

What would that argument do? We have the `sink` argument, in case someone wants to push the output to a logger, for instance.

adrinjalali: `sink` is not trivial to implement, it'd be nice if one could pass something like `os.devnull` and the output would be forwarded there.

I know by default pytest captures output, but even when we pass `-s` right now, there's almost no output, and this adds quite a bit there. If you think it's not worth fixing, I'm happy to merge as is.

BenjaminBossan: > `sink` is not trivial to implement, it'd be nice if one could pass something like `os.devnull` and the output would be forwarded there.

You should still be able to change `sys.stdout`, right?

> I know by default pytest captures output, but even when we pass -s right now, there's almost no output, and this adds quite a bit there.

I see. For `test_external.py`, I made a change to not print anything.

BenjaminBossan: CI is green, please review again

skops/io/tests/test_external.py
```diff
@@ -46,6 +54,11 @@ def rank_data(clf_data):
     return X, y, group
 
 
+def _null(*args, **kwargs):
+    # used to prevent printing anything to stdout when calling visualize
```
adrinjalali, line 58: passing this as `sink` is disabling a fair amount of code, not just printing. Are you sure you want this?

BenjaminBossan: This is true. The relevant part of the code is executed, though, i.e. the tests would fail without the bugfix. If you want me to still change it, I can think of something (I guess auto fixture that overrides `sys.stdout`, not sure if it conflicts with pytest).

adrinjalali: yes, but you're also specifically testing this bug anyway. So it'd be nice to have `visualize` to actually test the whole thing, in case later we add something which would break things.

BenjaminBossan: Okay, makes sense, done.

skops/io/tests/test_external.py
```diff
@@ -136,6 +163,14 @@ class TestXGBoost:
 
     """
 
+    @pytest.fixture(autouse=True)
```
adrinjalali, line 166: TIL

adrinjalali (approved):

PR #354, opened 2023-05-04
FIX: Broken CI for MacOS and Python 3.8

BenjaminBossan:
The job fails when trying to install catboost. The issue doesn't seem to be with catboost, though, but with how Python 3.8 detects the MacOS version inside GitHub runners.

If there is a better solution than just skipping, let me know.


adrinjalali (approved):
what's going on with GH actions 🤦🏼

PR #366, opened 2023-06-07
MNT handle deprecation in publish action

adrinjalali:
We seem to need to change `repository_url` to `repository-url`

failed run: https://github.com/skops-dev/skops/actions/runs/5197942965

Message:
```
Input 'repository_url' has been deprecated with message: The inputs have been normalized to use kebab-case. Use `repository-url` instead.
```


BenjaminBossan (approved):
Uh, that's annoying, thx for the fix.

PR #367, opened 2023-06-07
MNT fix issues in README and setup.py for pypi

adrinjalali:
Fixing these issues while uploading to `pypi`: 

```
 ERROR    `long_description` has syntax errors in markup and would not be        
         rendered on PyPI.                                                      
         line 30: Warning: Duplicate explicit target name: "here".              
WARNING  `long_description_content_type` missing. defaulting to `text/x-rst`. 
```

This makes the links in README.rst _anonymous_ with two underscores instead of one.


BenjaminBossan (approved):
Good catch, I'm always learning something now about rst.

PR #370, opened 2023-06-07
MNT include all packages via find_packages

adrinjalali:
we somehow forgot to include all packages in `packages` attribute. This adds them via `find_packages`.

We still get warnings for tests, but those we can fix later.

### DIFF
--- a/.github/workflows/persistence-performance.yml
+++ b/.github/workflows/persistence-performance.yml
@@ -1,6 +1,7 @@
 name: Test performance and file size of skops persistence
 
 on:
+  pull_request:
   schedule:
     - cron:  '0 9 * * 0'  # every sunday at 9:00 UTC
   workflow_dispatch:
--- a/.github/workflows/persistence-performance.yml
+++ b/.github/workflows/persistence-performance.yml
@@ -1,7 +1,6 @@
 name: Test performance and file size of skops persistence
 
 on:
-  pull_request:
   schedule:
     - cron:  '0 9 * * 0'  # every sunday at 9:00 UTC
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


adrinjalali:
This is dirty, but merging to get the release out for `0.7.post1`, and then tomorrow we can find a better solution.

adrinjalali:
Actually, I uploaded a new version with basically this change, we can clean this up before merging.

adrinjalali:
So I checked a few other packages, and it's normal to list all packages we want to include, so most people use `find_packages`. I'm not sure why two are missing from that list.

And yes, `skops.card` was included due to the entry in MANIFEST.in

I'm gonna read a bit more and try to make it cleaner.

BenjaminBossan:
I wonder if there are some easy measures we can take to prevent such problems with the release package in the future. I can imagine that if we add new sub-packages, it is easy to forget to add them in `setup.py`.

One possibility could be for the GH action, after publishing the package, to install it directly from pypi and to run a few tests (`from skorch import *`?) to ensure everything works. That way, we may catch this type of error when we do the test deploy, but it also seems cumbersome to do this. Any ideas?

adrinjalali:
> One possibility could be for the GH action, after publishing the package, to install it directly from pypi and to run a few tests (`from skorch import ...`?) to ensure everything works. That way, we may catch this type of error when we do the test deploy, but it also seems cumbersome to do this. Any ideas?

I think that's actually a good idea. Would you mind adding that step? I think now with this change, all tests and even the persistence checks should run.

BenjaminBossan:
Ah, nice solution with the wildcard, this should hopefully mean that we can add more sub-packages in the future without having to adjust the installation step.

PR #372, opened 2023-06-20
ENH add HTML repr in notebooks for Card

adrinjalali:
This makes for a nice visual in our examples and for people who use notebooks to work with the `Card` object.

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -13,6 +13,8 @@ v0.8
 ----
 - Adds the abillity to set the :attr:`.Section.folded` property when using :meth:`.Card.add`.
   :pr:`361` by :user:`Thomas Lazarus <lazarust>`.
+- Add HTML representation of the :class:`skops.card.Card` instances in Jupyter
+  notebooks.
 
 v0.7
 ----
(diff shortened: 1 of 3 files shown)


adrinjalali:
results can be seen here: https://skops--372.org.readthedocs.build/en/372/auto_examples/plot_model_card.html#sphx-glr-auto-examples-plot-model-card-py

two issues:
- it doesn't look great, but we could maybe improve that
- the metadata at the beginning of the file is annoying, we should have an easy way to remove that and get an actual model card

adrinjalali:
WDYT @merveenoyan @BenjaminBossan

BenjaminBossan:
Hmm, I'm ambivalent about this change. I totally see how it can be useful in notebooks. The way it's rendered in the docs was very confusing to me though. Apart from the two issues that you mentioned, I found it hard to distinguish what is part of the actual docs and what is part of the model card.

I wonder if, for the docs, we should keep the default behavior. Maybe rendering the markdown could be done with a different method, not by `__repr__`? This way, it could still be done if explicitly desired.

adrinjalali:
well, sphinx-gallery only understands `_repr_html_` and not `_repr_markdown_`, so removing `_repr_html_` would make the documentation look like before.

I agree with not being able to distinguish the notebook from the rendered modelcard though. I wonder if we could do something there to fix that.

BenjaminBossan:
> well, sphinx-gallery only understands `_repr_html_` and not `_repr_markdown_`, so removing `_repr_html_` would make the documentation look like before.

So do we want to go with this first? I.e. only have the nice repr in jupyter notebooks and leave the docs as they were before? If we can come up with a way to fix the identified issues of how docs are rendered, we can add them later.

adrinjalali:
Since this came up in some recent discussions, WDYT @glemaitre

PR #373, opened 2023-06-22
MNT support array functions in numpy>=1.25

adrinjalali:
Persistence is broken since from `numpy=1.25.0` `np.mean` is now dispatched via `_ArrayFunctionDispatcher`.

This fixes the issue.

Also opened an issue here: https://github.com/numpy/numpy/issues/24019

We should do a quick minor release for this probably.

### DIFF
--- a/skops/io/_numpy.py
+++ b/skops/io/_numpy.py
@@ -256,6 +256,17 @@ GET_STATE_DISPATCH_FUNCTIONS = [
     (np.random.RandomState, random_state_get_state),
     (np.random.Generator, random_generator_get_state),
 ]
+
+try:
+    # From numpy=1.25.0 dispatching for `__array_function__` is done via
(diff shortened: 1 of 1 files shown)


adrinjalali:
cc @skops-dev/maintainers

lazarust (approved):
I can't merge PRs but this looks good to me!

BenjaminBossan (approved):
Ugh, this is unfortunate.

It could very well happen sooner or later that we cannot rely on `singledispatch` anymore because of stuff like this. Hopefully I'm wrong but I wouldn't like to write a custom dispatcher (again).

BenjaminBossan:
> The real dispatch here would be `inspect.issubroutine` it seems, but we can't do that with `singledispatch`.

I would suggest that as long as it only affects this one function, we can live with the workaround; if there are more of exceptions needed in the future, we might have to roll our own dispatcher that can optionally use `issubroutine`.

PR #374, opened 2023-06-26
MNT fixed test for QuantileRegressor

EdAbati:

Some tests are failing because the default `solver="interior-point"` of `QuantileRegressor` is not available in `scipy>=1.11.0`. Please see [these tests](https://github.com/skops-dev/skops/actions/runs/5377399035/jobs/9755815655?pr=343) as an example.
`scikit-learn>=1.4.0` will change the default. In the meantime, this PR should fix the issue .

### DIFF
--- a/skops/io/tests/test_persist.py
+++ b/skops/io/tests/test_persist.py
@@ -49,6 +49,7 @@ from sklearn.utils.estimator_checks import (
     _enforce_estimator_tags_y,
     _get_check_estimator_ids,
 )
+from sklearn.utils.fixes import parse_version, sp_version
 
 import skops
 from skops.io import dump, dumps, get_untrusted_types, load, loads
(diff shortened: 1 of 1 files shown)


skops/io/tests/test_persist.py
```diff
@@ -136,7 +137,16 @@ def _tested_estimators(type_filter=None):
                     category=SkipTestWarning,
                     message="Can't instantiate estimator",
                 )
-                estimator = _construct_instance(Estimator)
+                if name == "QuantileRegressor" and sp_version >= parse_version(
+                    "1.11.0"
+                ):
+                    # The solver "interior-point" (the default solver in
+                    # scikit-learn < 1.4.0) is not available scipy >= 1.11.0. The
```
BenjaminBossan, line 144:
```suggestion
                    # scikit-learn < 1.4.0) is not available in scipy >= 1.11.0. The
```

BenjaminBossan (approved):
Besides a small typo, this LGTM. Thanks a lot for your initiative Edoardo.

The failing tests seem to be unrelated, so this can be merged from my POV.

What I do wonder though: Does this mean that using the _default solver_ for `QuantileRegressor` raises an error? That seems pretty bad. Ping @adrinjalali

BenjaminBossan:
The error that occurs is very strange, as it seems like an outdated sklearn version is being used. I could reproduce the error locally yesterday, but cannot do so today (new sklearn nightly release since then). I therefore assume the error is gone and it's a caching issue on GH or something. Anyway, the error is not related to this PR so it can be merged.

PR #375, opened 2023-06-27
MNT: Update repo URL of sklearn nightly build, fix sparse matrix persistence issue

BenjaminBossan:
See [this comment](https://github.com/scikit-learn/scikit-learn/issues/25273#issuecomment-1609430993).

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -63,7 +63,7 @@ jobs:
         pip install black=="22.6.0" isort=="5.10.1" mypy=="1.0.0"
         pip uninstall --yes scikit-learn
         if [ ${{ matrix.sklearn_version }} == "nightly" ];
-          then pip install --pre --extra-index https://pypi.anaconda.org/scipy-wheels-nightly/simple scikit-learn;
+          then pip install --pre --extra-index https://pypi.anaconda.org/scientific-python-nightly-wheels/simple scikit-learn;
           else pip install "scikit-learn~=${{ matrix.sklearn_version }}";
         fi
         if [ ${{ matrix.os }} == "ubuntu-latest" ];
--- a/skops/io/tests/_utils.py
+++ b/skops/io/tests/_utils.py
@@ -133,10 +133,25 @@ def _assert_vals_equal(val1, val2):
         _assert_generic_objects_equal(val1, val2)
 
 
+def _clean_params(params):
+    # this function deals with cleaning special parameters that for one reason
+    # or another should be removed or modified.
+    params = params.copy()
+
+    # see #375
+    keys_to_remove = ["_has_canonical_format", "_has_sorted_indices"]
+    for key in keys_to_remove:
+        params.pop(key, None)
+
+    return params
+
+
 def assert_params_equal(params1, params2):
     # helper function to compare estimator dictionaries of parameters
     if params1 is None and params2 is None:
         return
+
+    params1, params2 = _clean_params(params1), _clean_params(params2)
     assert len(params1) == len(params2)
     assert set(params1.keys()) == set(params2.keys())
     for key in params1:
--- a/skops/io/tests/test_persist.py
+++ b/skops/io/tests/test_persist.py
@@ -1047,3 +1047,21 @@ def test_compression_level():
     dumped_compressed = dumps(model, compression=ZIP_DEFLATED, compresslevel=9)
     # This reduces the size substantially
     assert len(dumped_raw) > 5 * len(dumped_compressed)
+
+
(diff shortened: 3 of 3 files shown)


BenjaminBossan:
Hmm, I'm really stumped about this error. First of all, even though pickle does not have the problem, saving and loading with scipy (which skops uses under the hood) has the same issue:

```python
>>> from scipy.sparse import load_npz, save_npz
>>> from scipy import sparse
>>> x = sparse.csr_matrix((3, 4))
>>> x.has_canonical_format
True
>>> x._has_canonical_format
True
>>> save_npz('/tmp/sparse_matrix.npz', x)
>>> y = load_npz('/tmp/sparse_matrix.npz')
>>> y._has_canonical_format
AttributeError: 'csr_matrix' object has no attribute '_has_canonical_format'
```

So in a way it's already broken in scipy, but probably it's not considered a problem there, since it's a private attribute.

What surprises me, however, is that the issue only occurs now. The change in scipy is already [10 years old](https://github.com/scipy/scipy/blame/899f4efbcd8020b538bd97eb54353b3249af6e43/scipy/sparse/_compressed.py#L1083), so this is not caused by a new scipy version. So then I thought that there was a change in sklearn recently, but the corresponding attribute was also last touched a long time ago ([5 years](https://github.com/scikit-learn/scikit-learn/blame/98f21a928c2820268477751afc36e8a174f60441/sklearn/feature_extraction/text.py#L1677)), therefore, this is also not an explanation.

The most likely explanation has thus to be some change in [`scipy.sparse.diags`](https://github.com/scipy/scipy/blame/2e16ec5bc62b7b7db21b37b838fc771c7d232006/scipy/sparse/_construct.py#L73-L196), which is what is being called by sklearn and has been changed in the last month. I suspect that somewhere inside that function, the attribute is indirectly created, but I cannot pinpoint out where and what recent change affected this.

Anyway, I'll push a solution that will ensure that the private attributes are present on the loaded object iff they were present on the original object. This is achieved by remembering if the attribute was there when storing the state. If it was, when loading the object, `loaded.has_canonical_format` is accessed, which, as a side effect, creates the private attributes. This is a somewhat ugly solution, but I think it is the "most correct" solution short of fixing the problem inside of scipy.

BenjaminBossan:
Okay, CI is green, ready for review @skops-dev/maintainers. Until this, or a different fix, is merged, CI will be red for everyone :/

adrinjalali:
This to me is similar to the flags issue we had with numpy, i.e. it doesn't affect the actual values in the matrix, and only gives an information about how it's stored.

I think we can safely just ignore this attribute instead of trying to save it.

BenjaminBossan:
So the suggestion would be to add a special case to the tests to explicitly remove those two arguments?

adrinjalali:
yeah I'd just remove them.

BenjaminBossan:
@adrinjalali pls take another look.

adrinjalali (approved):
Nice!

jot-s-bindra:
Its not related to this commit
I just wanted your help in setting up environment in local machine
```
git clone git@github.com:jot-s-bindra/skops.git
git init
git checkout -b new-feature
conda create -c conda-forge -n skops python=3.10
conda activate skops
python -m pip install -e ".[tests,docs]"
conda install -c conda-forge pre-commit
pre-commit install
pytest
```
during pytest i am recieving variable declare errors , modules not installed errors like
ERROR examples/plot_california_housing.py - KeyError: 'HF_HUB_TOKEN'
ERROR examples/plot_hf_hub.py - KeyError: 'HF_HUB_TOKEN'
ERROR examples/plot_intelex.py - KeyError: 'HF_HUB_TOKEN'
ERROR scripts/clean_skops.py - OSError: pytest: reading from stdin while output is captured!  Consider using `-s`.
ERROR spaces/skops_model_card_creator/app.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/create.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/edit.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/gethelp.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/start.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/tasks.py - ModuleNotFoundError: No module named 'streamlit'

adrinjalali:
In order to run hugging face related commands, you need to have your token present in `HF_HUB_TOKEN` environment variable.

The `spaces` folder is not for most users. If you want to run it locally, you need to have its dependencies including `streamlit` installed.

PR #377, opened 2023-06-28
Prepare release of v0.8

BenjaminBossan:
- Update `changes.rst`
- Bump the version


BenjaminBossan:
Being dirty here and merging without review. This is because Adrin is off until next week and we want to get the release out ASAP because of #373.

PR #389, opened 2023-09-08
[don't merge] Test hfh v0.17.0.rc0

BenjaminBossan:
For testing the new release, don't merge


BenjaminBossan:
Looks good. Some tests are failing but for unrelated reasons.

@adrinjalali there is a FutureWarning coming from fairlearn. Is that something that will be fixed in fairlearn or should we add an exception for it in skops?

PR #392, opened 2023-10-06
MNT: updates to HF Hub, manage fairlearn deprecation

BenjaminBossan:
Some maintenance work to make skops future proof:

1. Ignore `FutureWarning` stemming from fairlearn: New pandas versions give `FutureWarning` for applymap. We have to wait for fairlearn to update this.
2. Make the widget attribute an array: Starting from HF Hub 0.18, it has to be an array.
3. InferenceApi is deprecated move to InferenceClient: API changed a little bit, had to be adjusted.

Note: DON'T MERGE THIS YET. The min version of HF Hub has been increased to 0.18 to test the changes but it actually should be lowered to 0.17 once CI is green for 0.18.

### DIFF
--- a/skops/_min_dependencies.py
+++ b/skops/_min_dependencies.py
@@ -13,7 +13,7 @@ PYTEST_MIN_VERSION = "5.0.1"
 dependent_packages = {
     "scikit-learn": ("0.24", "install", None),
     "scikit-learn-intelex": ("2021.7.1", "docs", None),
-    "huggingface_hub": ("0.10.1", "install", None),
+    "huggingface_hub": ("0.18.0.rc0", "install", None),
     "tabulate": ("0.8.8", "install", None),
     "quantile-forest": ("1.0.0", "tests", None),
     "pytest": (PYTEST_MIN_VERSION, "tests", None),
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -13,7 +13,7 @@ from pathlib import Path
 from typing import Any, List, Literal, MutableMapping, Optional, Sequence, Union
 
 import numpy as np
-from huggingface_hub import HfApi, InferenceApi, snapshot_download
+from huggingface_hub import HfApi, InferenceClient, snapshot_download
 from sklearn.utils import check_array
 
 SUPPORTED_TASKS = [
@@ -755,9 +755,9 @@ def get_model_output(repo_id: str, data: Any, token: Optional[str] = None) -> An
         inputs = {f"x{i}": data[:, i] for i in range(data.shape[1])}
         inputs = {"data": inputs}
 
-    res = InferenceApi(repo_id=repo_id, task=model_info.pipeline_tag, token=token)(
-        inputs=inputs
-    )
+    client = InferenceClient(token=token)
+    res_bytes = client.post(json={"inputs": inputs}, model=repo_id)
+    res = json.loads(res_bytes.decode("utf-8"))
 
     if isinstance(res, list):
         return np.array(res)
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -18,7 +18,9 @@ filterwarnings = [
     # https://github.com/scikit-learn/scikit-learn/pull/23633
     "ignore:Unlike other reduction functions:FutureWarning",
     # https://github.com/scikit-learn/scikit-learn/pull/25157
-    "ignore:\\w+ is deprecated. Use files\\(\\) instead:DeprecationWarning"
+    "ignore:\\w+ is deprecated. Use files\\(\\) instead:DeprecationWarning",
+    # comes from fairlearn
+    "ignore:DataFrame.applymap has been deprecated. Use DataFrame.map instead:FutureWarning",
 ]
 markers = [
     "network: marks tests as requiring internet (deselect with '-m \"not network\"')",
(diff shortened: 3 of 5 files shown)


BenjaminBossan:
@adrinjalali This should be ready to review. CI is partly red, but that just seems to be due to rate limit. I'll restart the tests later.

skops/hub_utils/_hf_hub.py
```diff
@@ -755,9 +755,9 @@ def get_model_output(repo_id: str, data: Any, token: Optional[str] = None) -> An
         inputs = {f"x{i}": data[:, i] for i in range(data.shape[1])}
         inputs = {"data": inputs}
 
-    res = InferenceApi(repo_id=repo_id, task=model_info.pipeline_tag, token=token)(
-        inputs=inputs
-    )
+    client = InferenceClient(token=token)
+    res_bytes = client.post(json={"inputs": inputs}, model=repo_id)
+    res = json.loads(res_bytes.decode("utf-8"))
```
adrinjalali, line 760: don't we have a test for this?

adrinjalali (commented):
Otherwise LGTM.

skops/hub_utils/_hf_hub.py
BenjaminBossan: Yes, I forgot that we require `[CI inference]` in the commit message to trigger it. This is done now but of course we run into the rate limit again...

adrinjalali: Should we remove the functionality then since the rate limit is so low?

BenjaminBossan: It's mostly an issue for our CI, right? Which is probably why we made the test opt-in in the first place. I'd say we leave it, in case some users actually use the feature. If we want to remove it later on, we can do it properly with a deprecation cycle.

BenjaminBossan: We can also add a random sleep between 1 and 100 hours to avoid hitting the limit ;-P

adrinjalali: Hmm, not happy with a functionality that even our CI hits the limits pretty much every time. But we can merge and figure that out in a separate PR.

skops/_min_dependencies.py
```diff
@@ -13,7 +13,7 @@
 dependent_packages = {
     "scikit-learn": ("0.24", "install", None),
     "scikit-learn-intelex": ("2021.7.1", "docs", None),
-    "huggingface_hub": ("0.10.1", "install", None),
+    "huggingface_hub": ("0.18.0.rc0", "install", None),
```
adrinjalali, line 16: should this change before merge?

BenjaminBossan: Yes, the actual min version is 0.17. I can change it now or we can first ensure that all tests pass for 0.18 too -- which may take some time ;-)

BenjaminBossan: Okay, I'll let you know in a month or two.

adrinjalali: Is it really that bad? So it's not something we can maintain. Let's fix the version merge here, then we deal with the deprecation

BenjaminBossan: Yeah, the error now says that the daily quota has been exceeded. I don't remember it being this tight, not sure if it was reduced. We may be able to avoid it by using a HF employee token, but then we have the old issue of hiding it while still allowing it to be used in PRs, right?

adrinjalali: At the end of the day, it's not a free service, and it doesn't match the vibe of this library anymore I would say. We added it because everybody could simply use it, but with the new rate limits, I don't see why we would keep supporting it.

BenjaminBossan: I agree, the rate limit is pretty aggressive (still couldn't make all tests pass), so for practical purposes, it's not a free service anymore. I think we have a few options, I'll leave the decision up to you:

1. Still merge this PR, as I'm pretty sure it works (after adjusting the min version of course). Then create a new PR to deprecate the feature.
2. Don't merge this PR. Deprecate the feature.
3. Remove without deprecation.

adrinjalali: I'd be happy to merge this PR, then I'll open another PR to deprecate the feature.

jot-s-bindra:
Its not related to this commit
I just wanted your help in setting up environment in local machine
```
git clone git@github.com:jot-s-bindra/skops.git
git init
git checkout -b new-feature
conda create -c conda-forge -n skops python=3.10
conda activate skops
python -m pip install -e ".[tests,docs]"
conda install -c conda-forge pre-commit
pre-commit install
pytest
```
during pytest i am recieving variable declare errors , modules not installed errors like
ERROR examples/plot_california_housing.py - KeyError: 'HF_HUB_TOKEN'
ERROR examples/plot_hf_hub.py - KeyError: 'HF_HUB_TOKEN'
ERROR examples/plot_intelex.py - KeyError: 'HF_HUB_TOKEN'
ERROR scripts/clean_skops.py - OSError: pytest: reading from stdin while output is captured!  Consider using `-s`.
ERROR spaces/skops_model_card_creator/app.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/create.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/edit.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/gethelp.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/start.py - ModuleNotFoundError: No module named 'streamlit'
ERROR spaces/skops_model_card_creator/tasks.py - ModuleNotFoundError: No module named 'streamlit'

PR #394, opened 2023-10-10
Gets SciKeras script working

### LINKED ISSUE #388: Bug: dump fails on keras models with RecursionError: maximum recursion depth exceeded, after calling fit
skops io seems to work fine with scikeras models when they haven't been fit, but not when they've been fit. The error message (and potentially underlying cause) is the same as https://github.com/skops-dev/skops/issues/184

Example code (this works):
```
from tensorflow import keras
from sklearn.pipeline import Pipeline
from skops.io import dump
from scikeras.wrappers import KerasClassifier

# This simplifies the basic usage tutorial from https://adriangb.com/scikeras/stable/notebooks/Basic_Usage.html

def get_clf(meta):
    n_features_in_ = meta["n_features_in_"]
    model = keras.models.Sequential()
    model.add(keras.layers.Input(shape=(n_features_in_,)))
    model.add(keras.layers.Dense(1, activation="sigmoid"))
    return model
    
clf = KerasClassifier(
    model=get_clf,
    loss="binary_crossentropy"
)

pipeline = Pipeline(
    [("classifier", clf)]
)

# These both work
dump(clf, 'keras-test.skops') 
dump(pipeline, 'keras-test.skops')
```

However, running the following then gives a "RecursionError: maximum recursion depth exceeded" on the last line:

```
import numpy as np
from sklearn.datasets import make_classification
X, y = make_classification(1000, 20, n_informative=10, random_state=0)
clf.fit(X, y)
dump(clf, 'keras-test.skops') 
```

lazarust:
WIP still need to fix one test

#### Reference Issues/PRs
Fixes #388

This fixes a recursion error that was happening when dumping a scikeras model.

### DIFF
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -4,6 +4,7 @@ import io
 import json
 import operator
 import uuid
+import weakref
 from functools import partial
 from reprlib import Repr
 from types import FunctionType, MethodType
@@ -46,10 +47,12 @@ def dict_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
     for key, value in obj.items():
         if isinstance(value, property):
             continue
+        if isinstance(key, weakref.ref):
+            key = getattr(key(), "_name")
         if np.isscalar(key) and hasattr(key, "item"):
             # convert numpy value to python object
             key = key.item()  # type: ignore
-        content[key] = get_state(value, save_context)
+        content[str(key)] = get_state(value, save_context)
     res["content"] = content
     res["key_types"] = key_types
     return res
--- a/skops/io/_utils.py
+++ b/skops/io/_utils.py
@@ -157,14 +157,13 @@ def get_state(value, save_context: SaveContext) -> dict[str, Any]:
     # fails with `get_state`, we try with json.dumps, if that fails, we raise
     # the original error alongside the json error.
 
-    # TODO: This should help with fixing recursive references.
-    # if id(value) in save_context.memo:
-    #     return {
-    #         "__module__": None,
-    #         "__class__": None,
-    #         "__id__": id(value),
-    #         "__loader__": "CachedNode",
-    #     }
+    if id(value) in save_context.memo:
+        return {
+            "__module__": None,
+            "__class__": None,
+            "__id__": id(value),
+            "__loader__": "CachedNode",
+        }
 
     __id__ = save_context.memoize(obj=value)
 


lazarust:
It looks like all the pytest tests are failing with `You have exceeded our daily quotas for action: createRepo. We invite you to retry later.`

BenjaminBossan:
Should be addressed by #398

BenjaminBossan:
@lazarust Could you please solve the merge conflict? Regarding the uncovered new line: Would that be solved by adding tests for scikeras?

lazarust:
@BenjaminBossan I've fixed the merge conflict.

Yeah, I believe it would be but I was unsure if we wanted to have tests that included another library like that. Should I add one?

BenjaminBossan:
> Yeah, I believe it would be but I was unsure if we wanted to have tests that included another library like that. Should I add one?

Yes, it would be good, since that was the initial reason for the change. We have external library tests, see here:

https://github.com/skops-dev/skops/blob/main/skops/io/tests/test_external.py

For scikeras, it won't be possible to add a comprehensive coverage of all possible models, so a simple example should be enough. If it's possible to add a unit test independently of scikeras that explicitly tests weakrefs, that would also be good, not sure how easy it is to do.

Regarding the failing tests, at first glance, it appears to be a change in the model repr in the latest sklearn version, so not related to the changes here.

lazarust:
@BenjaminBossan Sorry this took me so long to get back to. I've added a test to hit that line.

skops/io/tests/test_external.py
```diff
@@ -427,3 +428,40 @@ def test_quantile_forest(self, quantile_forest, regr_data, trusted, tree_method)
...
+            model = keras.models.Sequential()
+            model.add(keras.layers.Input(shape=(n_features_in_,)))
+            model.add(keras.layers.Dense(1, activation="sigmoid"))
+            return model
+
+        clf = KerasClassifier(model=get_clf, loss="binary_crossentropy")
+
+        pipeline = Pipeline([("classifier", clf)])
+
+        dump(clf, "keras-test.skops")
```
BenjaminBossan, line 466: Instead of just dumping, could we please do a cycle of `dumps` and `loads`, similar to the other tests?

BenjaminBossan (changes requested):
Thanks so much for the added scikeras tests. I just have a small request to improve them a little bit.

BenjaminBossan:
@adrinjalali The tests for sklearn nightly are failing because the model repr was changed (not sure why). Normally, we could fix that by having the test check the sklearn verison, but this is a doctest. Any idea how it can be fixed, short of skipping it whole?

skops/io/tests/test_external.py
lazarust: Done! I now have it dumping the model, loading it back in and comparing the results.

BenjaminBossan: Can we get rid of `dump` completely in favor of `dumps`? That way, we also don't need to care about cleaning up any files created during the test.

lazarust: Fixed! Sorry about that. I didn't realize the difference between `dumps` and `dump` 🤯

skops/io/tests/test_external.py
```diff
@@ -427,3 +428,56 @@ def test_quantile_forest(self, quantile_forest, regr_data, trusted, tree_method)
...
+        clf = KerasClassifier(model=get_clf, loss="binary_crossentropy")
+
+        pipeline = Pipeline([("classifier", clf)])
+
+        dumps(clf, "keras-test.skops")
+        dumps(pipeline, "keras-test.skops")
+
+        X, y = make_classification(1000, 20, n_informative=10, random_state=0)
+        clf.fit(X, y)
+        dumped = dumps(clf, "keras-test.skops")
```
BenjaminBossan, line 480:
```suggestion
        dumped = dumps(clf)
```

2nd argument to `dumps` is the compression level. Honestly, I'm surprised that this didn't raise an error.

BenjaminBossan:
Ugh, the list of trusted modules is giant now :D I guess it's related to the tensorflow change. Could you please explain why that was necessary? Also, we now get this error on CI:

> E       AttributeError: module 'scikeras' has no attribute 'wrappers'

lazarust:
@BenjaminBossan Yeah, sorry I realized I had marked the test method as a `@pytest.fixture` 🤦🏽, so the test wasn't ever running (which is why it wasn't erroring when switching from `dump` to `dumps`). I'm hoping to get the test fixed and cleanup that list of trusted modules today.

lazarust:
@BenjaminBossan After staring at this all day, I could use some help lol. For some reason, there's some infinite recursion happening when constructing the tree that I can't figure out why.

Initially, I thought it was due to the `CachedNode`s in the tree, but the `construct` method isn't getting hit for those. I've gone through the tree and didn't see anything outrageous so if you could take a look that'd be great!

adrinjalali:
I'll have to check when I'm back. Still off till end of November. But I remember `CacheNode` caused some issues when I was working on it, partly due to small integers and other common objects having the same `id` in python.

lazarust:
@adrinjalali It looks like you're right, there are some JSONNodes that have the same `id` as some of the booleans that are getting cached. 

To me, there are a couple of solutions:
1. Find a better way to identify objects uniquely
    - Maybe we should use [uuid](https://docs.python.org/3/library/uuid.html) instead? 
3. Don't cache boolean values since these seem to have duplicate ids.
4. Include the data type in the `__id__` (so it would be something like "id/dtype")
    - This isn't the best idea since we'd have to update everywhere that uses that id.

BenjaminBossan:
> Find a better way to identify objects uniquely
> * Maybe we should use [uuid](https://docs.python.org/3/library/uuid.html) instead?

Could you explain further how that would work?

> Don't cache boolean values since these seem to have duplicate ids.

Note that booleans are not the only values that could be affected, e.g. small integers will have the same issue.

lazarust:
@BenjaminBossan On that first point, I don't think that would work anymore. I had thought `uuid` was just a more complicated `id` function but looking into the documentation it seems to be doing something differently than I understood. 

I also agree with your note, but I'm unsure of the best way forward. Do you have any thoughts?

adrinjalali:
Thinking about this, WDYT of saving TF objects using TF and saving them in our zipfile, not trusting any TF object by default, and when raising a warning about untrusted code, we also point users to this page (https://github.com/tensorflow/tensorflow/blob/master/SECURITY.md)

lazarust:
@adrinjalali That sounds good to me! After reading through the TensorFlow security documentation, showing the warning makes sense. 

Would we show the warning on `dump` or only on `load`? I would think both probably but I could see an argument to just show it on `load`.

adrinjalali:
It would only be on loading time, and by default load fails anyway, we just need to make sure when it fails, we also give them the right information and the right link.

lazarust:
@adrinjalali Sweet, so things that need to be changed:

1. When dumping a TensorFlow model save the TF object using TF and store it in the .zip file.
2. When loading a Tensorflow model raise a warning linking to the TF docs
    a. Since the model fails to load anyway, we're just changing the error message at this point, if I'm understanding things correctly.

Do I have that right? Thanks for your help on this solution.

lazarust:
@adrinjalali I am just pinging you again to see if you have any thoughts before I get started on a solution.

adrinjalali:
@lazarust yes, that sounds perfect to me.

lazarust:
@adrinjalali @BenjaminBossan This is ready for y'all to look at again! Sorry, it took a while, got a little busy around the holidays.

skops/io/_persist.py
```diff
@@ -33,12 +34,19 @@ def _save(obj: Any, compression: int, compresslevel: int | None) -> io.BytesIO:
...
-
-        state["protocol"] = save_context.protocol
-        state["_skops_version"] = skops.__version__
-        zip_file.writestr("schema.json", json.dumps(state, indent=2))
+        # If obj is scikeras model save the scikeras model via scikeras
+        if "scikeras" in obj.__class__.__module__:
+            with tempfile.NamedTemporaryFile(mode="w+", newline="") as temp_file:
+                file_name = temp_file.name + ".keras"
+                obj.model.save(file_name)
+                zip_file.write(file_name, "model.keras")
```
adrinjalali, line 42: Oh I see what you're doing here. This unfortunately doesn't work when you put a scikeras model inside a `Pipeline` for instance. This code should go in the same place similar to where we handle `sklearn` and `numpy` specific objects, and be added to our dispatch mechanism somehow. And the created dump files go in the same zipfile which includes the other objects.

adrinjalali (commented):
Sorry for the late review @lazarust , thanks for your patience and the ping.

lazarust:
@adrinjalali This is ready for you to take a look at again. The failing tests are due to this bug in scikeras https://github.com/adriangb/scikeras/issues/313, so we'll probably need to wait for that to get fixed before moving this forward

skops/io/_scikeras.py
```diff
@@ -0,0 +1,61 @@
...
+
+class SciKerasNode(Node):
+    def __init__(
+        self,
+        state: dict[str, Any],
+        load_context: LoadContext,
+        trusted: bool | Sequence[str] = False,
+    ) -> None:
+        super().__init__(state, load_context, trusted)
+        self.trusted = self._get_trusted(trusted, [KerasClassifier, KerasRegressor])
```
adrinjalali, line 41: we shouldn't be trusting any of keras objects by default for now.

adrinjalali (commented):
This looks much better.

skops/io/_scikeras.py
lazarust: Fixed 👍🏽

adrinjalali:
I think you forgot a future import and old python is complaining.

lazarust:
@adrinjalali I got it working for older versions of Python other than Windows on 3.8 (I think the Ubuntu 3.9 failure was just a codecov issue). Could you help me figure out how to get it working? Downgrading tensorflow to 2.11 just breaks more versions.

adrinjalali:
One of the main reasons we support 3.8 was colab, which now supports 3.10. I think we can drop 3.8 alltogether.

Would you be kind enough to submit a PR to remove 3.8 support?

lazarust:
@adrinjalali I created this PR to remove 3.8 support #418, but I'm running into an issue where `pip install scikit-learn-intelex` isn't finding any valid versions only on MacOS (this is happening both in the GH action and locally for me). I'm still looking into a fix

adrinjalali:
Intelex is very brittle, and I don't think it's used much if at all. I'd be in favor of removing it. It was only to handle the inference side and HF inference is not that well maintained anyway.

@BenjaminBossan any objections to remove it?

BenjaminBossan:
Works for me.

adrinjalali:
@lazarust you've been very kind here. Would you remove `sklearn-intelex` in a separate PR too?

lazarust:
@adrinjalali Done! I removed `sklearn-intelex` in #420

adrinjalali:
@lazarust let me know when the CI is green and ready for review here again.

lazarust:
@adrinjalali Sounds good, it may take me a couple of days to get to it. It looks like there's something wrong in `tensorflow` with how it's using `google.protobuf`

adrinjalali:
The errors seem like a bunch of deprecation warnings due to some updates in dependency packages maybe?

lazarust:
@adrinjalali I created an issue in tensorflow to discuss this and make sure I wasn't missing anything https://github.com/tensorflow/tensorflow/issues/68194.

It seems like the best way forward is to ignore the warnings for now until Tensorflow supports `protobuf` 5.0+. Does that sound good to you? Ignoring warnings doesn't sound like the best thing I'm just unsure of a better way forward.

adrinjalali:
Ignoring warnings in cases where we know why they're happening and with a comment as when the ignore statement should be removed is a normal practice indeed. Thanks for the follow up work on the TF side.

lazarust:
@adrinjalali This should be ready for you to look at again. The failing tests seem to just be a blip with codecov

skops/io/_scikeras.py
```diff
@@ -0,0 +1,63 @@
+from __future__ import annotations
+
+import io
+import os
+import tempfile
+from typing import Sequence, Type
+
+import tensorflow as tf
+from scikeras.wrappers import KerasClassifier, KerasRegressor
```
adrinjalali, line 9: this is making `scikeras` a hard dependency in effect, while it should be a soft dependency (we only care about it if user has it installed). Also, we shouldn't even try to import anything from `scikeras` if the user is not dealing with a scikeras model.

skops/io/_scikeras.py
```diff
@@ -0,0 +1,63 @@
...
+        "__module__": get_module(type(obj)),
+        "__loader__": "SciKerasNode",
+    }
+
+    obj_id = save_context.memoize(obj)
+    f_name = f"{obj_id}.keras"
+
+    with tempfile.TemporaryDirectory() as temp_dir:
+        file_name = os.path.join(temp_dir, "model.keras")
+        obj.model.save(file_name)
```
adrinjalali, line 28: so we only save the `model` attribute? This sounds odd.

adrinjalali (commented):
A few thoughts looking at this in more details (other than the inline comments)

- we are delegating the save / load to tensorflow? I'd like to see a test showing the user the right error when they try to load such a model without explicitly trusting including modules.
- we don't really include the inner modules in our json tree here, which means we have no idea what's inside that keras model, which means we're prone to massive exploits, this seems it beats the purpose of this format.

@lazarust you've done great work so far, let me know if you need me to have a more detailed look. I haven't personally tried to solve this project for TF, but I could have a look.

skops/io/_scikeras.py
lazarust: From my understanding of https://keras.io/guides/serialization_and_saving/, it seems that Keras is compressing all the pieces of the models into the `.keras` file. Should I change the name of the file to make it more clear?

skops/io/_scikeras.py
lazarust: Fixed this by adding a try except around the import, following how `_quantile_forest.py` handles importing

lazarust:
@adrinjalali I think I've addressed most of your comments, and apologize it took me a bit to get back to this. 

For
>we don't really include the inner modules in our json tree here, which means we have no idea what's inside that keras model, which means we're prone to massive exploits, this seems it beats the purpose of this format.

I'm a little confused by what you mean by inner modules... could you elaborate on what you mean? 

Sorry this PR has been taking so long to get ironed out!

adrinjalali:
@lazarust I went down the rabbit hole of reading the persistence code from keras, and I think it's easier if I push to this PR some changes. So I'll update this one, and you can review the work if that's okay.

lazarust:
@adrinjalali Sounds good to me, let me know if there's any thing I can do to help!

adrinjalali:
@lazarust , this is what I had in mind.

However, this has a major issue:

The user can use `keras.src.saving.saving_lib.save_model` with `weights_format="npz"`, or manually create a zip file with `npz` weights. Then The issue is we have this in tensorflow:

https://github.com/keras-team/keras/blob/d4a51168bfedf69a9aae7ddff289277972dfd85d/keras/src/saving/saving_lib.py#L1027

Which allows loading pickles through numpy. So we can only merge / support this, if we have a way to make sure there are no pickle objects in that zip file.

lazarust:
@adrinjalali Ah the way you used tf directly makes sense to me. 

As for the pickle file issue, I don't think there's a good way of doing this... Do you know if the way TF loads the files only looks for `.npz` files or does it try and load any file in the zip through numpy?

adrinjalali:
I don't think it matters which files TF finds to load, as long as it allows loading pickle files. What we can investigate to move this further, is if we can reliably monkey-patch and disable the pickle machinery while loading a `skops` file, so that in case there's a pickle load happening somewhere, we make sure we fail.

lazarust:
Yeah I agree, I guess I just thought if TF only tries to load `.npz` files, maybe `skops` could just ignore all `.npz`.

Is monkey-patching TF something we'd want to support long term? If TF changed how model saving/loading worked in a future version we'd have to make updates anyways...

adrinjalali:
It would be monkey patching `pickle`, not TF.

lazarust:
Ah, yeah I don't think that would be too bad... we'd just need to monkey-patch the `load` functionality

lazarust:
@adrinjalali Sorry I still haven't gotten to this. Have you made any progress on patching `pickles`?

adrinjalali:
No I need to get to it soon

lazarust:
@adrinjalali I just realized that this PR is still open. Would you like me to work on merging `man` into it? It looks like last we were looking at it, you were going to look into patching `pickle`

adrinjalali:
yeah I haven't had a chance into looking into how reliable patching `pickle` can be. I don't have a good feeling about it. Ideally we would fix the code in `keras` and make sure there's a way to disable using `pickle` the same way you can do for `numpy` objects.

PR #396, opened 2023-10-12
DEPR deprecate hub_utils.get_model_output

adrinjalali:
This deprecates the function, to be removed in the next released.

### DIFF
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -744,11 +744,9 @@ def get_model_output(repo_id: str, data: Any, token: Optional[str] = None) -> An
     available.
     """
     warnings.warn(
-        (
-            "This feature is no longer free on hf.co and therefore this function will"
-            " be removed in the next release. Use `huggingface_hub.InferenceClient`"
-            " instead."
-        ),
+        "This feature is no longer free on hf.co and therefore this function will"
+        " be removed in the next release. Use `huggingface_hub.InferenceClient`"
+        " instead.",
         FutureWarning,
     )
     model_info = HfApi().model_info(repo_id=repo_id, use_auth_token=token)  # type: ignore
--- a/skops/hub_utils/_hf_hub.py
+++ b/skops/hub_utils/_hf_hub.py
@@ -744,9 +744,12 @@ def get_model_output(repo_id: str, data: Any, token: Optional[str] = None) -> An
     available.
     """
     warnings.warn(
-        "This feature is no longer free on hf.co and therefore this function will"
-        " be removed in the next release. Use `huggingface_hub.InferenceClient`"
-        " instead."
+        (
+            "This feature is no longer free on hf.co and therefore this function will"
+            " be removed in the next release. Use `huggingface_hub.InferenceClient`"
+            " instead."
+        ),
+        FutureWarning,
     )
     model_info = HfApi().model_info(repo_id=repo_id, use_auth_token=token)  # type: ignore
     if not model_info.pipeline_tag:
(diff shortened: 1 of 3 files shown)


skops/hub_utils/tests/test_hf_hub.py
```diff
@@ -512,6 +513,12 @@ def test_inference(
     assert np.allclose(output, y_pred)
 
 
+def test_get_model_output_deprecated():
```
BenjaminBossan, line 516: If, in the test above, you add the `match` to `pytest.warns`, this test would be completely redundant, right? This might be a worthwhile change, so that we avoid hitting the endpoint even more often. If you want to leave this as a separate test, it should have the `@pytest.mark.network` and `@pytest.mark.inference` decorators markers, right?

skops/hub_utils/_hf_hub.py
```diff
@@ -737,8 +743,12 @@ def get_model_output(repo_id: str, data: Any, token: Optional[str] = None) -> An
     Also note that if the model repo is private, the inference API would not be
     available.
     """
-    # TODO: the "type: ignore" should eventually become unncessary when hf_hub
-    # is updated
+    warnings.warn(
+        "This feature is no longer free on hf.co and therefore this function will"
+        " be removed in the next release. Use `huggingface_hub.InferenceClient`"
+        " instead.",
+        FutureWarning,
```
BenjaminBossan, line 750: I wonder if `DeprecationWarning` is more appropriate.

skops/hub_utils/_hf_hub.py
```diff
@@ -737,8 +743,12 @@ def get_model_output(repo_id: str, data: Any, token: Optional[str] = None) -> An
     Also note that if the model repo is private, the inference API would not be
     available.
     """
-    # TODO: the "type: ignore" should eventually become unncessary when hf_hub
-    # is updated
+    warnings.warn(
+        "This feature is no longer free on hf.co and therefore this function will"
```
BenjaminBossan, line 747: Technically speaking, it's still free, just heavily rate-limited.

BenjaminBossan (commented):
Thanks for adding this.

I'm not sure why `test_push_download[pickle-False]` is failing. At first I thought another rate limit, but I ran the tests a couple of times locally and they passed. Maybe it's a Windows issue.

skops/hub_utils/tests/test_hf_hub.py
adrinjalali: Here I'm not passing any token and not passing a valid model name, so the function would fail anyway (hence the `pytest.raises(Exception)`, therefore I don't think it even needs a network marker, and it can be in all situations and not only with a specific commit message or a merker for pytest.

skops/hub_utils/_hf_hub.py
adrinjalali: So it's not free 😁

adrinjalali (commented):
> I'm not sure why test_push_download[pickle-False] is failing. At first I thought another rate limit, but I ran the tests a couple of times locally and they passed. Maybe it's a Windows issue.

Can figure that out in another PR probably :)

skops/hub_utils/_hf_hub.py
BenjaminBossan: A philosophical question...

skops/hub_utils/_hf_hub.py
BenjaminBossan: TIL. Makes sense.

skops/hub_utils/tests/test_hf_hub.py
BenjaminBossan: I see, then it's good.

BenjaminBossan (approved):

adrinjalali:
@BenjaminBossan so should we merge?

PR #397, opened 2023-10-12
MNT update linters and use ruff

adrinjalali:
This replaces flake8 and isort with ruff, which is much faster, updates linter versions, and fixes issues due to the version upgrades and using ruff.

### DIFF
--- a/.github/workflows/build-test.yml
+++ b/.github/workflows/build-test.yml
@@ -55,7 +55,7 @@ jobs:
     - name: Install dependencies
       run: |
         pip install .[docs,tests]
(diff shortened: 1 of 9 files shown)


BenjaminBossan:
So is the CI red here because we require #398 to be merged and CI on #398 is red because it requires this PR to be merged?

adrinjalali:
The CI here is passing the linter stage, so we can kinda consider this green I guess.

adrinjalali:
> Did you ruff to be better apart from speed, which I don't feel was a big issue to begin with.

It's speed, as well as having all the rules in one linter. We've switched in large projects to ruff and we're very happy with it.

PR #403, opened 2023-12-05
TST expect warning in API call

adrinjalali:
While running the tests, I realized this one fails since we raise a warning. This PR fixes that issue.

cc @BenjaminBossan


BenjaminBossan (approved):
Thanks for the fix. The Python 3.11 should probably pass after merging main?

PR #414, opened 2024-03-04
CI pin pytest to <8 for flaky

adrinjalali:
Flaky seems to fail with pytest 8, trying to pin then.


adrinjalali:
@BenjaminBossan I think we can merge this, do a release, and slowly remove deprecated HF API stuff, so that we can also slowly get rid of `flaky` then

PR #420, opened 2024-05-05
MNT - Removes sklearn-intelex and updates changelog

lazarust:
#### Reference Issues/PRs
Brought up in #394 

Removes `sklearn-intelex`


adrinjalali (commented):
we have a page in our documentation which uses this, and it's used in the API as well, those need to be removed as well.

lazarust:
@adrinjalali Sorry about that, everything should be removed now!

docs/changes.rst
```diff
@@ -11,7 +11,8 @@ skops Changelog
 
 v0.10
 ----
-
+- Removes Pythn 3.8 support and adds Python 3.12 Support :pr:`418` by :user:`Thomas Lazarus <lazarust>`.
+- Removes `sklearn-intelex` as a dependency :pr:`420` by :user:`Thomas Lazarus <lazarust>`.
```
adrinjalali, line 15: this was never a dependency of the package. We can write:

```suggestion
- Removes a shortcut to add `sklearn-intelex` as a not dependency.
  :pr:`420` by :user:`Thomas Lazarus < lazarust > `.
```

lazarust: Ah, so it wasn't technically a dependency since we were dynamically using it if it existed? Do I have the understanding of that right?

adrinjalali: Yes, correct. It was only a test/doc dependency. Not an install dependency

adrinjalali:
I think there's also a dead link here, the whole paragraph should be removed

```
- Using Intel(R) Extension for scikit-learn:
    :ref:`sphx_glr_auto_examples_plot_intelex.py` is an example of using
    Intel(R) Extension for scikit-learn to speed up inference of classical
    machine learning algorithms and how performance-optimized models work with
    Hugging Face Hub.
```

in `examples.rst`

adrinjalali (approved):
Thank you very much @lazarust

PR #421, opened 2024-06-03
MNT refresh expired token

adrinjalali:
HF invalidated our token again, so here's another one.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,6 +2,7 @@ comment: false
 codecov:
   branch: main
   require_ci_to_pass: true
+  token: 2b8d4d69-6de6-4e1d-840a-5ccf9d849565
   notify:
     after_n_builds: 12
     wait_for_ci: true
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -2,7 +2,6 @@ comment: false
 codecov:
   branch: main
   require_ci_to_pass: true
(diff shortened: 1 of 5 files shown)


BenjaminBossan:
Yes, tokens were invalidated last week: https://huggingface.co/blog/space-secrets-disclosure

I wonder if we should just deactivate all the tests that are based on HF tokens? Maybe just skip them unless an env var is set, and only run those tests locally? Not sure how relevant these parts of the skops functionality are overall vs the other parts, you probably know better.

adrinjalali:
codecov action from version 4 automatically switches to tokenless upload, which is useless 😠 

https://docs.codecov.com/docs/codecov-uploader#supporting-token-less-uploads-for-forks-of-open-source-repos-using-codecov

adrinjalali:
issue on codecov side: https://github.com/codecov/codecov-action/issues/1469

adrinjalali:
> I wonder if we should just deactivate all the tests that are based on HF tokens? Maybe just skip them unless an env var is set, and only run those tests locally? Not sure how relevant these parts of the skops functionality are overall vs the other parts, you probably know better.

Yeah those are I think the least used features. We can slowly deprecated them maybe. I would be in support of that. Can do in another PR.

@BenjaminBossan this is now ready for a merge :)

PR #422, opened 2024-06-06
remove `trusted=True` from skops.io.load(s)

adrinjalali:
We got a CVE report where an attacker can execute arbitrary code with `trusted=True`. While one can argue this is a feature and not a bug, it does pose the issue that people might be overly optimistic about the security of the format.

Therefore removing it might be a better idea. This doesn't change any functionality since users can still trust the file by blindly passing `trusted=get_untrusted_types(...)`

cc @BenjaminBossan 

I also have to say, having typehints made the process a LOT easier here.

CVE:https://www.recordedfuture.com/vulnerability-database/CVE-2024-37065

### DIFF
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -14,6 +14,8 @@ v0.10
 - Removes Pythn 3.8 support and adds Python 3.12 Support :pr:`418` by :user:`Thomas Lazarus <lazarust>`.
 - Removes a shortcut to add `sklearn-intelex` as a not dependency.
   :pr:`420` by :user:`Thomas Lazarus < lazarust > `.
+- ``trusted=True`` is now removed from ``skops.io.load`` and ``skops.io.loads``.
+  :pr:`XXX` by `Adrin Jalali`_.
 
 v0.9
 ----
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -15,6 +15,9 @@ v0.10
 - Removes a shortcut to add `sklearn-intelex` as a not dependency.
   :pr:`420` by :user:`Thomas Lazarus < lazarust > `.
 - ``trusted=True`` is now removed from ``skops.io.load`` and ``skops.io.loads``.
+  This is to further encourage users to inspect the input data before loading
+  it. :func:`skops.io.get_untrusted_types` can be used to get the untrusted types
+  present in the input.
   :pr:`422` by `Adrin Jalali`_.
 
 v0.9
--- a/docs/changes.rst
+++ b/docs/changes.rst
@@ -15,7 +15,7 @@ v0.10
 - Removes a shortcut to add `sklearn-intelex` as a not dependency.
   :pr:`420` by :user:`Thomas Lazarus < lazarust > `.
 - ``trusted=True`` is now removed from ``skops.io.load`` and ``skops.io.loads``.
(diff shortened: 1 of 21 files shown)


BenjaminBossan:
> We got a CVE report where an attacker can execute arbitrary code with `trusted=True`. While one can argue this is a feature and not a bug, it does pose the issue that people might be overly optimistic about the security of the format.

IMO this is a bit silly, with the same logic the whole pickle module would need to be removed. But I guess since there is a simple (albeit not really safer) solution, I'd be okay with the change. I'm wondering, however, if this should be deprecated?

> This doesn't change any functionality since users can still trust the file by blindly passing `trusted=get_untrusted_types(...)`

We could add a check that `__repr__` has been called at least once on the output of `get_untrusted_types` and at least 1 second per item has passed since. Just kidding of course, but yeah...

> I also have to say, having typehints made the process a LOT easier here.

Nice!

adrinjalali:
> I'm wondering, however, if this should be deprecated?

We used to have the hint that the API is really subject to change though. That's why I didn't deprecate. However, this has been used now for a while and pretty stable. So I removed the comment here and from now we can be even more conservative in changes.

docs/changes.rst
```diff
@@ -14,6 +14,8 @@ v0.10
 - Removes Pythn 3.8 support and adds Python 3.12 Support :pr:`418` by :user:`Thomas Lazarus <lazarust>`.
 - Removes a shortcut to add `sklearn-intelex` as a not dependency.
   :pr:`420` by :user:`Thomas Lazarus < lazarust > `.
+- ``trusted=True`` is now removed from ``skops.io.load`` and ``skops.io.loads``.
+  :pr:`422` by `Adrin Jalali`_.
```
BenjaminBossan, line 18: Maybe add an explanation.

BenjaminBossan (changes requested):
The changes LGTM (though I still don't think they're really necessary).

I think mypy should pretty much make sure that the change was propagated completely through the code base. I still grepped through the code and found some instances that may need updates (see comments).

Apart from that, how about adding a paragraph to the docs that explicitly mentions that the `trusted=True` option was removed (and why using it was a bad idea anyway). Since this is a breaking change, making sure that users can quickly find a reference to this would be beneficial.

adrinjalali (commented):
I think I addressed the comments @BenjaminBossan

adrinjalali:
@BenjaminBossan you okay with this now?

BenjaminBossan (approved):
Thanks for making the adjustments, this LGTM.

I have a nit about a type check, please take a look and decide if you want to fix it (you can merge no matter what).

While reviewing, I also stumbled upon the fact that `from typing import Sequence` is apparently deprecated:

https://docs.python.org/3/library/typing.html#typing.Sequence

Not sure if the tests should have caught that. Anyway, this can be addressed in a separate PR.

adrinjalali:
> While reviewing, I also stumbled upon the fact that from typing import Sequence is apparently deprecated:

Cool, I'll open a separate PR for that.

judahrand:
Could a release with this change be put out so that we can silence our CVE warning 😅

adrinjalali:
@judahrand yes, working on a numpy2 compatibility PR and will release with that one.

adrinjalali:
Release is out.

Issue #364, opened 2023-06-06
Space creator app action failing


adrinjalali:
The space creator app fails due to token permissions it seems:

```
ValueError: You must use your personal account token for write-access methods. To generate a write-access token, go to https://huggingface.co/settings/tokens
Creating and pushing to repo: scikit-learn/skops-model-card-creator-89d4537b-275b-4054-8c5e-599419aed8ac
Error: Process completed with exit code 1.
```

failing on `main`: https://github.com/skops-dev/skops/actions/runs/5142464818/jobs/9256239463

cc @BenjaminBossan

PR #423, opened 2024-06-06
Space creator on `main` only

adrinjalali:
Space creator uses a token, and now tokens need to be secret. This makes it run only on `main` and not PRs. Also updated the token secret.

fixes https://github.com/skops-dev/skops/issues/364


adrinjalali:
cc @BenjaminBossan

PR #426, opened 2024-06-19
MNT Improve typehints.

adrinjalali:
I started replacing `Sequence` with `list`, and it ended up being a rabbit hole that I couldn't easily get out of.

It also did raise some legit issues, so not all too bad, but there's a certain part that I can't figure out how to fix.

If you get a chance, I'd appreciate your wisdom @BenjaminBossan

### DIFF
--- a/.github/workflows/deploy-model-card-creator.yml
+++ b/.github/workflows/deploy-model-card-creator.yml
@@ -2,6 +2,7 @@ name: Deploy-Space-Creator
 
 on:
   - push
+  - pull_request
 
 concurrency:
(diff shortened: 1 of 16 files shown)


BenjaminBossan:
> I started replacing `Sequence` with `list`

Could you explain the reasoning? In general, for input arguments, it is preferred to use general types, so `Sequence` instead of `list` or `tuple`. For output types, it's the opposite. Was there a problem with `Sequence`?

> but there's a certain part that I can't figure out how to fix.

Could you explain what parts? Failing CI seems to be unrelated.

TamaraAtanasoska:
looking into this!

PR #434, opened 2024-07-16
DOC add warning on skops convert

adrinjalali:
Adds a warning in the docs for `skops convert` to tell people that this loads the pickle file.

cc @BenjaminBossan


BenjaminBossan (approved):
Makes a lot of sense.

PR #445, opened 2024-10-23
MNT add python 3.13 tests

adrinjalali:
Testing 3.13 and getting ready for a release.


adrinjalali:
@BenjaminBossan we've been having issues with the HF tokens in this repo. So I've removed the job updating the model card creator. I'm not sure how often it gets used, but since we haven't gotten any issues/feedback from it, I'm taking it as not very often.

I think we can merge this one, do a release, and then fix the issues for the unreleased version of sklearn and do another release once that's released.

BenjaminBossan (approved):
Thanks, makes sense to remove the broken workflow.

PR #456, opened 2024-12-10
MNT add funding file to apply to floss.fund

adrinjalali:
This is to apply for https://floss.fund/ which requires this file somewhere. I'd remove the file once the decision on the funding is made, either accept or reject.

Would this be okay @BenjaminBossan

### DIFF
--- a/docs/.funding/floss-funding.json
+++ b/docs/.funding/floss-funding.json
@@ -1,0 +1,62 @@
+{
+    "version": "v1.0.0",
+    "entity": {
+      "type": "individual",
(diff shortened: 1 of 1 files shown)


BenjaminBossan (approved):
I haven't heard of this fund, but :crossed_fingers: that it works out.

Also feel free to remove me from the maintainers list if it helps, since, realistically, I'm not doing maintenance work on skops anymore.

adrinjalali:
I still appreciate your reviews and it's nice not to have a single person with permissions, at least until we find a new maintainer.

PR #459, opened 2024-12-13
[don't merge] Test hfh v0.27.0.rc0

hanouticelina:

### DIFF
--- a/skops/_min_dependencies.py
+++ b/skops/_min_dependencies.py
@@ -13,7 +13,7 @@ PYTEST_MIN_VERSION = "7"
 #     "tomli": ("1.1.0", "install", "python_full_version < '3.11.0a7'"),
 dependent_packages = {
     "scikit-learn": ("0.24", "install", None),
-    "huggingface_hub": ("0.17.0", "install", None),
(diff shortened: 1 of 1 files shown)


BenjaminBossan:
@hanouticelina There has been a big refactor of the CI recently in #451. That's why this PR can't be run.

@adrinjalali do you think it's still necessary to check skops against HFH?

adrinjalali:
> @adrinjalali do you think it's still necessary to check skops against HFH?

I don't think so. They're not maintained and we've removed some functionalities and deprecated some (to be removed in the coming release).

I was actually debating whether to keep them or not, since we haven't really had any usage of the hub front, and removing the dependency (or making it optional) would make the package significantly lighter. WDYT?

BenjaminBossan:
Then I lean towards removing the dependency. For most dependencies, I might have told users: We have tested until version x.y.z, there is no support for newer versions. But given the nature of the HFH package, it's probably a better idea to remove the support completely.

hanouticelina:
thanks @BenjaminBossan, noted! I will close this PR then

PR #461, opened 2025-01-21
MNT fix pixi config and support sklearn>=1.3

adrinjalali:
We don't really support old numpy, and we haven't had an issue anybody complaining.

There was an issue with the pixi config that was picking up new versions, this fixes that.


adrinjalali:
This PR drops support for older python and scikit-learn. Once I fixed the CI to actually test against them, it showed we don't actually support them. So this PR makes it more accurate on what we support. Happy to see a PR fixing for the issues on older versions if somebody's up for it.

I'll fix the sklearn17 issue in a separate PR.

PR #462, opened 2025-01-21
MNT remove deprecated hub_utils

adrinjalali:
This PR removes deprecated `hub_utils` functions, and made me realise the other tools are not really used then (we haven't had an issue about them ever).

So I think it'd make sense to remove `hub_utils`, and references to it. It would mean to change the docs and model card related work as well. Before moving forward, I thought I'd ask if we're okay with me moving forward with it.

As far as I can tell, there hasn't been any activity about this on the HF hub since 2023, so on the HF side people probably wouldn't mind if these are removed.

cc @BenjaminBossan


adrinjalali:
And I'm not sure about the model card utils as well, since I haven't seen it used.

BenjaminBossan (approved):
I'm fine with removing the functionality, as there does not appear to be any usage. In general, it feels like skops is used for the serialization part and the rest is pretty much ignored.

adrinjalali:
Merging this, will tackle the CI issue in the other PR.

PR #466, opened 2025-01-29
MNT replace tabular with prettytable

adrinjalali:
tabulate started being flaky with the latest python/scikit-learn, and I checked and it's not very active anymore. Here we switch to `prettytable` which is very active and maintained. It slightly changes the table formats, but it should be fine.

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -14,8 +14,8 @@ from reprlib import Repr
 from typing import Any, Iterator, List, Literal, Optional, Sequence
 
 import joblib
+from prettytable import PrettyTable, TableStyle
 from sklearn.utils import estimator_html_repr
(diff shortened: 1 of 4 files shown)


skops/card/_model_card.py
```diff
@@ -214,13 +201,20 @@ def _check_table(self) -> None:
 
     def format(self) -> str:
         if self._is_pandas_df:
-            headers = self.table.columns  # type: ignore
+            pass  # type: ignore
         else:
-            headers = self.table.keys()
+            self.table.keys()
```
BenjaminBossan, line 206: Do these lines do anything? They can just be removed now, right?

adrinjalali: Nice catch

Issue #464, opened 2025-01-27
datetime.datetime support


mirmozavr:
Having issues using deserialization of objects, that have datetime objects in them (Python 3.11)

```python
from skops.io import dumps, loads
import datetime


cur_date = datetime.datetime.now()

d = dumps(cur_date)

l = loads(d, trusted=["datetime.datetime"])

```

>
    Traceback (most recent call last):
      File "C:\Program Files\JetBrains\PyCharm Community Edition 2024.1.1\plugins\python-ce\helpers\pydev\pydevconsole.py", line 364, in runcode
        coro = func()
               ^^^^^^
      File "<input>", line 9, in <module>
      File "D:\rc\venv_rc\Lib\site-packages\skops\io\_persist.py", line 192, in loads
        instance = tree.construct()
                   ^^^^^^^^^^^^^^^^
      File "D:\rc\venv_rc\Lib\site-packages\skops\io\_audit.py", line 165, in construct
        self._constructed = self._construct()
                            ^^^^^^^^^^^^^^^^^
      File "D:\rc\venv_rc\Lib\site-packages\skops\io\_general.py", line 456, in _construct
        instance = cls.__new__(cls)  # type: ignore
                   ^^^^^^^^^^^^^^^^
    TypeError: function missing required argument 'year' (pos 1)
>

adrinjalali:
Nice! will patch and do a release soon. Thanks for the report.

mirmozavr:
I discovered, that `builtins.range` doesn't work either

PR #467, opened 2025-01-31
ENH support types with a safe __reduce__

adrinjalali:
Fixes https://github.com/skops-dev/skops/issues/464

Sometimes the output of `__reduce__` is of the form `(type, (constructor_args,)` where `type` is the same as the `type(obj)`. In such cases, we can consider loading them _safe_. We still fail if the given type is not _trusted_, but there's no danger of loading those objects since we want to load them anyway. We would be calling their `__new__` and `__setstate__` otherwise.

This adds supports to many other types such as `datetime` and `slice`.

### DIFF
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -392,6 +392,8 @@ def object_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
     # This method is for objects which can either be persisted with json, or
     # the ones for which we can get/set attributes through
     # __getstate__/__setstate__ or reading/writing to __dict__.
+
+    # We first check if the object can be serialized using json.
     try:
         # if we can simply use json, then we're done.
         obj_str = json.dumps(obj)
@@ -405,6 +407,23 @@ def object_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
     except Exception:
         pass
 
+    # Then we check if the output of __reduce__ is of the form
+    # (constructor, (constructor_args,))
+    # If the constructor is the same as the object's type, then we consider it
+    # safe to call it with the specified arguments.
+
+    reduce_output = obj.__reduce__()
+    # note that we do "=="" to compare types instead of "is", since we only accept
+    # exact matches here.
+    if len(reduce_output) == 2 and reduce_output[0] == type(obj):
+        return {
+            "__class__": obj.__class__.__name__,
+            "__module__": get_module(type(obj)),
+            "__loader__": "ConstructorFromReduceNode",
+            "content": get_state(reduce_output[1], save_context),
+        }
+
+    # Otherwise we recover the object from the __dict__ or __getstate__
     res = {
         "__class__": obj.__class__.__name__,
         "__module__": get_module(type(obj)),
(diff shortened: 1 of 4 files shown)


adrinjalali:
WDYT @BenjaminBossan

skops/io/_general.py
```diff
@@ -405,6 +407,23 @@ def object_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
...
         pass
 
+    # Then we check if the output of __reduce__ is of the form
+    # (constructor, (constructor_args,))
+    # If the constructor is the same as the object's type, then we consider it
+    # safe to call it with the specified arguments.
+
+    reduce_output = obj.__reduce__()
+    # note that we do "=="" to compare types instead of "is", since we only accept
+    # exact matches here.
```
BenjaminBossan, line 417: I don't understand this comment. Checking `reduce_output[0] == type(obj)` is fine but when would using `is` not be fine too?

skops/io/_general.py
```diff
@@ -405,6 +407,23 @@ def object_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
...
+    if len(reduce_output) == 2 and reduce_output[0] == type(obj):
+        return {
+            "__class__": obj.__class__.__name__,
```
BenjaminBossan, line 420: For clarity, should we use `reduce_output[0].__name__` or `type(obj).__name__` here (i.e. one of the two objects used above)?

skops/io/_general.py
```diff
@@ -427,6 +446,24 @@ def object_get_state(obj: Any, save_context: SaveContext) -> dict[str, Any]:
...
+    def __init__(
+        self,
+        state: dict[str, Any],
+        load_context: LoadContext,
+        trusted: Optional[Sequence[str]] = None,
+    ) -> None:
+        super().__init__(state, load_context, trusted)
+        self.children = {
+            "content": get_tree(state["content"], load_context, trusted=trusted)
+        }
```
BenjaminBossan, line 459: Tbh I'm a bit rusty on this, do we need to set `self.trusted`?

skops/io/tests/test_persist.py
```diff
@@ -1119,3 +1120,17 @@ def test_dictionary(cls):
...
+    loaded_obj = loads(dumps(obj), trusted=[datetime])
+    assert obj == loaded_obj
+    assert type(obj) is datetime
+
+
+def test_slice():
+    obj = slice(1, 2, 3)
+    loaded_obj = loads(dumps(obj))
+    assert obj == loaded_obj
+    assert type(obj) is slice
```
BenjaminBossan, line 1136: Can we construct an example object that uses this `__reduce__` pattern but is not a trusted type, so that we can ensure that loading this object fails?

BenjaminBossan (commented):
Thanks for adding this. I have a few comments, but up to you whether you think they're worth addressing.

Also a more general observation: It feels like each new PR with code changes requires updating pixi stuff. Is this just coincidence, a transition thing that'll go away, or the "new normal"?

skops/io/_general.py
adrinjalali: I somehow thought `is` applies to subclasses, but it doesn't. So changing to `is`

skops/io/_general.py
adrinjalali: using `type(obj).__name__` now

skops/io/_general.py
adrinjalali: we pass trusted here if the user passes them, and sometimes we'd add to the trusted values if we have some default trusted types.

skops/io/tests/test_persist.py
adrinjalali: the `datetime` example above fails since it's not trusted, but I can also add another test.

adrinjalali (commented):
The pixi lockfile shouldn't change much under normal circumstances. Here it changes a lot since I managed to add an environment and support an older version of sklearn.

PR #472, opened 2025-03-05
FIX use rich.tree instead of rich.print

glemaitre:
This PR fixes the `rich` tree visualization.

It also using some colorblind color scheme: cyan / orange1 / magenta


adrinjalali:
I fixed a few things, but it seems the remaining two failing tests are legit?

glemaitre:
I’ll have a look. I’ll also try to test a couple of stuff if possible.

Guillaume Lemaitre
Open source engineer at :probabl.

adrinjalali (approved):

PR #473, opened 2025-03-19
ENH: Adding is_trusted function

kylermurphy:
Adding functionality to determine if datatype is trusted in the skops framework.

Intended to help users determine what they are able to save safely so that a model can be loaded without the need to okay/verify untrusted datatypes.

#### Reference Issues/PRs

No references.

This is meant to add simple functionality to determine if types are trusted within skops. The function is meant to help users identify what they can save within skops so that files can be loaded without the need to use the ```trusted``` keyword. This will hopefully help streamline the loading of ```.skops``` files.

#### Any other comments?

Also put a small change in ```skops\docs\persistence.rst``` to help clarify that *Supported* libraries also meant *Trusted* libraries. When reading the documentation it was unclear exactly what was *Trusted*. 

This is my first pull request in a larger project. I've tried to make sure I've followed the rules as layed out in the Contributing guidelines. Apologies if I have not!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -122,3 +122,6 @@ node_modules
 
 exports
 trash
+
+#spyder
+.spyproject
--- a/docs/persistence.rst
+++ b/docs/persistence.rst
@@ -231,7 +231,7 @@ mean you can just call `sio.load(<file>,
 trusted=get_untrusted_types(file=<file>))` on this object -- only pass the
 types you really trust to the ``trusted`` argument.
 
-Supported libraries
+Supported/Trusted libraries
 -------------------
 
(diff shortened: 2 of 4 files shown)


kylermurphy:
Thanks for the explanation!

If a node is injecting trusted types though that should be fine. The point of the function is to identify what types they can save to a skops file (e.g. numpy vs pandas) without have to use the ```trusted``` keyword to load the file at a later date.

adrinjalali:
In that case, you can do something along the lines of:

```py
def is_trusted_by_default(obj):
    return get_untrusted_types(dumps(obj)) == []
```

That seems easy enough that users don't need a function for it, WDYT?

PR #482, opened 2025-07-24
ENH harden Method and Operator node audits

adrinjalali:
@io-no WDYT of these hardened versions?


BenjaminBossan (commented):
Thanks for taking care of this and also big thanks @io-no for the great reports.

I'm too far out of this project by now to give a confident review of these changes, especially given the nature of this topic. I can only say that from my perspective, the PR looks solid.

PR #485, opened 2025-07-25
ENH harden Card by adding a flag to allow/disallow insecure pickle loading

io-no:
The PR introduces additional hardening for models loaded through `Card`.
If the model is in an insecure pickle format, it is now required to explicitly set a flag during `Card` object creation to allow the operation.

@adrinjalali

### DIFF
--- a/skops/card/_model_card.py
+++ b/skops/card/_model_card.py
@@ -227,7 +227,7 @@ class TableSection(Section):
         return f"{self.__class__.__name__}({nrows}x{ncols})"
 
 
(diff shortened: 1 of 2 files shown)


adrinjalali (commented):
This looks pretty nice. Could also please add a changelog entry?

Also tests to check the raised error message and to make sure having a pickle file with a `skops` extension fails by default.

io-no:
Hi @adrinjalali, added the changelog and tests

adrinjalali (approved):

PR #515, opened 2026-05-14
Guard loading against suspicious zip members

massy-o:
## Summary
- add a guarded ZIP member read helper to `LoadContext`
- reject members with excessive uncompressed size or suspicious compression ratios before reading them into memory
- route NumPy, bytes, and scipy sparse member reads through the guarded helper
- add a regression test covering `loads()` and `get_untrusted_types()`

## Tests
- `uv run --with-editable '.[rich]' --with pytest --with pytest-cov --with flaky --with pandas --with matplotlib pytest skops/io/tests/test_persist.py -k 'compression_ratio or compression_level or sparse_matrix'`
- local malformed `.skops` artifact check against `load()` and `get_untrusted_types()`

Note: `ruff check` on the touched files currently reports an existing E721 issue in `skops/io/_general.py` outside this change.

### DIFF
--- a/skops/io/_general.py
+++ b/skops/io/_general.py
@@ -633,7 +633,9 @@ class BytesNode(Node):
     ) -> None:
         super().__init__(state, load_context, trusted)
         self.trusted = self._get_trusted(trusted, [bytes])
-        self.children = {"content": io.BytesIO(load_context.src.read(state["file"]))}
+        self.children = {
+            "content": io.BytesIO(load_context.read_zip_member(state["file"]))
+        }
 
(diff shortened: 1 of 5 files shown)


adrinjalali (commented):
Unless there's some common patterns in loading zip files in other libraries which we're replicating here, I wouldn't want to put these limits. This basically means the user can't load large objects, and users might have legit usecases for loading large objects. So I'm not sure if this heurestic is sensible here.

Happy to reopen if there are good arguments for this change, otherwise closing.

PR #517, opened 2026-05-26
MNT remove HF-specific overflow injection from model plot

adrinjalali:
## Summary

The sklearn nightly tests started failing on PR #516 because newer sklearn (1.9+) emits `sk-top-container` twice in the HTML repr — once as a CSS selector (`.sk-top-container.sk-global { ... }`) and once as an HTML class attribute. A naive string replace would have corrupted the new CSS.

Looking at why the injection existed in the first place: it was specifically to work around how sklearn's HTML repr renders inside the HuggingFace Hub models page. Since skops no longer targets HF Hub (see #462 and #512), this HF-specific HTML manipulation is no longer needed.

- Remove the inline `style=\"overflow: auto;\"` injection from `_add_model_plot`.
- Drop the corresponding `test_no_overflow` test.
- Add a changelog entry.

## Test plan

- [x] `pixi run -e ci-sklearn18 pytest skops/card/tests/test_card.py` (140 passed)
- [x] Initial fix commit passed sklearn-nightly on Ubuntu and macOS in CI
- [ ] Full matrix CI on this PR


skops/card/_model_card.py
```diff
@@ -771,10 +771,6 @@ def _add_model_plot(
 
         """
         model_plot_div = re.sub(r"\n\s+", "", str(estimator_html_repr(model)))
-        if model_plot_div.count("sk-top-container") == 1:
```
adrinjalali, line 774: these were for HF specific overflow issue which is not used anymore.

PR #521, opened 2026-07-05
Add trusted types for KNeighbors models

DresdenGman:
Adds the sklearn helper types used by `KNeighborsClassifier(algorithm="kd_tree")` to the default trusted types list, so these models can be loaded without manually passing those internal sklearn types.

The change follows the existing trusted-types pattern and adds regression coverage for the KNN persistence path.


DresdenGman:
Checks are green on my side. This is ready for review; the change adds the trusted KNeighbors internal types and the regression test for the kd_tree persistence path. Please let me know if you want any adjustments.

DresdenGman:
Friendly follow-up: this PR remains clean and ready for review. It adds the trusted internal KNeighbors types together with regression coverage for the `kd_tree` persistence path. Please let me know if you would prefer any change to the trusted-type scope or tests.

adrinjalali:
Hey, sorry for the delay. Been pretty hectic on my side.

Regarding this, we need to check if there's a way for somebody to create a persisted version which upon loading, gives a segfault. Otherwise LGTM.

